PR #107, opened 2025-11-06
Fixes to update

axelpey:


vignesh-anand (approved):
Approving, but i don't actually understand the update system completely, so you'll have to explain it to me at some point

PR #108, opened 2025-11-08
Logging basic

axelpey:


axelpey:
We'll have to log crashes later, this is nice to begin with

PR #110, opened 2025-11-09
Vig/combined arm head cpp node

vignesh-anand:
Got performance of latency down from like 300ms + to sub 60ms both wired and on network.

Achieved through:

1) Jetson wifi configs, that disable power saving behavior that was causing massive jitter and in general bad latency 
2) Entire arm node is now multi threaded cpp, with different serial read/write methods profiled, current method reads both position and velocity in one read command (saves 15 ms each loop), optimized write behavior (about 10ms) general cpp improvements and (not profiled exactly but significant). Overall going from sub 30 hz control to well above 100 with low latency
3) Head and arm on same uart bus running at 1M baud, this makes separate head package unnecessary, and also servo manager unnecessary so deleted. 

its been a very very very long time since i wrote cpp, so feedback appreicated.


ros2_ws/src/maurice_bot/maurice_arm/maurice_arm/arm.cpp
```diff
@@ -0,0 +1,571 @@
...
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+
+namespace maurice_arm {
+
+class MauriceArmNode : public rclcpp::Node {
+public:
+    MauriceArmNode() : Node("maurice_arm") {
+        RCLCPP_INFO(this->get_logger(), "Maurice Arm Node starting...");
```
axelpey, line 22: this is low-key ridiculous, is there not just a "log_info" function we can define so that we don't have to use this->get_logger every time?

ros2_ws/src/maurice_bot/maurice_arm/maurice_arm/arm.cpp
```diff
@@ -0,0 +1,571 @@
...
+        // Create callback groups for parallel execution
+        timer_callback_group_ = this->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
+        service_callback_group_ = this->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
+        
+        // Declare parameters
+        this->declare_parameter("baud_rate", 1000000);
+        this->declare_parameter("control_frequency", 100.0);
+        this->declare_parameter("joints", "{}");
+        
+        int baud_rate = this->get_parameter("baud_rate").as_int();
```
axelpey, line 33: why "this->"? can't we just call the thing?

axelpey (changes requested):

ros2_ws/src/maurice_bot/maurice_arm/maurice_arm/arm.cpp
```diff
@@ -0,0 +1,571 @@
...
+    void initializeServos() {
+        RCLCPP_INFO(this->get_logger(), "Configuring all 7 servos...");
+        
+        // Configure all servos (IDs 1-7) uniformly
+        for (const auto& config : joint_configs_) {
+            RCLCPP_INFO(this->get_logger(), "Configuring servo %d", config.servo_id);
+            
+            RCLCPP_INFO(this->get_logger(), "  Disabling torque on servo %d", config.servo_id);
+            dynamixel_->disableTorque(config.servo_id);
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
nickjw1243, line 182: Startup can probably be sped up if instead of using a blocking sleep, we wait until a correct status packet is returned

ros2_ws/src/maurice_bot/maurice_arm/maurice_arm/dynamixel.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+void Dynamixel::enableTorque(int motor_id) {
+    uint8_t dxl_error = 0;
+    int dxl_comm_result = packet_handler_->write1ByteTxRx(
+        port_handler_, motor_id, ADDR_TORQUE_ENABLE, 1, &dxl_error);
+    
+    if (dxl_comm_result != COMM_SUCCESS) {
+        throw std::runtime_error("Failed to enable torque for motor " + std::to_string(motor_id));
+    }
```
nickjw1243, line 56: Could add a short loop here to read status packets until torque is confirmed enable to allow for eliminating 100ms delays in arm.cpp

PR #133, opened 2025-12-05
fix: fix ros topic list so that is shows on first try

theo-michel:
It silently starts the ROS 2 daemon in the background without showing any output or error messages. This is commonly done at startup to ensure the daemon is running before you start using ROS 2 commands, improving their performance.


axelpey (approved):
LGTM

PR #144, opened 2025-12-11
Add head camera TF transform publisher

Heemyk:
- Publish dynamic transforms for head_camera_link, head_camera_optical_frame, and camera frames
- Publish static transform for head_pivot frame (rotation axis)
- Camera geometry based on calibration data: base offset + rotating offset
- tf2_ros dependency added to CMakeLists.txt and package.xml


axelpey:
@vignesh-anand pls review soon

PR #150, opened 2025-12-17
Proxy Client OS Refactor

Heemyk:


axelpey:
In general, weird that proxyopenai imports proxyclient. I think proxyclient should contain both, but not proxyopenai containing proxyclient

ros2_ws/src/brain/brain_client/brain_client/input_types.py
```diff
@@ -176,6 +181,42 @@ def is_active(self) -> bool:
...
+        Usage:
+            # Access services
+            self.proxy.openai.realtime.connect_sync(...)
+            self.proxy.cartesia.tts.sse(...)
+            
+            # Access config
+            model = self.proxy.config.get("openai_realtime_model", "default")
+        
+        Returns:
+            ProxyClient instance or None if not configured
```
theo-michel, line 185: I think those comments are not usefull, it's a bit overkill

theo-michel (approved):
My only suggestion is that there are a few too many comments explaining parts of the code that are already self-explanatory due to clear naming and structure. Well-written code should generally be readable without relying heavily on comments. Removing the unnecessary ones would make the file less verbose and easier to scan, while keeping the truly helpful comments that add context or explain non-obvious logic.

But it's good overall!

.env.template
```diff
@@ -2,12 +2,30 @@
...
+# GOOGLE_API_KEY=your_google_api_key_here
+
+# ===== Service Configuration =====
+# Note: These are now ROS2 launch parameters with sensible defaults.
+# Only add here if you need to override at environment level.
+# See brain_client.launch.py for configurable parameters:
+#   - cartesia_voice_id (default: f786b574-daa5-4673-aa0c-cbe3e8534c02)
+#   - openai_realtime_model (default: gpt-4o-realtime-preview)
+#   - openai_realtime_url (default: wss://api.openai.com/v1/realtime)
+#   - openai_transcribe_model (default: gpt-4o-mini-transcribe)
```
theo-michel: I think we can remove that probably!

inputs/micro_input.py
```diff
@@ -29,140 +27,232 @@
...
+                error_obj = event.get("error", {})
+                error_code = error_obj.get("code", "")
+                error_msg = error_obj.get("message", "")
+                error_param = error_obj.get("param", "")
+                if error_code != "input_audio_buffer_commit_empty" and self.logger:
+                    self.logger.error(f"❌ OpenAI error: {error_code} - {error_msg} (param: {error_param})")
+            else:
+                # Log other message types for debugging
+                if self.logger:
+                    self.logger.info(f"📨 OpenAI event: {etype}")
```
theo-michel: I think we can simplify this. Try rewriting the code so that the indentation is less deep.

inputs/micro_input.py
```diff
@@ -29,140 +27,232 @@
...
+                self.logger.info("📤 session.update sent")
+        
+        def on_error(error):
+            if self.logger:
+                self.logger.error(f"[ws error] {error}")
+        
+        def on_close():
+            if self.logger:
+                self.logger.warning("WebSocket closed")
+
```
theo-michel: The if self.logger everywhere makes the code less nice. There are better patterns for this maybe just havving a logger instance. That you init and then the instance knows when logging is activated or not.

```
import os


class Logger:
    def __init__(self, enabled: bool = False):
        self._enabled = enabled or os.getenv("DEBUG", "").lower() in ("1", "true", "yes")
    
    def info(self, msg: str):
        if self._enabled:
            print(f"[INFO] {msg}")
    
    def debug(self, msg: str):
        if self._enabled:
            print(f"[DEBUG] {msg}")
    
    def error(self, msg: str):
        if self._enabled:
            print(f"[ERROR] {msg}")


class RobotController:
    def __init__(self, debug: bool = False):
        self.logger = Logger(enabled=debug)
    
    def navigate(self, x: float, y: float):
        self.logger.info(f"Navigating to ({x}, {y})")
        self.logger.debug("Path computed")
```

theo-michel (commented):
Last Nitpick the logger usage is inconsistent across the code, it's done differently a bit everywhere makes the code hard to read.

And try to minimize indentation depth whenever possible, those lead to more errors, and harder to read code. Early return and rethinking logic often helps, llms usually are bad at this as they don't think far enough ahead when writing the code to factorize different parts of if else statements for example.

axelpey:
@greptile review please

PR #177, opened 2025-12-22
Directives to agents primitives to skills

axelpey:


theo-michel:
@axelpey the greptile stuff is nitpicks, but would be good to address

PR #182, opened 2025-12-30
Switched ttyTHS1 for ttyACM0

Heemyk:
yep


vignesh-anand:
Uart.py has nothing to do with the arm, i would revert it

PR #183, opened 2025-12-30
Fix sim docker start with zenoh

axelpey:


vignesh-anand (approved):

karmanyaahm (approved):
Tested on my machine. Launching nodes and rviz work well

PR #184, opened 2025-12-31
Sync hostname to roboname

karmanyaahm:
- refactor robot_info.json set/get logic to be independent functions
- every 30 seconds, check system hostname and compare to robot name from robot_info. If they don't match, set the system hostname. 
- Now we can connect to robots like `ssh jetson1@iron-bird.local`
- Tested on iron bird seems to work fine

Closes INN-185


karmanyaahm:
@greptileai

axelpey:
Confirmed it works. Mergign

axelpey:
Note that we should maybe display in the app under "Wifi" Tab in settings, what is the address we can connect to under ssh, including this hostname one, or with the standard numbered ip.

axelpey:
It's INN-233

PR #185, opened 2026-01-04
Improve switching navigation modes

karmanyaahm:
- use lifecycle managed version of slam-toolbox (custom apt package defined in https://github.com/innate-inc/innate-packages)
- Make an alternate planner and bt navigator that runs in both mapfree mode and under /mapfree namespace in navigation mode
- Publish empty map and map->odom tf in mapfree mode
- Make grid localizer and everything lifecycle managed
- Make mode_manager the lifecycle manager so when switching between mapped and mapfree we can switch as optimally as possible, and remove all the launch file starts and stops
- Change some costmap stuff for better performance

End result: faster switching and possibly startup between mapped, mapfree, mapping, and most importantly, the ability to run primitives in odom frame (like navigate in sight).


axelpey:
@karmanyaahm I see commits up to 14hrs ago, is this ready?

karmanyaahm:
@greptile-apps

scripts/update/post_update.sh
```diff
@@ -343,7 +369,7 @@ if [ -d "$REPO_DIR/ros2_ws/src" ]; then
     cd "$REPO_DIR/ros2_ws"
 
     # Run as the actual user, not root
-    sudo -u "$ACTUAL_USER" bash -c "cd $REPO_DIR/ros2_ws && source /opt/ros/humble/setup.bash && rm -rf build/ install/ log/ && colcon build"
+    sudo -u "$ACTUAL_USER" bash -c "cd $REPO_DIR/ros2_ws && source /opt/ros/humble/setup.bash &&  colcon build"
```
karmanyaahm:
```suggestion
    sudo -u "$ACTUAL_USER" bash -c "cd $REPO_DIR/ros2_ws && source /opt/ros/humble/setup.bash && rm -rf build/ install/ log/ && colcon build"

```

ros2_ws/src/maurice_bot/maurice_nav/config/mapfree_costmap.yaml
vignesh-anand, line 1: It would be nice if this was in the same file ads costmap?

vignesh-anand (commented):
1) How are we handling concurrency if both a global and local goal is sent 
2) It would be nice to specificy different xy goal tolerance based on mode

axelpey:
@vignesh-anand perf test
seems pretty similar. nav before refacto seemed to have gone higher. This 75% was sustained over about 20-25sec.

So I killed the webrtc node and made global cpu measurements (cause there's several components to this module so it'sbetter to measure overall differences)

with refacto
in mapfree: around 59% idle, up to 67% when setting a goal pose
in nav: 59.5% idle, 68% with a goal pose
in mapping: 63% idle, 63% when moving

without refacto
nav: 63.5% idle, 75% witha. goal pose
mapfree: 59% (and no goal pose settable)
mapping: 59%, up to 63% when moving

ros2_ws/src/maurice_bot/maurice_nav/config/mapfree_costmap.yaml
karmanyaahm: fixed

axelpey:
Pending agent merge with ability to navigate locally, good to me

PR #189, opened 2026-01-05
Add basic and demo agents, enhance gaze tracking, and implement turn …

theo-michel:
…and move skill

- Performance optimization 10sec to 2.5 (with new backend cloud agent
- Introduced BasicAgent and DemoAgent classes for basic and demo functionalities.
- Enhanced gaze tracking capabilities with a new GazeController and ROSPersonTracker.
- Implemented TurnAndMove skill for basic robot movement via voice commands.
- Updated BrainClientNode to manage gaze tracking based on current directive.
- Improved WebSocket client with automatic reconnection and optimized message handling.


axelpey:
Just rebased

axelpey:
@theo-michel ok this is pretty cool BUT it seems like the fast agent is always running even when there's no agent triggered at all. Is that what you wanted?

axelpey:
Also when I ask the agent if it can do something like waving it says it's a multi step thing which I don't think it is?

```
Decision: defer_to_expert - The user is asking me to wave, which requires movement and is a multi-step task. This requires me to defer to the expert.
websocket-server-1  | [Brain R7-9] [FastAgent] Deferring to vision agent - message stored for next image
websocket-server-1  | [Brain R7-9] Stored user message for next image: 'That's not what I'm asking you. Can you wave at me...'
```

But also this could be just happening cause it does not see the available skills cause it's not running an agent?

axelpey:
Like it's pretty cool definitely, but not mergeable yet

ros2_ws/src/brain/brain_client/brain_client/brain_client_node.py
```diff
@@ -661,25 +681,57 @@ def _wait_for_input_manager(self, timeout_sec=5.0):
...
+            payload["image_b64"] = image_b64
+        
+        outgoing_msg = MessageIn(type=MessageInType.CHAT_IN, payload=payload)
         self.ws_bridge.send_message(outgoing_msg)
-        self.get_logger().info(f"\033[1;92mSent MessageIn: {outgoing_msg}\033[0m")
+        self.get_logger().info(f"Sent MessageIn: {outgoing_msg}")
 
     def tts_callback(self, msg: String):
         """Handle direct TTS requests from /brain/tts topic."""
+        if not self.is_brain_active:
```
axelpey: Users still might want to TTS manually from manual control screen though

axelpey (commented):
Working on it

PR #191, opened 2026-01-08
Disable shared memory on Zenoh Router

karmanyaahm:
Remote Zenoh Routers were unable to communicate with the main one because of this.

Disabling SHM for router still allows the local nodes to talk to each other using SHM, but when something goes to the router (i.e. when a remote node is subscribed to a topic) it gets sent over the network, instead of using SHM.

From testing: as soon as a remote node subscribes to the image topic, it goes down to 12fps or so for ALL subscribes (even local ones). As soon as it unsubscribes, it goes back up to 30fps. So any transport selection is per-message, which is the best we can get I guess without a lossy zenoh serializing proxy

Closes INN-230


vignesh-anand (approved):
approved, subject to testing

karmanyaahm:
@greptile-apps

karmanyaahm:
For posterity, we don't care about any of the other settings changes in satellite mode, that's why the config is overriding not appending.

PR #195, opened 2026-01-13
Tf tree update

nickjw1243:
Updating TF and URDF for navigation improvement.

URDF translation values updates, meshes updated and head and left/right camera links added

Added publishing of arm joints, head joint, and odom to joint_state_publisher

Need to delete odom, update URDF (redundant as of current state)


vignesh-anand (commented):
#Fix camera in bring up, better variable names in arm

vignesh-anand (approved):
Noticed 5 percent cpu use but looks good

PR #198, opened 2026-01-14
Improve recorder to cpp

axelpey:


vignesh-anand (commented):
1) It would be good to implement some episode max time limit ( untill we implement the streaming thing) to prevent robot crashes if people leave it running too long
2) Before merging i want to do a quick cpu test of running this node as part of my camera container (should save 10% cpu)
3) Critical - > There is a mismatch between the python and cpp in that the replay topic in python uses raw images , cpp uses compressed ( my camer pr chnaged python to raw to save decent amount of cpu)

PR #204, opened 2026-01-17
fix post update always requiring reboot

karmanyaahm:
closes INN-245


axelpey:
btw fyi, total build time here is measured at 3mn19s in the post-update here

PR #208, opened 2026-01-22
Replace rosbridge with C++ bridge

karmanyaahm:
Leaving rosbridge in for sim because there are still some unimplemented features in RWS

Closes INN-168


axelpey:
Confirmed it runs

axelpey:
Actions work well too

PR #209, opened 2026-01-24
Enable release build in post_update

karmanyaahm:
This needs extremely thorough testing for both performance and reliability

Closes: INN-265


axelpey:
Closing the long-stale draft. The Release-build idea and required reliability testing remain tracked in INN-265 and should be revisited from current main.

PR #210, opened 2026-01-26
Vig/run policy

axelpey:


axelpey:
AFAIK this works and runs policies properly, but the socks policy still needs to be better

PR #278, opened 2026-02-16
Codex/open vocab grasp exp chess skill main rebased

axelpey:


axelpey:
Superseded by #279 (clean recreation from main with a single feature commit).

PR #289, opened 2026-02-20
Hi-res wrist camera for chess

axelpey:


axelpey:
Closing as obsolete against the removed chess/brain architecture. Any still-useful behavior should be reimplemented from current main.

PR #318, opened 2026-03-16
fix(ci): disable Zenoh SHM in integration tests to fix OS error 12

innate-openclaw:
## Problem

Both integration tests (`test_skill_integration` and `test_pose_image`) fail immediately on startup with:

```
zenohc::shm: Unable to create POSIX shm segment: OS error 12
```

`ENOMEM` — Zenoh SHM transport fails to allocate a shared memory segment inside the CI Docker container, even with `shm_size: 2g` set. This kills `skills_action_server.py` and `brain_client_node.py` before any test logic runs.

## Fix

Disable Zenoh SHM transport for tests via a `zenoh_test_config.json5` config file mounted into the container. Nodes still communicate fine over UDP/TCP locally — SHM is only an optimization and is not needed for correctness in CI.

**Changes:**
- `zenoh_test_config.json5` — new config with `transport.shared_memory.enabled: false`
- `docker-compose.test.yml` — mount config + set `ZENOH_ROUTER_CONFIG_URI` and `RMW_ZENOH_CONFIG_FILE` env vars


karmanyaahm:
@innate-openclaw can u hear me

PR #323, opened 2026-04-02
Import simulator

axelpey:
## Reviewer UX: How A New Developer Uses This

This PR makes the simulator feel like a first-class local workflow inside `innate-os`, while keeping the actual simulator implementation under `sim/`. A reviewer who has never used the simulator before should think of the root `./innate` file as the one tiny entrypoint, not as where the simulator lives. The launcher code is in `sim/launcher/`.

The intended fresh-clone flow is:

```bash
git clone git@github.com:innate-inc/innate-os.git
cd innate-os
./innate setup
./innate sim up
```

`./innate setup` is the short first-run setup command. `./innate sim setup` also exists and calls the same setup path, but the shorter command is what we should show new users during onboarding. After setup, day-to-day simulator commands stay namespaced under `./innate sim ...`.

On first setup, the launcher prepares the simulator Python environment, builds the frontend if needed, pulls the locked simulator asset pack, fetches required scene data, and prompts for a hosted Innate service key when hosted brain mode is configured. On `./innate sim up`, it tries to pull matching prebuilt GHCR images for the current checkout and falls back to local Docker/ROS builds only when those exact images are unavailable.

The normal daily workflow is:

```bash
./innate sim up              # start, or reattach the dashboard if already running
./innate sim down            # stop OS, sim, and optional local cloud agent
./innate sim status          # show runtime status
./innate sim logs startup    # inspect startup logs
./innate sim logs brain      # inspect OS brain logs
./innate sim logs simulator  # inspect simulator logs
```

Useful variants are:

```bash
./innate sim up --once       # start and print one status snapshot, no live dashboard
./innate sim up --vis        # start with the native Genesis viewer enabled
```

For reviewers: the happy path should end in the terminal dashboard showing OS, simulator, brain/backend, FPS, frame age, and queue health. The browser UI is served at `http://localhost:8000`. The simulator should not silently look healthy when the hosted backend is missing a key or still connecting.

## Summary

This PR imports the simulator into the Innate OS repository and adds the monorepo-local workflow around it.

The first part is the simulator import itself under `sim/`, preserving the simulator history. The rest of this description focuses on what changed after that import so reviewers can separate the imported app from the integration work done on top of it.

### One-command local workflow

- Added the root `./innate sim ...` CLI backed by `sim/launcher/main.py`.
- Added stack lifecycle commands for `init`, `up`, `down`, `status`, and `logs`.
- Added launcher env templates and docs for local simulation setup.
- Added tmux/zsh launch helpers for ROS and sim processes.
- Hardened macOS/Docker startup handling with clearer progress reporting, container readiness polling, and compact dashboard output.
- Auto-fetches required simulator data when missing, instead of failing later with unclear file errors.

### Monorepo configuration cleanup

- Simplified simulator/OS local config handling and env loading.
- Added `config/os.toml.template`, `sim/config.toml.template`, and launcher `.env` support.
- Ignored legacy local OS config so developer-local secrets/config do not leak into the repo.
- Added runtime env inspection helpers for debugging launcher/container state.

### Docker, ROS build, and startup performance

- Added Docker build caching for apt, uv/pip, and colcon where applicable.
- Switched container Python package installation paths to `uv`.
- Pruned simulation image dependencies, especially unnecessary hardware/GPU-side packages for sim builds.
- Added support for a prebuilt ROS simulation image so developers do not need to rebuild the full ROS workspace locally every time.
- Made the prebuilt ROS install self-contained and refreshed when the selected image changes.
- Auto-selects the matching prebuilt sim image using dependency/input digests, with a local-build fallback when the exact image is unavailable.
- Improved fallback logging so a missing prebuilt image is shown as an expected fallback instead of a noisy Docker failure.

### Sim runtime performance

- Upgraded Genesis to `0.4.6`.
- Reworked simulator startup waits to avoid unnecessary delay.
- Lightened simulator collision/rigid settings while keeping collisions enabled for the current local workflow.
- Aligned quadrants/Genesis versions and fixed Genesis 0.4.x mesh orientation compatibility issues.

### Image publishing/CD groundwork

- Added Cloud Build config for building simulator images.
- Added GitHub Actions workflows to publish simulator images only from `main`.
- Publishes both required simulator images:
  - ROS/prebuilt image
  - runtime/dev simulator image
- Added image tag cleanup workflow for old simulator image tags.
- Uses GHCR image tags for `main`, input digests, and commit SHAs so local setup can pull a compatible prebuilt image when available.

### Frontend and simulator UX fixes

- Leaves the simulator frontend with no behavior agent selected on initial load.
- Preserves an explicitly selected agent only while it remains available.
- Adds clearer UI/backend status when the hosted brain backend is disconnected, missing credentials, or still warming up.
- Avoids showing a misleading connected state during short-lived hosted-agent websocket connects/disconnects.
- Surfaces missing or placeholder `INNATE_SERVICE_KEY` clearly instead of leaving the frontend looking healthy while the backend cannot authenticate.

### Hosted brain / OS connection logging

- Normalizes hosted brain message types such as `brain/chat_out` before validation.
- Replaces noisy Pydantic validation dumps with concise unsupported-message logs.
- Adds brain backend status tracking through the sim bridge and `/get_available_agents`.
- Improves websocket close/error logs, including recent backend auth/config context when the socket closes with code `1000` and no reason.
- Fixes the manipulation server import path so the behavior server boots correctly in the stack.

## Validation

Local checks run across the follow-up PRs merged into this stack branch:

- `python3 -m py_compile` on changed launcher, sim bridge, and ROS brain/manipulation Python files.
- `npx eslint src/App.tsx src/services/rosbridgeService.ts` in `sim/frontend`.
- `npm run build` in `sim/frontend`.
- `git diff --check`.
- `./innate sim up --once` for end-to-end local startup.
- Browser/API verification at `http://localhost:8000` for:
  - frontend loads
  - no default agent is selected initially
  - backend missing-key state is visible instead of silently showing healthy/connected
  - sim status/dashboard reports backend connection state

## Notes

- The branch intentionally keeps the imported simulator history visible, so the PR commit list is large. The integration work after the import is summarized above.
- If a matching GHCR prebuilt image is unavailable, local setup falls back to building Docker/ROS locally.
- Hosted cloud-agent mode requires a real `INNATE_SERVICE_KEY`; placeholder values are now treated as missing config and surfaced clearly.


sim/data/urdf/turtlebot3_description/meshes/sensors/astra.dae
theo-michel, line 1: I think we can remove this file, it's not used

sim/data/urdf/turtlebot3_description/urdf/turtlebot3_waffle.urdf
theo-michel, line 1: We can probably remove all the turtlebot

theo-michel:
Note: We should try to not have things in double here from sim and from non sim

PR #327, opened 2026-04-08
Comment out CARTESIA_VOICE_ID in .env.template

karmanyaahm:
Comment out the CARTESIA_VOICE_ID for configuration.


karmanyaahm:
sets that up in setup script otherwise

PR #347, opened 2026-04-10
manipulation server optimization

theo-michel:
image

Also removed STT running all the time
Added a service to give status about the training
Also manipulation server optimization
Decoding with nvdecode
skills server opti


karmanyaahm (approved):
lgtm but check if you can just use the topic instead of the single skill service ig

PR #348, opened 2026-04-10
Mirror INNATE_SERVICE_KEY to /etc/innate/.env

karmanyaahm:
## Summary
- `post_update.sh` reads `INNATE_SERVICE_KEY` from `~/innate-os/.env` in a subshell and mirrors it to `/etc/innate/.env` (mode 644, dir 755) so processes without access to the user's home can still authenticate.
- `maurice_bringup.env_loader.load_env_file` and `auth_client/__init__.py` now read `/etc/innate/.env` as a **last-resort** fallback (`override=False`), so any value already in the environment or in the per-process `.env` wins. Missing file is a silent no-op.
- `input_manager.launch.py` is refactored to use `load_env_file()` instead of its own inline parser — picks up the fallback for free.

Coverage:
- Launch files (brain_client, brain_client.sim, input_manager) → covered via `env_loader`.
- Cloud nodes (innate_logger, innate_training_node, innate_uninavid, innate_proxy) → covered via `auth_client/__init__.py`, which they all import.

## Test plan
- [ ] On a robot: run `sudo ./scripts/update/post_update.sh`, confirm `/etc/innate/.env` is created with `INNATE_SERVICE_KEY=...` and mode 644.
- [ ] Confirm post_update doesn't fail when `~/innate-os/.env` is missing or has no `INNATE_SERVICE_KEY`.
- [ ] With `INNATE_SERVICE_KEY` removed from `~/innate-os/.env` but present in `/etc/innate/.env`, restart `ros-app.service` and verify the brain client + cloud nodes still authenticate.
- [ ] With `INNATE_SERVICE_KEY` set in **both** files, verify the user `.env` value wins.

### DIFF
--- a/ros2_ws/src/brain/brain_client/launch/input_manager.launch.py
+++ b/ros2_ws/src/brain/brain_client/launch/input_manager.launch.py
@@ -3,21 +3,13 @@ from launch.actions import DeclareLaunchArgument
 from launch.substitutions import LaunchConfiguration
 from launch_ros.actions import Node
 from brain_client.logging_config import get_logging_env_vars
-import os
-from pathlib import Path
+from maurice_bringup.env_loader import load_env_file
 
 
 def generate_launch_description():
-    # Load environment variables from .env file if it exists
-    innate_os_root = os.environ.get('INNATE_OS_ROOT', os.path.join(os.path.expanduser('~'), 'innate-os'))
-    env_file_path = Path(innate_os_root) / ".env"
-    if env_file_path.exists():
-        with open(env_file_path) as f:
-            for line in f:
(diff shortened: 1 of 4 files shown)


karmanyaahm:
@copilot+claude-opus-4.6 edit this pr to keep only the post update migration, and not the rest. restrict the perms too

karmanyaahm:
@copilot+claude-opus-4.6 instead of grep use source no cuz its in env format

karmanyaahm:
and also @copilot update the pr descriptoin

karmanyaahm:
oh i guess @theo-michel already did that

PR #354, opened 2026-04-11
Vig/best webapp ever

vignesh-anand:


theo-michel:
This is a temporary solution for the duraciton of the hackaton, we will switch to more lightweight

theo-michel (approved):
Temporary for the hackaton ok!

PR #356, opened 2026-04-11
Teleop

greg-schneid:
- Add `leader_arm_sender.py` — a UDP client that reads Dynamixel servo positions from a locally-connected leader arm
   and streams them to the robot's `udp_leader_receiver` node
   - Add `innate agent` CLI subcommands (`list`, `set`, `startup`, `reload`, `activate`) for managing robot agents via ROS2 services


axelpey:
This is interesting but we'll have to implement the agent cli ourselves, and provide the teleop script separately :)

PR #361, opened 2026-04-12
Fix arm part alignment in MARS/Maurice MJCF model

yunho-c:
The alignment of arm in existing MJCF model (`ros2_ws/src/maurice_bot/maurice_sim/mjcf/maurice.mjcf`) doesn't seem to be correct—maybe it was introduced during URDF to MJCF conversion. This PR does a best-effort fix, just going off of the visuals and the joint limits of servos by inspection, and may not be entirely accurate. 

Before fix:
image

After fix:
_ 2026-04-11 at 5 55 02 PM


karmanyaahm:
Hey @yunho-c — we're relicensing innate-os from AGPLv3 to Apache 2.0 (the standard for the ROS ecosystem). Could you confirm you agree to have your contributions relicensed under Apache 2.0? A 👍 or reply on this issue works.

yunho-c:
Yes confirmed! Thanks for checking :)

PR #373, opened 2026-04-12
Feat/chemi agent

praveenVnktsh:


axelpey:
PR from the hackathon winners, not really good to merge as is BUT should go at some point in our list of use-cases. Linear is taking notes

PR #375, opened 2026-04-13
feat: cache compiled policy on GPU to skip reload on consecutive runs

aryanmangal769:
## Summary

Every learned policy currently takes ~1 minute to load because of `torch.load()`, `torch.compile()`, and the warm-up inference passes that follow. This happens on **every single execution** because the `_execute_behavior` finally block calls `_release_policy()`, which deletes the compiled model and frees GPU memory after each run.

We participated in RoboHacks and discovered that when a user runs the same learned policy twice consecutively, they have to wait for the full ~1 minute load+compile cycle again — even though the exact same checkpoint was just loaded and compiled moments ago.

This change caches the last loaded policy on GPU by tracking the checkpoint path and action dimension. When the same policy is requested again, the system reuses the already-compiled model (only resetting its internal state), skipping the entire load+compile+warmup pipeline. If a *different* policy is requested, the old one is released and the new one is loaded normally.

## Changes

- Added `_cached_checkpoint_path` and `_cached_action_dim` fields to track the currently loaded policy
- Removed `_release_policy()` call from the `_execute_behavior` finally block so the compiled model persists between runs
- Added a cache-hit check at the top of `_load_policy_for_behavior` that reuses the existing model when checkpoint and action_dim match
- Cache is properly invalidated in `_release_policy()` when the policy is explicitly released

No other files need to be modified — the caching is fully self-contained within `manipulation_server.py`.

### DIFF
--- a/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
+++ b/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
@@ -112,8 +112,12 @@ class ManipulationServer(Node):
         self.execution_running = False
         self.current_goal_handle = None
         self.current_policy = None
-        self.current_action_dim = 10  # Add this to track current policy's action dimension
+        self.current_action_dim = 10
         self._cancel_requested = threading.Event()
+        
+        # Policy cache: skip reload+compile when the same checkpoint is requested
+        self._cached_checkpoint_path = None
+        self._cached_action_dim = None
         
         # Sensor data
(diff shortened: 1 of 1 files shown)


theo-michel:
Thank you! Will take a look

axelpey:
Closing the stale implementation: the policy-loading runtime has substantially changed, and this version retains unresolved cache reset and GPU cleanup failure modes. Re-cut from current main if policy caching is still needed.

PR #377, opened 2026-04-21
Vig/stream hdf5 recorder

vignesh-anand:


karmanyaahm (commented):
I don't notice anything crazy skimming through this but you should probably also delete the equivalent code in the training client in the same pr? other than that it lgtm

theo-michel (commented):
LGTM, haven't tried it on the machine. Curious to see if this has any performance penalties, if it has there are ways to make the writing on disk more efficient!

vignesh-anand:
> LGTM, haven't tried it on the machine. Curious to see if this has any performance penalties, if it has there are ways to make the writing on disk more efficient!


ram much much better, cpu neglegible (2%)

PR #378, opened 2026-04-21
systemd(ros-app): wait for clock sync and DNS before launching ROS

vignesh-anand:
Problem
-------
On cold boots the ros-app.service launches the ROS stack immediately after network-online.target, which on this Jetson is not enough to prevent two distinct crashes during OIDC auth in the training node:

1. DNS not ready: OIDC discovery against auth-v1.innate.bot fails with URLError (Errno -3, EAI_AGAIN) because the resolver isn't usable yet when the unit starts.

2. Clock not ready: this carrier board has no RTC coin cell, so every cold boot starts at 1970-01-01. systemd-timesyncd currently takes ~3m24s to complete its first sync, and until then TLS to auth-v1.innate.bot fails with "certificate is not yet valid" because the cert's notBefore is in the future relative to epoch.

Both surface as AuthError in auth_client.provider._discover_token_endpoint and hard-crash TrainingNode.__init__.

Fix
---
Add two ExecStartPre gates to ros-app.service, self-contained so no other units (systemd-time-wait-sync, NetworkManager-wait-online, etc.) need to be enabled:

* Poll `timedatectl show -p NTPSynchronized --value` every 2s for up to 240s, proceed once it reports "yes".
* Poll `getent hosts auth-v1.innate.bot` every 2s for up to 120s, proceed once DNS resolves.

Both gates fail-open (exit 0 on timeout) so a genuinely offline robot still starts and surfaces the real error in logs rather than leaving the unit stuck in `activating` forever. Progress is logged to journald ("clock synced after Xs" / "DNS ready after Xs") for easy post-mortem.

TimeoutStartSec is raised to 600s to accommodate the worst-case wait (240s clock + 120s DNS + ROS launch) without systemd marking the unit failed mid-wait.

Notes
-----
- Does not touch any other systemd unit or install any packages.
- Does not change the [Unit] ordering; After=network-online.target and Requires=zenoh-router.service are unchanged.
- Longer-term mitigations (fake-hwclock, chrony, app-level auth retry) are complementary and can be layered on top without reverting this.

Made-with: Cursor

### DIFF
--- a/ros2_ws/src/cloud/clients/auth-client/auth_client/provider.py
+++ b/ros2_ws/src/cloud/clients/auth-client/auth_client/provider.py
@@ -24,9 +24,11 @@ import base64
 import json
 import logging
 import threading
+import time
 import urllib.request
 import urllib.error
 from datetime import datetime, timezone
+from typing import Callable
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


karmanyaahm (changes requested):
We shouldn't be blocking the whole ros app for the network? That could add 120seconds to boot if someone is without internet?

this should probably be in launch_ros_in_tmux.sh in front of the training client or any nodes that need network

PR #381, opened 2026-04-21
Validate behavior_config in manipulation_server with pydantic

vignesh-anand:
## Summary

- New `manipulation/config_validation.py` module with pydantic v2 schemas (`LearnedExecCfg` / `PosesExecCfg` / `ReplayExecCfg`) + a top-level `validate_behavior_config()` that returns a typed, already-parsed behavior config plus the resolved absolute path to the referenced checkpoint / replay file. `BehaviorConfigError` carries a human-readable `execution.<field>: <reason>` message.
- `manipulation_server` now validates twice:
  - **accept-time** in a new `goal_callback` -> `GoalResponse.REJECT` so clients get immediate feedback instead of "accepted, then aborted".
  - **execute-time** in `execute_behavior_callback` as belt-and-braces against a file being deleted between accept and execute.
- `_execute_learned_behavior` / `_execute_poses_behavior` / `_execute_replay_behavior` now consume the typed config objects plus pre-resolved file paths. Drops the per-site `.get(...)` calls, the empty-list-as-None coercion, and the redundant `os.path.exists` guards (all moved into the validator).
- New `test/test_config_validation.py` pytest suite with 60 cases covering each schema's happy path and every edge case that motivated this change (bool-as-int, NaN/inf, string-as-number, empty-list-as-None, wrong-length poses, missing files), plus a regression check on the real `skills/wave/metadata.json`.

Net effect: a malformed `metadata.json` now surfaces as `execution.duration: Input should be greater than 0 (got 0)` at goal-accept time, instead of a bare `TypeError` deep inside a running policy loop.

## Test plan

- [x] `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest ros2_ws/src/brain/manipulation/test/ -q` -> 60 passed
- [x] `python3 -m py_compile ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py`
- [x] Compat sweep: parsed every `metadata.json` under `skills/` and `~/skills/` with the new validator. The three hand-authored / activated skills (`tape-test`, `pick_socks`, `wave`) parse cleanly; the remaining six are untrained placeholders with `execution: {}`, which are correctly rejected with `execution.checkpoint: Field required` (these were already broken under old code, just with a worse error message).
- [ ] End-to-end smoke test on the Jetson: activate a run, execute the skill, confirm the action succeeds (QA once the PR is reviewed).
- [ ] Submit a malformed metadata (e.g. `"duration": "120"`, `"start_pose": [0, 0, 0]`) via the action client and confirm `goal_callback` rejects with a clean message.

## Out of scope (tracked as follow-ups)

- Moving validation into `brain_client` so it can reject at skill-loading time.
- Adding `code`-type skill validation (different story: Python modules, not metadata-driven).
- Wiring the validated `n_action_steps` through to `create_act_config` (lands with `vig/policy_params_metadata`; trivial rebase once both merge).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ros2_ws/src/brain/manipulation/manipulation/config_validation.py
+++ b/ros2_ws/src/brain/manipulation/manipulation/config_validation.py
@@ -1,0 +1,329 @@
+"""Validation for per-behavior execution configs stored in a skill's metadata.json.
+
+Parses and validates the ``behavior_config`` payload that
+``manipulation_server`` receives on every ``/behavior/execute`` goal. The
+single source of truth for defaults lives here, so the rest of
+``manipulation_server`` can operate on typed, already-validated config
+objects instead of scattering ``dict.get(key, default)`` calls and ad hoc
+type checks through every execution path.
+
+Contract:
+
+- Happy path: ``validate_behavior_config`` returns a :class:`ValidatedBehavior`
+  whose ``params`` is one of :class:`LearnedExecCfg` / :class:`PosesExecCfg`
+  / :class:`ReplayExecCfg`, and whose ``resolved_path`` (for learned /
+  replay skills) is the absolute path to the file referenced by
+  ``execution.checkpoint`` / ``execution.replay_file``.
+- Any failure (bad JSON, unknown type, missing/wrong-type/out-of-bounds
+  field, missing file on disk) raises :class:`BehaviorConfigError` with a
+  human-readable message prefixed by ``execution.<field>`` where
+  applicable.
+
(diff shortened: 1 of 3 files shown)


vignesh-anand:
Superseded by #380 — the single commit from this branch was cherry-picked onto `vig/policy_params_metadata` so all the work lands in one PR.

PR #383, opened 2026-04-29
Fix Training Manager creating metadata in __pycache__

axelpey:
## Summary
- skip __pycache__ and non-skill directories when Training Manager lists skills, datasets, and runs
- require an existing metadata.json before list endpoints call the locked metadata helper
- clean up Ruff issues in the touched Training Manager files

## Verification
- ruff check ros2_ws/src/cloud/clients/training-manager/training_manager/api/skills.py ros2_ws/src/cloud/clients/training-manager/training_manager/api/datasets.py ros2_ws/src/cloud/clients/training-manager/training_manager/api/training.py
- temp skills-dir repro: list_skills/list_datasets/list_all_runs return the real skill and do not create __pycache__/metadata.json

Fixes INN-489

### DIFF
--- a/ros2_ws/src/cloud/clients/training-manager/training_manager/api/datasets.py
+++ b/ros2_ws/src/cloud/clients/training-manager/training_manager/api/datasets.py
@@ -26,16 +26,14 @@ from training_client.src.skill_manager import (
 )
 from training_client.src.types import ClientConfig, ProgressUpdate
 
+from training_manager.api.skill_paths import iter_skill_dirs, skills_dir
+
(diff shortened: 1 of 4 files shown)


vignesh-anand (commented):
it looks like we repeat the logic of listing skills thrice, should we make it one thing and call it in the different files?

axelpey:
Addressed in f8cb56eb: extracted the shared skill-directory discovery into training_manager.api.skill_paths and switched the skills, datasets, and training endpoints to use it instead of repeating the same directory walk/filter.

vignesh-anand (approved):

PR #387, opened 2026-05-04
fix(skills_action_server): harden against bad metadata and runtime errors

vignesh-anand:
## Summary

A single empty/corrupt `metadata.json` was crashing the `skills_action_server` (SAS) process at startup with `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, which left `/brain/available_skills` un-published and the brain client warning `No primitives received from /brain/available_skills after 25s`.

This PR adds the missing exception guards and fixes the underlying race that was producing empty metadata files.

### 1. `_load_physical_skills` — the actual crash
Wrapped each per-directory body in `try/except` (`JSONDecodeError`, `OSError`, catch-all). Also rejects non-dict top-level JSON so a `null` or `[…]` payload doesn't `AttributeError` later. Inverted the nested `if`s into early `continue`s for readability.

### 2. `_publish_skills_list` — one bad skill must not silence the topic
Each of the three iteration loops (code / ready physical / in-training) now has its own `try/except`. Extracted three helpers:
- `_inspect_skill_inputs` — best-effort introspection, falls back to `{}` on signature failure
- `_safe_skill_string` — guarded `guidelines()` / `guidelines_when_running()` calls
- `_build_physical_skill_info` — guarded build path for physical skills

The final `publish()` is also wrapped, so a serialization failure can't take down the publisher.

### 3. `_execute_physical_skill` — missing `except` clause
The function had `try: … finally:` with no `except`. Any exception (`json.dumps` on weird metadata, behavior-server disconnect, executor errors) bubbled into the action executor and left the goal in an undefined state. Added a catch-all that logs, calls `goal_handle.abort()` (also guarded), and returns a `FAILURE` result.

### 4. `_handle_create_physical_skill` — atomic write (root-cause fix)
The previous `with open(metadata_path, "w") as f: json.dump(...)` truncates the file immediately, so the hot-reload watcher / a concurrent reload could read a 0-byte file. Switched to write-tmp + `flush()` + `fsync()` + `os.replace()`, with cleanup of the tmp on failure. This eliminates the source of the empty `metadata.json` in the first place.

## What's already guarded (left alone)

`execute_callback`, `head_position_callback`, `cancel_callback`, `_reload_physical_skill`, `_handle_reload_skills*`, `_state_update_thread_func`, `_execute_code_skill`, per-skill class instantiation in `__init__` and `_reload_skills`.

## Out of scope

- `skill_loader.py`: already wraps h5py validation and `_get_episode_count` json reads.
- `manipulation_server.py`: separate process, already pydantic-validated since `f1e8be33`.

## Test plan

- [x] Lint clean (`ReadLints`)
- [x] `python3 -m py_compile` clean
- [x] Smoke test of hardened loader against 5 failure modes (empty file, malformed JSON, JSON `null`, JSON list, permission-denied) — all logged and skipped, valid skill still loaded
- [x] Smoke test of atomic write — success path replaces correctly with no leftover tmp; failure path leaves the original file untouched and cleans the tmp
- [ ] On-device: drop a 0-byte `~/skills/broken/metadata.json`, restart SAS, confirm error log + `/brain/available_skills` still publishes
- [ ] On-device: confirm `_handle_create_physical_skill` still creates valid skills end-to-end

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/skills_action_server.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/skills_action_server.py
@@ -232,6 +232,116 @@ class SkillsActionServer(Node):
         msg.directory = directory or ""
         return msg
 
+    def _inspect_skill_inputs(self, skill_id: str, skill_instance) -> dict:
+        """Best-effort introspection of a code skill's execute() signature.
+
+        Returns ``{param_name: type_str}``. Any failure (no execute method,
+        unsupported signature, weird annotations) is logged and an empty
+        dict is returned so the skill can still be published.
+        """
+        if not hasattr(skill_instance, "execute"):
+            return {}
+
+        try:
+            signature = inspect.signature(skill_instance.execute)
+        except (TypeError, ValueError) as e:
+            self.get_logger().warn(
+                f"Could not inspect execute() signature for '{skill_id}': {e}"
+            )
+            return {}
+
+        inputs: dict = {}
+        for param_name, param in signature.parameters.items():
+            if param_name == "self":
+                continue
+            param_type = "any"
+            try:
+                if param.annotation != inspect.Parameter.empty:
+                    if (
+                        isinstance(param.annotation, (types.UnionType, types.GenericAlias))
+                        or hasattr(param.annotation, "_name")
+                        and param.annotation._name in ["List", "Optional", "Dict", "Tuple", "Union"]
+                    ):
+                        param_type = str(param.annotation)
+                    elif hasattr(param.annotation, "__name__"):
+                        param_type = param.annotation.__name__
+                    else:
+                        param_type = str(param.annotation)
+                    param_type = param_type.replace("typing.", "")
+            except Exception as e:
+                self.get_logger().warn(
+                    f"Could not stringify annotation for '{skill_id}.{param_name}': {e}"
+                )
+                param_type = "any"
+            inputs[param_name] = param_type
+        return inputs
+
+    def _safe_skill_string(self, skill_id: str, skill_instance, attr: str) -> str:
+        """Call ``skill_instance.<attr>()`` with full error isolation.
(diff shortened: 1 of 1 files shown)


vignesh-anand:
## On-device verification — both checks pass

Tested on-robot against PID 82894 running the rebuilt `brain_client` install.

### Test 1: bad metadata.json no longer crashes startup

Planted `~/skills/_test_bad_metadata/metadata.json` (0 bytes) and started the SAS. Log excerpt:

```
[INFO] Successfully loaded 13 code skills
[INFO] Loaded physical skill: innate-os/wave (type: replay)
...
[ERROR] Skipped /home/jetson1/skills/_test_bad_metadata/metadata.json: invalid JSON (Expecting value: line 1 column 1 (char 0))
[INFO] Successfully loaded 2 physical skills
[INFO] Found 12 in-training skills
[INFO] Total skills available: 15
[INFO] Published 27 skills on /brain/available_skills
```

The bad file is logged and skipped; every other skill loads; `/brain/available_skills` publishes — exactly the failure mode that was crashing SAS before. `ros2 topic echo --once /brain/available_skills` returned the full skill list.

### Test 2: atomic write of metadata.json

Called `/brain/create_physical_skill` with `name: "Atomic Write Test"`:

- Result: `success=True`, `skill_directory=/home/jetson1/skills/atomic-write-test`
- File written: 306 bytes, valid JSON, top-level dict with all expected keys
- No `.tmp` leftover in the directory
- Hot reload picked up the new skill and republished 28 skills (and again skipped the bad metadata file gracefully)

### Cleanup

Removed both test directories, triggered `/brain/reload_primitives`: response `Reloaded 13 code, 2 physical skills`. Stack is back to its baseline.

All four PR-body checkboxes can be marked done.

PR #396, opened 2026-05-20
Prompt for hosted service key during sim setup

axelpey:
## Summary
- add a setup-time hosted brain credential prompt for missing/placeholder INNATE_SERVICE_KEY
- add a top-level ./innate setup alias for the simulator setup path
- allow INNATE_SERVICE_KEY from the current shell to override .env without persisting unless the user confirms

## Tests
- python3 -m py_compile dev/launcher/*.py
- ./innate --help
- ./innate setup --help
- ./innate sim setup --help
- setup_wizard helper checks via PYTHONPATH=dev/launcher python3

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -2,6 +2,7 @@
 # Keep non-secret runtime settings in config/os.toml
 # and simulator settings in sim/config.toml.
 
+# Filled by `./innate setup` / `./innate sim setup` when you use hosted brain mode.
 INNATE_SERVICE_KEY=your_service_key_here
 
 # Explore other voices at https://play.cartesia.ai
--- a/dev/launcher/README.md
+++ b/dev/launcher/README.md
@@ -31,12 +31,12 @@ innate-os/
 
 ```bash
 cd innate-os
-./innate sim setup
+./innate setup
 ./innate sim up
 ```
 
 If any local config file does not exist yet, the CLI creates it from its template automatically.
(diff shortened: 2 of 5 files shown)


axelpey:
Addressed the Greptile comments in `e46a39c8`:

- `write_env_value` now ignores commented key examples, replaces the first active assignment, and drops duplicate active assignments so last-value-wins parsing cannot keep a stale placeholder.
- `is_configured_secret` now rejects obvious non-keys shorter than 16 chars.
- Shell-provided `INNATE_SERVICE_KEY` now updates both `raw_env` and `user_env` for the current run even when it is not persisted.

Validation:
- `python3 -m py_compile dev/launcher/setup_wizard.py dev/launcher/config.py dev/launcher/main.py`
- Targeted temp `.env` script covering commented key + duplicate active placeholders, placeholder/min-length detection, and non-interactive shell-key config updates.
- `git diff --check`

axelpey:
The remaining 4/5 was because Greptile found one more P1 after the previous fixes: shell injection in `get_config()` accepted placeholder values like `INNATE_SERVICE_KEY=your_service_key_here`, which could shadow a real `.env` key in `raw_env` while the wizard saw `user_env` as valid and returned early.

Addressed in `32de3349`:
- Secret validation is now centralized in `config.is_configured_secret_value`.
- `get_config()` only lets shell secret values override `.env` when they pass the same placeholder/min-length validation as the setup wizard.
- The setup wizard now uses that shared validation helper, so the two paths stay in sync.

Validation added to the targeted script for the exact shadow case: real `.env` key + exported placeholder keeps the real `.env` key in `raw_env`. Pycompile and `git diff --check` also pass.

PR #405, opened 2026-05-26
Require public simulator asset pulls

axelpey:
## Summary

This makes the simulator asset package contract explicit: the locked asset image must be publicly pullable from GHCR so a fresh developer can run `./innate setup` without logging into Docker/GHCR.

## Changes

- Removes the GHCR login step from the simulator asset validation workflow.
- Drops the workflow `packages: read` permission, so CI validates the same anonymous pull path a new developer uses.
- Adds a launcher-side GHCR manifest check before `docker pull`, with a clear error if the locked asset image is private, missing, or unreachable because of a network issue.

## Why

Theo hit this during first-run setup:

```text
Pulling simulator asset pack failed.
unauthorized
```

That happened because the current `innate-os-sim-assets` package is private, while our onboarding UX assumes `git clone && ./innate setup` works without a GHCR login step.

## Package Setting

`ghcr.io/innate-inc/innate-os-sim-assets` is now public, so fresh simulator setup can pull the locked asset pack without Docker/GHCR login.

GitHub notes that public Container registry packages can be pulled anonymously, and that making a package public is not reversible back to private.

## Validation

- `python3 -m py_compile sim/launcher/assets.py`
- helper URL parsing smoke test for tag and digest refs
- `git diff --check`
- `./innate sim assets validate --ci` passes locally against the public GHCR asset package.

Additional validation after making the package public:

- Anonymous GHCR manifest request succeeds with an anonymous bearer token.
- `./innate sim assets validate --ci` passes locally.
- `python3 -m py_compile sim/launcher/assets.py`.
- GHCR network failures are caught and surfaced as `StackError` messages instead of raw Python tracebacks.

### DIFF
--- a/.github/workflows/sim-assets.yml
+++ b/.github/workflows/sim-assets.yml
@@ -14,16 +14,12 @@ on:
 
 permissions:
   contents: read
-  packages: read
 
 jobs:
   validate-assets:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-
-      - name: Log in to GHCR
-        run: echo "${{ github.token }}" | docker login ghcr.io -u "${{ github.actor }}" --password-stdin
 
       - name: Validate simulator assets
         run: ./innate sim assets validate --ci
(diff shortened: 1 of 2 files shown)


axelpey:
Closing: the actual setup issue was fixed by making the GHCR simulator asset package public, so this code-side enforcement PR is not needed.

PR #407, opened 2026-05-26
fix: wait briefly for learned behavior sensors

axelpey:
Replacement for #374 with the sensor retry timeout tuned to 2 seconds.

Includes Aryan's original fix for retrying sensor availability after learned-behavior subscriptions are created, plus a follow-up commit changing the wait deadline/log from 5s to 2s.

Verification:
- git diff --check
- python3 -m py_compile ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py

### DIFF
--- a/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
+++ b/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
@@ -281,10 +281,14 @@ class ManipulationServer(Node):
             if not self._load_policy_for_behavior(checkpoint_path, action_dim):
                 return "FAILURE", f"Failed to load policy from {checkpoint_path}"
             
-            # Check sensor availability before starting
-            if not self._check_sensor_availability():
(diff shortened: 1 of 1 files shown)


axelpey:
Good to merge. I checked the diff: this is Aryan's sensor-availability retry from #374 with the timeout tuned from 5s to 2s, plus matching log text. Verified with git diff --check and py_compile on the touched file.

axelpey:
CI note: the format workflow is currently failing because it runs ruff format --check ros2_ws/src/ and reports 122 existing files needing formatting. I did not add a repo-wide formatting commit because this PR is intentionally a narrow behavioral fix. The actual code diff is still just the sensor wait loop plus 2s timeout/log text; local checks passed: git diff --check and py_compile on manipulation_server.py.

karmanyaahm:
Hey @aryanmangal769 — we're relicensing innate-os from AGPLv3 to Apache 2.0 (the standard for the ROS ecosystem). Could you confirm you agree to have your contributions relicensed under Apache 2.0? A 👍 or reply on this issue works.

PR #411, opened 2026-05-26
Alternate simulator dashboard UI

axelpey:
NO LONGER WIP; LOOKS READY TO MERGE

## Summary
- make the alternate simulator dashboard the default and only frontend path; the old classic UI and `?ui=classic` fallback are removed
- reduce `App.tsx` to the app-level simulator wiring and delegate the rendered experience to `AlternativeSimDashboard`
- remove classic-only frontend artifacts: old `App.css`, `ToggleViewMode`, mic/voice-input controls, and the unused `groq-sdk` dependency
- keep the new harness model: either the Innate harness is running with `empty_directive`, or the brain is stopped
- show the installed skill catalog on the left, with active-skill toggles that update the running harness through `/set_active_skills` without changing code files
- support manual skill runs from the UI, including parameter forms, running/stopping state, and interaction-log entries for triggered skills
- keep backend URL override support for `innate_service_key` / brain URI params so the UI can update the brain backend at runtime
- use the ROS URDF-backed MARS rendering path, initialize the arm in its rest pose, and keep noisy arm bridge/service-call logs hidden in quiet mode
- remove vendored `rws` from `innate-os`; the ros-websocket service-advertise fix now lives in the upstream `ros-websocket` PR/package path

## Validation
- `cd sim/frontend && yarn build`
- `cd sim/frontend && yarn lint` (passes with the existing `Chat.tsx` hook dependency warning)
- `git diff --check`
- simulator was run locally during the PR to verify the new UI can connect with a URL-provided service key, start/stop the harness, toggle skills, run manual skills, and display camera/chat/skill-log updates

## Notes
- This PR is still draft/WIP while we continue validating simulator behavior and the external `rws` package path.
- After the updated `rws` package is available through apt, a fresh checkout should get the service-advertise behavior without vendoring `rws` in this repo.

## Remaining before ready

- Fix the simulator camera placement/frustum so the main camera does not clip through or render from inside the arm. This is the last known product issue before the PR should move out of WIP/draft.


axelpey:
Update: replaced the temporary Python rosbridge 9091 workaround with a vendored patched RWS package.

Why: the issue was in our RWS implementation: it did not support rosbridge `advertise_service` / `service_response`, so simulator-owned arm services could not be materialized into the ROS graph through RWS.

What changed here:
- Added `ros2_ws/src/rws` temporarily from innate-inc/ros-websocket branch `codex/advertise-service-support`.
- Removed the extra `sim-services` tmux window and port `9091`.
- Simulator now advertises arm services on the normal `ws://localhost:9090` RWS data connection.

Validation:
- `./innate sim down && ./innate sim up --once` boots with only port 9090 exposed.
- `ros2 service list` sees all six arm services: `goto_js`, `goto_js_v2`, `goto_js_trajectory`, `torque_on`, `torque_off`, `reboot`.
- `ros2 service call /mars/arm/torque_on std_srvs/srv/Trigger '{}'` returns success.
- `ros2 service call /mars/arm/goto_js_v2 maurice_msgs/srv/GotoJS ...` returns success.

Upstream RWS PR: https://github.com/innate-inc/ros-websocket/pull/1

axelpey:
Follow-up cleanup: removed the vendored `ros2_ws/src/rws` source from this PR. The RWS implementation belongs upstream in https://github.com/innate-inc/ros-websocket/pull/1 and should come into `innate-os` through the apt package once released.

This PR keeps the simulator-side single-port RWS wiring and quiet-log cleanup, but no longer carries RWS source code.

ros2_ws/src/brain/brain_client/brain_client/robot/manipulation.py
```diff
@@ -98,14 +99,23 @@ def stop(self):
         self._arm_state = None
 
     def spin_node_to_refresh_topics(self, count: int = 10, timeout_sec: float = 0.001):
-        """Spin the underlying ROS node to process pending callbacks.
+        """Yield briefly while the node's executor processes callbacks.
 
-        Args:
-            count: Number of spin iterations.
-            timeout_sec: Timeout per iteration.
+        The skills action server already owns this node in a MultiThreadedExecutor.
```
theo-michel: **Deadlock: every IK solve / motion service call inside an action-triggered skill times out.**

This docstring's premise is false: no `MultiThreadedExecutor` exists anywhere in brain_client (grep finds only `SingleThreadedExecutor` in `hot_reload.py` and `camera_provider.py`). `skills_server.py` `main()` spins single-threaded (`while rclpy.ok(): rclpy.spin_once(action_server, timeout_sec=1.0)`), and `_execute_code_skill` calls `skill.execute()` synchronously inside the action `execute_callback` — on the one spinning thread. `ReentrantCallbackGroup` permits interleaving but adds no threads.

The old `rclpy.spin_once`/`spin_until_future_complete` calls this PR removed were the only callback pump for that path. With passive `time.sleep`/`Event` waits, `/mars/arm/ik_solution` messages and service-response futures can only be delivered by the executor thread that is blocked inside `skill.execute()` — so `solve_ik` waits out its full timeout and returns None, and `goto_js`/`torque_on` futures never complete.

CLI-submitted skills still work because they run on `_cli_skill_worker`, leaving the main thread free to spin — which is why local testing wouldn't catch this. Fix options: run the skills server in an actual `MultiThreadedExecutor`, or execute skills on a worker thread like the CLI path does.

ros2_ws/src/brain/brain_client/brain_client/skills/registration.py
```diff
@@ -79,7 +79,8 @@ def register(self) -> None:
             return
         directive = self._state.current_directive
         primitives = self._state.registry.metadata
-        included = [p for p in primitives if p["id"] in directive.get_skills()]
+        active_skill_ids = set(self.active_skill_ids_for_registration())
+        included = [p for p in primitives if p["id"] in active_skill_ids]
```
theo-michel, line 83: **Registering with zero active skills sends `"primitives": null` and crashes the cloud agent's registration handler.**

With the new harness model this is now a default path: `empty_directive.get_skills()` returns `[]`, so `included == []` on every harness start (and whenever the last skill is toggled off via `/brain/set_active_skills`). Line 88 below (`"primitives": included if included else None`) then sends `None`, and pydantic's `model_dump_json` emits `"primitives": null`.

On the cloud side (`innate-cloud-agent/src/brain.py:623`), `message.payload.get("primitives", [])` returns `None` because the key is *present*, `state.primitives_list.clear()` runs, then `for p in primitives_data:` raises `TypeError`. The exception is swallowed by the dispatch loop, so no `PRIMITIVES_AND_DIRECTIVE_REGISTERED` ack is ever sent and the robot's registered skill set silently stays stale (with the cloud-side primitive list wiped).

Fix: send `included` (an empty list) instead of `None` — the `included if included else None` falsy-empty-list pattern is the bug.

ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
```diff
@@ -347,7 +476,16 @@ def _svc_get_directives(self, request, response):
...
                 }
             )
-        response.directives = [json.dumps(details)]
+        response.directives = [
+            json.dumps(
+                {
+                    "agents": details,
+                    "skills": self.state.registry.metadata,
+                    "active_skills": self.catalog.active_skill_ids_for_registration(),
+                    "brain_active": self.state.is_brain_active,
```
theo-michel, line 443: **Breaking change for the mobile controller app: `GetAvailableDirectives` response changed from a JSON array to a JSON object.**

`innate-controller-app/src/contexts/RobotCoreContext.tsx:191` does `const directives: DirectiveInfo[] = JSON.parse(result.directives[0])` with no shape check, and downstream callers use array methods — `AgentScreen.tsx:192` `directives.find(...)`, `HomeScreen.tsx:142` `directives.map(...)`, `FloatingBehaviorIndicator.tsx:106` `directives.find(...)` — which throw `TypeError` on a plain object. The agent-selection screens break on every robot running this brain version.

The PR updates the *sim* consumer to handle both shapes (dict and list) but ships no compat for the app. Either keep `directives[0]` as the legacy array and move the new fields elsewhere (e.g. additional response entries), or land a tolerant parser in the app first.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -179,9 +429,30 @@ async def inbound_data_loop(ws, shared_queues):
...
+                        task_status=payload.get("taskStatus") or payload.get("task_status"),
+                        primitive_id=payload.get("primitiveId") or payload.get("primitive_id"),
+                        skill_id=payload.get("skillId") or payload.get("skill_id"),
+                        failure_reason=payload.get("failureReason") or payload.get("reason"),
+                    )
+                    # Forward to sim/UI without allowing stale history to pile up.
+                    shared_queues.enqueue_chat_from_bridge(chat_msg)
+
+            elif topic == "/brain/skill_status_update":
+                payload = json.loads(msg_data.get("data", ""))
```
theo-michel: **Unguarded `json.loads` — one malformed message tears down the entire rosbridge data connection.**

A `std_msgs/String` on `/brain/skill_status_update` with empty or non-JSON `data` (`json.loads("")` also throws) raises `JSONDecodeError`; the topic-dispatch block has no per-message guard, so the exception propagates out of `inbound_data_loop`, the `asyncio.wait(FIRST_COMPLETED)` supervisor cancels `outbound_data_loop`, and the websocket closes — camera, `/cmd_vel`, odom, and chat all drop until the ~2 s reconnect. A publisher repeatedly sending bad payloads keeps the bridge in a reconnect loop, and the failure is silent (the task exception is never retrieved; only the generic "Connection lost. Retrying..." logs).

The adjacent `/brain/websocket_status` branch wraps its `json.loads` in try/except — same pattern needed here (the pre-existing `/brain/chat_out` branch at ~line 422 has the same flaw, worth fixing while here).

theo-michel (commented):
Automated deep review (7 finder angles, per-finding verification). 8 verified issues below, ranked by severity. Two findings from the same review were already addressed by the latest commits (manual skill runs proxied through the sim backend, `brain_active` reported in the directives payload) — nice.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -579,7 +907,16 @@ async def outbound_data_loop(ws, shared_queues, service_call_queue):
                 directive_msg = {"data": msg.directive}
                 outbound = rosbridge_publish("/brain/set_directive", directive_msg)
                 await ws.send(json.dumps(outbound))
+                shared_queues.set_current_agent(msg.directive)
                 print(f"[ROSBridge] Published directive: {msg.directive}")
+            elif isinstance(msg, ActiveSkillsCmd):
+                active_skills_msg = {"data": json.dumps({"agent_id": msg.agent_id, "skills": msg.skills})}
+                outbound = rosbridge_publish("/brain/set_active_skills", active_skills_msg)
+                await ws.send(json.dumps(outbound))
+                shared_queues.update_active_skill_ids(msg.skills)
```
theo-michel, line 991: **Skill toggles silently desync: the requested list is recorded as truth before the brain accepts it.**

This records `msg.skills` into `shared_queues` immediately after publishing, but `_on_set_active_skills` (brain_client_node.py:294-299) drops the update with only a warn log when `agent_id` doesn't match its current directive — and the frontend always sends `agent_id="empty_directive"` (App.tsx:608). It's a one-way topic with no ack/nack, and nothing self-corrects: the `/brain/available_skills` subscription updates only the catalog, not `active_skill_ids`.

So if the brain's directive is changed out-of-band (controller app publishing `/brain/set_directive`, etc.), the UI and `/available_agents` both show the toggle applied while the cloud agent never registered the skill — no error anywhere.

Suggested fix: have the brain publish its accepted active-skill set (e.g. include it in the existing `/brain/available_skills` payload or a status topic) and treat that as the source of truth here, instead of echoing the request.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -648,6 +986,28 @@ async def outbound_data_loop(ws, shared_queues, service_call_queue):
             # no messages to publish right now
             pass
 
+        clock_msg, arm_state_msg, nav_feedback_msg = shared_queues.pop_latest_agent_updates()
```
theo-michel, line 1064: **Throttled pop silently discards messages — including the final nav feedback of a goal — and hammers the sim thread's lock ~10k times/sec.**

`pop_latest_agent_updates()` takes the lock and clears all three slots on every iteration of this loop (which sleeps 0.0001 s), but each popped message is only published if its 0.05–0.1 s throttle window has elapsed; otherwise the local variable is simply dropped — there's no code path that re-stores an unpublished message.

Concrete loss: `NavigationController._publish_feedback` produces at 10 Hz and stops on SUCCEEDED/CANCELED, so the final `distance_to_goal` is frequently popped inside a closed 0.1 s window and dropped with no successor — `/sim_navigation/feedback` consumers see roughly half the updates and a stale final distance. Secondary cost: ~10,000 cross-thread lock acquisitions/sec for at most ~60 useful messages.

Fix: check the per-topic interval deadlines first and only pop (take the lock) when at least one window is open, retaining unpublished messages.

sim/frontend/src/services/rosbridgeService.ts
```diff
@@ -209,33 +561,38 @@ export async function getAvailableAgentsDirect(
...
+    parsedPayload !== null && Array.isArray(parsedPayload.active_skills);
+  const activeSkillIds = parseSkillIds(parsedPayload?.active_skills);
+  const brainActive =
+    typeof parsedPayload?.brain_active === "boolean"
+      ? parsedPayload.brain_active
+      : typeof values.brain_active === "boolean"
+        ? values.brain_active
+        : undefined;
+  try {
+    skills = await getAvailableSkillsFromTopicDirect(wsUrl);
```
theo-michel: **Topic read unconditionally overwrites the skills already parsed from the service payload, and the helper has no `onclose`.**

Line 573 parses skills from the `GetAvailableDirectives` payload, then this line replaces them with a fresh `/brain/available_skills` latched-topic read with no guard — a successful subscribe that yields an empty/stale list (e.g. last published with 0 skills right after a registry reload) blanks a non-empty payload list, and the dashboard renders no skill toggles. `getAvailableSkillsFromTopicDirect` also defines `onopen`/`onmessage`/`onerror` but no `onclose`, so a cleanly-closed socket stalls every agent fetch for the full 1200 ms timeout. And each fetch pays an extra WebSocket handshake even when the payload already had the data.

Fix: only fall back to the topic when `parsedPayload.skills` was empty, and reject on `onclose`.

sim/frontend/src/App.tsx
```diff
@@ -1310,44 +486,245 @@ export default function App() {
...
+      : "enable";
 
-      const data = await response.json();
-      console.log("Directive response:", data);
+    setPendingSkillChanges({ [skillId]: action });
+    setSkillUpdateError(null);
+    try {
+      await handleSetActiveSkills(EMPTY_DIRECTIVE_ID, nextSkillIds);
+      setActiveSkillIds(nextSkillIds);
+      window.setTimeout(() => void fetchAgents(), 150);
```
theo-michel, line 610: **The 150 ms refetch can visibly revert the user's toggle.**

`POST /set_active_skills` just enqueues `ActiveSkillsCmd` and returns; `shared_queues.active_skill_ids` is only updated after the bridge's outbound loop dequeues and publishes it (one message per iteration from a queue shared with occupancy grids and arm state, and not at all during reconnect windows). This delayed `fetchAgents()` then unconditionally overwrites `activeSkillIds` from that cache (the `pendingSkillChanges` guard is already cleared in `finally` before this timeout fires) — so a busy or reconnecting bridge flips the toggle back in the UI even though the command later goes through.

Fix: drop the refetch (the optimistic update plus the existing error rollback covers it), or guard `fetchAgents` against in-flight toggles instead of clearing the guard before the timer fires.

ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
```diff
@@ -359,7 +488,16 @@ def _svc_get_directives(self, request, response):
                     "source": getattr(directive, "source", "user"),
                 }
             )
-        response.directives = [json.dumps(details)]
+        response.directives = [
```
theo-michel, line 437: **[Critical] `get_available_directives` response shape change breaks the mobile controller app**

`_svc_get_directives` now returns `directives[0]` as a JSON **object** `{agents, skills, active_skills, brain_active}` instead of the previous bare JSON **array** of agents. The sim frontend was updated to accept both shapes, but the React Native controller app (`innate-controller-app`) was not: `RobotCoreContext.tsx` does `JSON.parse(result.directives[0])` typed as `DirectiveInfo[]` and `HomeScreen.tsx` calls `directives.map(...)`. After this merges, the Home screen throws `directives.map is not a function` on every robot and the agent list fails to load.

**Fix:** ship a controller-app compatibility shim (`const parsed = JSON.parse(...); const agents = parsed.agents ?? parsed;`) in lockstep with this PR.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -116,11 +164,235 @@ def rosbridge_service_response(service: str, result: dict, call_id: str = None)
...
+            await _send_service_response(ws, service_name, {"success": False}, call_id)
+            return
+
+        parsed = _parse_arm_trajectory_args(args)
+        if parsed is None:
+            await _send_service_response(ws, service_name, {"success": False}, call_id)
+            return
+
+        waypoints, durations = parsed
+        asyncio.create_task(
```
theo-michel: **[Critical] Fire-and-forget `create_task` can be garbage-collected mid-trajectory**

`_run_arm_trajectory_service(...)` is launched via `asyncio.create_task(...)` but the returned Task is never stored (the other `create_task` sites keep theirs in the `tasks` list). The event loop holds only a *weak* reference, so the pending task can be GC'd between waypoints (`await asyncio.sleep(duration)`), leaving the arm stopped partway through a trajectory with no error logged.

**Fix:** keep a strong reference, e.g. `self._bg_tasks.add(t); t.add_done_callback(self._bg_tasks.discard)`.

ros2_ws/src/brain/brain_client/brain_client/skills/registration.py
```diff
@@ -79,7 +79,8 @@ def register(self) -> None:
             return
         directive = self._state.current_directive
         primitives = self._state.registry.metadata
-        included = [p for p in primitives if p["id"] in directive.get_skills()]
+        active_skill_ids = set(self.active_skill_ids_for_registration())
```
theo-michel, line 82: **[High] Active-skill registration is no longer scoped to the current directive**

`register()` dropped the old `included = [p for p in primitives if p["id"] in directive.get_skills()]` guard; `active_skill_ids_for_registration()` / `set_active_skill_ids()` now validate only against *all* globally-available skills, and `_on_set_active_skills` checks only the agent id, not the directive. Publishing `/brain/set_active_skills {agent_id: A, skills: ["skill_b"]}` where `skill_b` belongs to a different directive registers it under directive A, widening A's capability surface beyond what `get_skills()` declares.

**Fix:** intersect the requested/active skill ids with `current_directive.get_skills()` before registration.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -778,9 +1167,10 @@ async def rosbridge_loop(shared_queues, rosbridge_uri: str, retry_interval: floa
     service_task = asyncio.create_task(
         _service_connection_loop(shared_queues, rosbridge_uri, service_call_queue, retry_interval)
     )
+    tasks = [data_task, service_task]
 
     # If either loop exits (e.g. exit_event set), cancel the other
-    done, pending = await asyncio.wait([data_task, service_task], return_when=asyncio.FIRST_COMPLETED)
+    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
```
theo-michel, line 1219: **[High] Pending connection-loop tasks are cancelled but never awaited**

After `asyncio.wait(..., FIRST_COMPLETED)` the pending task gets `task.cancel()` (here and in the data/service loops at ~1109 / ~1138) but is never awaited. When the `async with websockets.connect` block exits and reconnects, fresh inbound/outbound loops are spawned while the not-yet-finalized old loop is still running. Two `outbound_data_loop`s can briefly overlap and race `shared_queues.nav_controller = NavigationController(...)`.

**Fix:** `await asyncio.gather(*pending, return_exceptions=True)` after cancelling, at all three sites.

sim/frontend/src/components/AlternativeSimDashboard.tsx
```diff
@@ -0,0 +1,1570 @@
...
+        ...previous,
+        [skill.id]: { text: message || "Skill finished." },
+      }));
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      setSkillRunMessages((previous) => ({
+        ...previous,
+        [skill.id]: { text: message, error: true },
+      }));
+    } finally {
```
theo-michel, line 1168: **[High] Stop-skill can wedge the UI for 5 minutes and block all other skills**

`executingSkillId` / `runningSkillCancel` are cleared only in this `finally`, after `await run.promise`. In `rosbridgeService.ts` the action promise settles only on `op === "action_result"`; a cancel acknowledged by any other frame leaves it pending until `DEFAULT_ACTION_TIMEOUT_MS = 300000`. So a skill whose cancel yields no terminal `action_result` shows "Stopping…" and disables RUN on every other skill (`otherSkillExecuting`) for up to 5 minutes. The sim/HTTP path is weaker still — cancel never touches the original `/execute_skill` fetch promise.

**Fix:** settle/reject the promise on cancel-ack and on a short cancel timeout; clear executing state independently of the action promise.

sim/src/shared_queues.py
```diff
@@ -154,31 +184,139 @@ def update_available_agents(
...
+
+    def enqueue_chat_from_bridge(self, msg: ChatMessage) -> bool:
+        timestamp = float(msg.timestamp or 0.0)
+        key = (msg.sender, msg.text, round(timestamp, 3))
+
+        with self.chat_from_bridge_lock:
+            if key in self._recent_chat_from_bridge_key_set:
+                return False
+
+            while self.chat_from_bridge.full():
```
theo-michel, line 294: **[Medium] Chat dedup set/deque desync on queue-full eviction**

The queue-full eviction loop discards the dropped key from `_recent_chat_from_bridge_key_set` but never removes it from the `_recent_chat_from_bridge_keys` deque. The set and deque drift apart, so a later `popleft()` can discard the dedup key of a still-queued live message (opening a duplicate gap), or evicted keys' stale deque slots mis-time membership.

**Fix:** keep the set and deque mutations symmetric on every path (also pop/append the deque on eviction).

sim/src/shared_queues.py
```diff
@@ -154,31 +184,139 @@ def update_available_agents(
...
+            arm_state_msg = self.latest_arm_state_msg
+            nav_feedback_msg = self.latest_nav_feedback_msg
+            self.latest_clock_msg = None
+            self.latest_arm_state_msg = None
+            self.latest_nav_feedback_msg = None
+        return clock_msg, arm_state_msg, nav_feedback_msg
+
+    def enqueue_chat_from_bridge(self, msg: ChatMessage) -> bool:
+        timestamp = float(msg.timestamp or 0.0)
+        key = (msg.sender, msg.text, round(timestamp, 3))
```
theo-michel, line 288: **[Medium] Chat dedup key omits `task_status` → terminal skill status can be dropped**

The key is `(sender, text, round(timestamp, 3))`. Skill lifecycle events all share `sender="task_activated"` + the skill name as text (only `task_status` differs, and it's not in the key). A fast/instant-fail skill emitting `running` then `failed`/`completed` in the same rounded-ms bucket (or history replay with a fixed timestamp) drops the terminal event, so the Interaction Log shows the skill stuck "running".

**Fix:** include `task_status` (and/or `primitive_id`) in the dedup key.

sim/src/simulation/simulation_node.py
```diff
@@ -1776,30 +2052,24 @@ def run(self):
...
-                # Update arm wrist camera (mounted on link5, looking forward along the arm)
-                if self.arm_wrist_camera is not None:
-                    link5 = self.robot.get_link("link5")
-                    link5_pos = link5.get_pos()
-                    link5_quat = link5.get_quat()
-                    arm_forward = np.array([1.0, 0.0, 0.0])  # Arm points along X axis
-                    arm_look_dir = rotate_vector(arm_forward, link5_quat)
-                    arm_lookat = link5_pos.cpu().numpy() + arm_look_dir
-                    self.arm_wrist_camera.set_pose(pos=link5_pos.cpu().numpy(), lookat=arm_lookat)
+                if self.robot_camera is not None and self.main_camera_mount is not None:
```
theo-michel, line 2059: **[Medium] First-person camera silently freezes when URDF camera links don't resolve**

The render loop only updates the camera pose `if self.main_camera_mount is not None`. If `_init_camera_link_refs` fails to resolve any of `camera_optical_frame` / `head_camera_left` / `head_camera_link`, `main_camera_mount` stays `None` with no warning and the camera is frozen at its init pose for the whole session. Given the PR's open "camera clips the arm" note, a hard log/error here would surface misconfiguration instead of a plausible-but-wrong feed.

**Fix:** log a warning/error when no camera frame resolves.

sim/frontend/src/services/rosbridgeService.ts
theo-michel: **[Medium] `getAvailableAgentsDirect` discards payload skills and adds ~1.2s latency per refresh**

The service response already contains the skill list, but the code unconditionally overwrites it by awaiting `getAvailableSkillsFromTopicDirect`, which subscribes to the event-driven (non-latched) `/brain/available_skills` and blocks up to 1200ms. This runs on mount and 150ms after every skill toggle, so each refresh eats ~1.2s and can show empty skills when the topic doesn't re-publish in time.

**Fix:** use the payload skills; only fall back to the topic when they're absent.

sim/src/routes/video_api.py
```diff
@@ -24,8 +33,127 @@ class SetBrainActiveRequest(BaseModel):
...
+        while time.monotonic() < deadline:
+            timeout = max(0.0, deadline - time.monotonic())
+            raw = await asyncio.wait_for(ws.recv(), timeout=timeout)
+            try:
+                message = json.loads(raw)
+            except json.JSONDecodeError:
+                continue
+            if message.get("op") != "action_result" or message.get("id") != call_id:
+                continue
+            status = message.get("status")
```
theo-michel: **[Medium] `execute_rosbridge_action` treats a missing `status` as terminal success**

`status = message.get("status")` → `is_terminal_action_status(None)` returns `True`, and `normalize_action_result(None, …)` matches neither the cancelled(5) nor aborted(6) branch, so a status-less `action_result` frame is reported as a completed, non-failing skill before it actually finished.

**Fix:** require `status` to be present (not None) before treating a result as terminal.

theo-michel (commented):
## Automated code review

Extra-high-effort review of this PR: multi-angle finder pass → per-finding adversarial verification → gap sweep, scoped to lines this PR actually changes. **11 in-scope findings** are posted as inline comments below, ranked by severity. 19 other candidates were checked and refuted (e.g. the all-zero cancel `goal_id` is the ROS "cancel-all" sentinel and works; `manipulation.py`'s event-wait is driven by the main-thread spin loop; the chat-broadcast create-task check has no `await` between check and set so it can't double-spawn).

### ⚠️ Merge-blocker
Finding #1 changes the `get_available_directives` response shape and will break the `innate-controller-app` Home screen on every robot unless the app ships a compatibility shim in the same release.

### Pre-existing / adjacent (real, but in code this PR does not modify — flagging, not blocking)
- **video_api.py — mjpeg frame race:** `mjpeg_generator` reads then nulls `latest_frames[cam]` with no lock; concurrent viewers / the producer race and drop frames. (Streaming code is untouched here.)
- **Chat.tsx — reconnect gives up permanently:** `reconnectAttempts` is reset only in `onopen`, so after `MAX_RECONNECT_ATTEMPTS` failures `scheduleReconnect` returns forever and the socket never recovers without a reload.
- **Chat.tsx — TTS re-fire:** the new-message `useEffect` has deps `[messages]` only, reading a stale `isScrolledToBottom` and re-speaking on non-new-robot updates.
- **Chat.tsx — duplicate messages:** `appendUniqueMessage` dedups on exact float timestamp equality; the **new** `parseSkillStatusMessage` falls back to `Date.now()/1000`, so the same logical message via two paths can append twice.
- **agent_websocket_bridge.py — cancel-without-await also at the data/service loops (~1109 / ~1138):** same pattern as the inline comment; on reconnect these race `shared_queues.nav_controller`.
- **chat_api.py — broadcast drops messages when no client is connected** without clearing their dedup key, so a later identical message in the same rounded-ms bucket is silently deduped away. (Also: the broadcast now forwards `taskStatus/primitiveId/skillId/failureReason` but not `primitiveName`, which `Chat.tsx` reads.)

sim/frontend/src/components/AlternativeSimDashboard.tsx
```diff
@@ -0,0 +1,1570 @@
...
+  values: Record<string, unknown>,
+  schemas: Record<string, SkillInputSchema>,
+) {
+  const parts = Object.entries(values).map(([key, value]) => {
+    const type = getInputType(schemas[key] ?? "string").toLowerCase();
+    if (
+      ["float", "double", "number"].includes(type) &&
+      typeof value === "number"
+    ) {
+      const formatted = Number.isInteger(value) ? `${value}.0` : String(value);
```
theo-michel: **[Medium] `stringifyInputsWithFloatHints` emits invalid JSON for exponential-notation floats**

The float branch does `Number.isInteger(value) ? `${value}.0` : String(value)`. A finite float entered as `1e21` (passes `Number.isFinite` validation) becomes `1e+21.0`, so the brain's `json.loads` throws and the skill silently fails with "Invalid inputs JSON".

**Fix:** build the inputs object and `JSON.stringify` it instead of hand-concatenating the float branch.

ros2_ws/src/brain/brain_client/brain_client/skills/registration.py
axelpey: fixed

ros2_ws/src/brain/brain_client/brain_client/skills/registration.py
axelpey: fixed

theo-michel (commented):
## Automated code review — correctness pass

Reviewed the full diff via multi-angle finders + per-finding verification. Inline comments below mark each issue; the two must-fixes are the **IK regression** (`manipulation.py`) and **cancel-all** (`video_api.py` / `rosbridgeService.ts`). One issue that doesn't anchor to a diff line:

**Unbounded `service_call_queue` replays stale commands on reconnect** (`agent_websocket_bridge.py:1207`) — `asyncio.Queue()` with no `maxsize`, shared across reconnects. While the service connection is down but the data loop runs, every UI reset / brain-active / refresh accumulates and is flushed in order on reconnect, applying outdated commands.

### Also noted (lower severity — efficiency / cleanup / robustness)
- **Service-key validation relaxed to 'non-empty'** (`ws_config.py:25`, `launcher/config.py`, `setup_wizard.py`): placeholder-rejection + min-length checks deleted, so a leftover `your_service_key_here` now counts as configured → opaque auth failure. (Likely intentional given the template now ships empty.)
- **`_send_loop` cadence drift** (`uninavid/ws_client.py`): fixed-rate `create_timer` replaced with `sleep(interval); await send()`, so upload latency adds to the period — effective image rate sags below the configured ~49 Hz under load.
- **`pop_latest_agent_updates` lock acquired ~10 kHz** (`agent_websocket_bridge.py`, loop `sleep(0.0001)`) for data published at ≤20 Hz.
- **Chat.tsx perf**: `appendUniqueMessage` does an O(n) scan + full re-sort of an already-sorted array per message; `messages` grows unbounded; `filteredMessages`/`groupedMessages` recomputed every render with no `useMemo`.
- **video_api opens a fresh rosbridge WebSocket per call** and relies on a 0.12s linger-sleep with no ack to flush a publish — message can be lost on close under load.
- **Arm-trajectory reports success purely from `sleep(duration)`**, not sim confirmation; torque-disabled arm commands are silently dropped / return `{success: False}` with no reason message.
- **Altitude / reuse**: `empty_directive` magic id duplicated across orchestrator / initializer / App.tsx; `ROBOT_ARM_HOME_POSITIONS` + link-set hardcoded in `simulation_node.py` (related to the camera-clip issue); `_quat_wxyz_multiply` hand-rolled while scipy `Rotation` is imported; video_api re-implements the bridge's rosbridge op encoders; `_service_connection_loop` ≈ copy of `_data_connection_loop`.
- **Dead code from the rewrite**: `executeSkillDirect` / `callRosbridgeAction` unused; `episode_count` / `directory` plumbed through 4 files but never read; duplicate view-mode toolbar hidden via CSS; `onSetDirective={() => undefined}` wired to a live submit.
- **Possible base-frame mismatch (unverified)** (`simulation_node.py`): with `SIM_ROBOT_URDF` defaulting to `ros` (no planar base joints), base pose is written to the articulation root but read back from `base_link`; an offset would cause constant odom/nav drift — worth checking against the camera-clip product issue.

The `rws` fresh-checkout concern raised during review was **refuted** — `ros-humble-innate-rws` is in the apt dependency manifests and the readiness probe greps for the running process.

**Verdict:** strong direction, but not ready to merge — the IK regression and cancel-all are must-fix, and several others surface in normal use. The `register()` contract change needs cross-checking against the cloud-agent repo.

_(Note: the trajectory-task GC issue flagged mid-review was already partially fixed in a later commit — acknowledged inline.)_

ros2_ws/src/brain/brain_client/brain_client/robot/manipulation.py
```diff
@@ -98,14 +99,23 @@ def stop(self):
...
+        Calling rclpy.spin_once/spin_until_future_complete here can remove the node
+        from that executor, making later skill goals or cancels hang.
         """
         for _ in range(count):
-            rclpy.spin_once(self.node, timeout_sec=timeout_sec)
+            time.sleep(timeout_sec)
```
theo-michel, line 124: **Bug — IK silently fails for code skills on the agent/action path.** `solve_ik` now busy-waits via `spin_node_to_refresh_topics` → `time.sleep` instead of spinning, with a comment assuming a `MultiThreadedExecutor`. But `skills_server` runs single-threaded `rclpy.spin_once` (`skills_server.py:562`) and `execute_callback` runs on that one thread — the physical-skill path even nested-spins via `spin_until_future_complete` precisely because there is no background executor. So while a code skill runs, the `/ik_solution` subscription callback can never be delivered → `solve_ik` times out and returns `None` → every IK-based `goto_pose` silently fails. The CLI path works (separate worker thread); the agent path does not. **Highest-severity issue in the PR.**

sim/src/routes/video_api.py
```diff
@@ -248,13 +402,99 @@ async def set_brain_active(request: Request, brain_request: SetBrainActiveReques
...
+            {"status": "no_shared_queues", "error": "Simulation not initialized"},
+            status_code=503,
+        )
+
+    try:
+        values = await call_rosbridge_service(
+            "/execute_skill/_action/cancel_goal",
+            {
+                "goal_info": {
+                    "goal_id": {"uuid": [0] * 16},
```
theo-michel: **Bug — cancelling one skill cancels ALL in-flight skills.** This sends the action `cancel_goal` with `uuid = [0]*16` + stamp 0, the ROS2 *cancel-all* sentinel, instead of the goal id generated at start in `execute_rosbridge_action` (which is discarded). When a manual skill overlaps an agent-driven skill, or two run back-to-back, this aborts every in-flight `ExecuteSkill` goal. Mirrored in `rosbridgeService.ts:695` — track and pass the real goal id.

sim/frontend/src/services/rosbridgeService.ts
```diff
@@ -276,3 +672,53 @@ export async function resetBrainDirect(
...
+      skill_type: skillType,
+      inputs: inputsJson,
+    },
+  );
+}
+
+export async function cancelSkillExecutionDirect(wsUrl: string): Promise<void> {
+  await callRosbridgeService(wsUrl, "/execute_skill/_action/cancel_goal", {
+    goal_info: {
+      goal_id: { uuid: Array(16).fill(0) },
```
theo-michel: **Bug (mirror of `video_api.py:482`) — cancel-all sentinel.** `goal_id.uuid = Array(16).fill(0)` is the ROS2 cancel-ALL sentinel. `startSkillExecutionDirect().cancel()` calls both `action.cancel()` (specific goalId) *and* `cancelSkillExecutionDirect` (all-zeros), so cancelling one skill cancels every in-flight goal. Pass the real goalId.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -116,11 +164,272 @@ def rosbridge_service_response(service: str, result: dict, call_id: str = None)
...
+            await _send_service_response(ws, service_name, {"success": False}, call_id)
+            return
+
+        parsed = _parse_arm_trajectory_args(args)
+        if parsed is None:
+            await _send_service_response(ws, service_name, {"success": False}, call_id)
+            return
+
+        waypoints, durations = parsed
+        task = asyncio.create_task(
```
theo-michel, line 395: Good — retaining the task in `shared_queues.arm_trajectory_task` (recent commit) fixes the GC risk. One remaining gap: on data-connection teardown `_data_connection_loop` only cancels `data_task`/`service_task`, not the active trajectory task, so it wakes after `asyncio.sleep` and calls `ws.send` on the closed socket (caught, but then `_send_service_response` runs on the same closed ws). Consider cancelling `arm_trajectory_task` on disconnect too.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -116,11 +164,272 @@ def rosbridge_service_response(service: str, result: dict, call_id: str = None)
...
+            call_id,
+        )
+    except queue.Full:
+        await _send_service_response(ws, service_name, {"success": False}, call_id)
+    except Exception as exc:
+        print(f"[ROSBridge] Arm trajectory service failed: {exc}")
+        await _send_service_response(ws, service_name, {"success": False}, call_id)
+
+
+async def _handle_arm_service_call(ws, shared_queues, service_name, call_id, args):
```
theo-michel, line 356: **Bug — unknown service call hangs the caller.** `_handle_arm_service_call` branches on arm goto / trajectory / trigger services and then falls off the end with no default. A `call_service` op for any non-arm or typo'd service name returns without sending a `service_response`, so the rosbridge caller's promise never resolves until its own timeout. Add an `else` that sends an error response.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -398,8 +783,11 @@ async def inbound_service_loop(ws, shared_queues):
                 shared_queues.update_available_agents(
                     agents=agents,
                     current_agent_id=current_directive,
-                    startup_agent_id=startup_directive,
+                    active_skill_ids=active_skill_ids,
+                    brain_active=brain_active,
                 )
+                if skills:
```
theo-michel, line 790: **Bug — stale skills never clear on Reload.** This `if skills:` guard skips an empty list, so `update_available_skills` is never called with `[]`. The `/brain/available_skills` topic handler clears unconditionally. Delete all skills (or select `empty_directive`) then click *Reload skills* → the UI keeps showing nonexistent skills. The two paths should agree on empty-list semantics.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -365,19 +696,47 @@ async def inbound_service_loop(ws, shared_queues):
...
+                    directives_payload = json.loads(json_string)
+                    if isinstance(directives_payload, dict):
+                        agents_list = directives_payload.get("agents", [])
+                        metadata_payload = directives_payload
+                    else:
+                        agents_list = directives_payload
+                        metadata_payload = {}
+
+                    if len(directive_entries) > 1:
+                        metadata_payload = json.loads(directive_entries[1])
```
theo-michel, line 724: **Bug — a malformed metadata entry wipes the whole agent list.** Agents are parsed from `directive_entries[0]` but appended *after* this `json.loads(directive_entries[1])`. If entry[1] raises `JSONDecodeError`, control jumps to the `except` and `update_available_agents([])` runs even though entry[0] had valid agents → UI shows no agents/skills. Parse/guard the metadata entry independently.

ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
```diff
@@ -283,6 +289,75 @@ def _on_tts(self, msg: String) -> None:
             self.get_logger().info(f"TTS request received: {text[:50]}...")
             self.chat.speak(text)
 
+    def _on_set_active_skills(self, msg: String) -> None:
+        """Update which primitives are registered for the current directive."""
+        payload = json.loads(msg.data)
```
theo-michel, line 294: **Bug — malformed payload crashes the brain node.** `_on_set_active_skills` calls `json.loads(msg.data)` with no try/except, unlike sibling `_on_chat_in` (line 247). `main()` spins via manual `spin_once` and re-raises non-`RCLError` exceptions, so a malformed publish to `/brain/set_active_skills` kills `brain_client_node`. Guard it like the sibling.

ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
```diff
@@ -283,6 +289,75 @@ def _on_tts(self, msg: String) -> None:
...
+
+    def _skill_name_for_id(self, skill_id: str) -> str:
+        for metadata in self.state.registry.metadata:
+            if metadata.get("id") == skill_id:
+                return str(metadata.get("name") or skill_id)
+        return skill_id
+
+    def _on_manual_skill_event(self, msg: String) -> None:
+        """Record manually-triggered simulator skill runs in brain/agent history."""
+        payload = json.loads(msg.data)
```
theo-michel, line 325: **Bug — same crash via `_on_manual_skill_event`.** Unguarded `json.loads(msg.data)` plus unguarded `payload["status"]`/`payload["skill_id"]` indexing → a malformed or missing-key publish raises and kills the node. Also feeds the KeyError at `lifecycle.py:28`.

ros2_ws/src/brain/brain_client/brain_client/skills/lifecycle.py
```diff
@@ -0,0 +1,28 @@
...
+    primitive_id: str | None,
+    reason: str | None = None,
+    output: str | None = None,
+) -> MessageIn:
+    payload = {"primitive_name": primitive_name, "primitive_id": primitive_id}
+    if reason and status == "failed":
+        payload["reason"] = reason
+    if output and status == "completed":
+        payload["output"] = output
+    return MessageIn(type=PRIMITIVE_LIFECYCLE_MESSAGE_TYPES[status], payload=payload)
```
theo-michel, line 28: **Bug — KeyError on unknown status.** `PRIMITIVE_LIFECYCLE_MESSAGE_TYPES[status]` raises for any status outside {running, completed, interrupted, failed}. Reached from `_on_manual_skill_event` with an unvalidated `payload["status"]`, so an unexpected status string crashes the brain node. Use `.get()` + validation.

sim/src/shared_queues.py
```diff
@@ -154,31 +186,139 @@ def update_available_agents(
...
```
theo-michel, line 288: **Bug — dedup drops genuinely-distinct messages.** Key `(sender, text, round(timestamp, 3))` over a session-long set (old code had no dedup). Two real events with the same sender+text rounding to the same millisecond — e.g. a skill run, completed, then quickly re-run — collapse to one and the UI silently misses the second. Add a monotonic sequence id to the key.

sim/src/shared_queues.py
```diff
@@ -57,7 +83,10 @@ def __init__(self, log_everything=False, sim_log_mode: str = "debug"):
 
         # Queues specifically for chat messages
         self.chat_to_bridge = queue.Queue(maxsize=5000)
-        self.chat_from_bridge = queue.Queue(maxsize=5000)
+        self.chat_from_bridge = queue.Queue(maxsize=200)
```
theo-michel, line 86: **Bug — chat history loss.** `maxsize` dropped 5000→200 and `enqueue_chat_from_bridge` `get_nowait()`s the *oldest* item when full. A burst >200 (history replay + live status churn) before the consumer drains silently discards the oldest undelivered messages.

sim/frontend/src/App.tsx
```diff
@@ -1310,44 +486,245 @@ export default function App() {
...
-      const response = await fetch(`${baseUrl}/set_directive`, {
-        method: "POST",
-        headers,
-        body: JSON.stringify({ text: directive }),
-      });
+    const skillOrder = availableSkills.map((skill) => skill.id);
+    const nextSkillIds = activeSkillIds.includes(skillId)
+      ? activeSkillIds.filter((id) => id !== skillId)
+      : skillOrder.filter(
+          (id) => id === skillId || activeSkillIds.includes(id),
```
theo-michel, line 599: **Bug — enabling one skill can disable another.** `nextSkillIds` is built by filtering `skillOrder` (= `availableSkills`), so any currently-active id *not* in `availableSkills` is dropped from the set sent to the backend. If `active_skill_ids` and `skills` diverge (refresh race, directive-declared skill not yet in the payload), enabling skill X silently deactivates the missing active skill.

sim/frontend/src/App.tsx
theo-michel, line 610: **Bug — rapid toggles race.** This `setTimeout(fetchAgents, 150)` handle is never captured/cleared, so a quick second toggle gets clobbered when the first toggle's delayed `fetchAgents` lands and overwrites `activeSkillIds` with the server's pre-second-toggle list. Track + `clearTimeout`, or refetch only after the in-flight mutation settles.

sim/frontend/src/services/rosbridgeService.ts
```diff
@@ -231,11 +247,369 @@ export async function getAvailableAgentsDirect(
...
+          status?: number | string;
+          values?: T;
+        };
+        if (message.op !== "action_result" || message.id !== callId) {
+          return;
+        }
+
+        const status = message.status;
+        const terminalStatus =
+          status === undefined ||
```
theo-michel, line 440: **Bug — a status-less action_result resolves as success.** `status === undefined` is treated as terminal and the resolve branch returns `message.values` → a malformed/early `action_result` with no status is reported as a false 'completed'. Diverges from the Python path (`video_api.py:65`), which treats missing status as non-terminal.

sim/src/routes/video_api.py
```diff
@@ -24,8 +33,115 @@ class SetBrainActiveRequest(BaseModel):
...
+def rosbridge_uri() -> str:
+    return os.getenv("ROSBRIDGE_URI", "ws://localhost:9090")
+
+
+def make_rosbridge_id(prefix: str) -> str:
+    return f"{prefix}_{int(time.time() * 1000)}_{uuid.uuid4().hex[:8]}"
+
+
+def is_terminal_action_status(status) -> bool:
+    return status in (4, 5, 6)
```
theo-michel, line 67: **Bug — 300s hang on a status-less action_result (diverges from the TS client).** `is_terminal_action_status` returns `status in (4,5,6)`, so an `action_result` with no/None status is non-terminal and `execute_rosbridge_action` loops until the 300s `ROSBRIDGE_ACTION_TIMEOUT`. The TS client (`rosbridgeService.ts:409`) treats the same missing status as terminal *success* — same input, opposite wrong behavior. Align the two.

ros2_ws/src/brain/brain_client/brain_client/skills/registration.py
```diff
@@ -79,19 +79,42 @@ def register(self) -> None:
...
+        included = [p for p in primitives if p["id"] in active_skill_ids]
 
         reg_msg = MessageIn(
             type=MessageInType.REGISTER_PRIMITIVES_AND_DIRECTIVE,
             payload={
-                "primitives": included if included else None,
+                "primitives": included,
```
theo-michel, line 88: **Contract change — verify cloud side.** Now sends `"primitives": []` (was `None` when empty) and `"directive": None` (from `EmptyDirective.get_prompt()`, `empty_directive.py:19`). Since `empty_directive` is the default harness state, the cloud agent receives `directive: None` + `primitives: []` on every registration. If the agent treated absent/None as 'keep existing' vs explicit empty, this can wipe tool definitions or break the system prompt. Confirm cloud-agent handling.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -579,7 +981,16 @@ async def outbound_data_loop(ws, shared_queues, service_call_queue):
                 directive_msg = {"data": msg.directive}
                 outbound = rosbridge_publish("/brain/set_directive", directive_msg)
                 await ws.send(json.dumps(outbound))
+                shared_queues.set_current_agent(msg.directive)
                 print(f"[ROSBridge] Published directive: {msg.directive}")
+            elif isinstance(msg, ActiveSkillsCmd):
+                active_skills_msg = {"data": json.dumps({"agent_id": msg.agent_id, "skills": msg.skills})}
+                outbound = rosbridge_publish("/brain/set_active_skills", active_skills_msg)
+                await ws.send(json.dumps(outbound))
+                shared_queues.update_active_skill_ids(msg.skills)
```
theo-michel, line 991: **Bug — optimistic update with no rollback.** `update_active_skill_ids(msg.skills)` records the requested ids immediately, but the brain may reject/filter them (unavailable skill, stale directive). `/available_agents` then reports them active until the next poll — the toggle 'succeeds' in the UI when it didn't.

sim/src/agent/agent_websocket_bridge.py
```diff
@@ -602,6 +1013,7 @@ async def outbound_data_loop(ws, shared_queues, service_call_queue):
 
                 print(f"[ROSBridge] Setting brain active: {msg.active}")
                 await service_call_queue.put(brain_active_srv)
+                shared_queues.set_brain_active(msg.active)
```
theo-michel, line 1017: **Bug — optimistic brain-active update.** `set_brain_active(msg.active)` is recorded right after merely *queuing* the `/brain/set_brain_active` service call. If the service connection is down / the call fails, the UI shows the harness running while the brain never activated. No rollback path.

ros2_ws/src/cloud/innate_uninavid/innate_uninavid/node.py
```diff
@@ -109,6 +121,42 @@ def __init__(self) -> None:
...
+    def _is_configured_service_key(service_key: str) -> bool:
+        return bool((service_key or "").strip())
+
+    def _make_auth_provider(self, service_key: str) -> AuthProvider | None:
+        if not self._is_configured_service_key(service_key):
+            return None
+        return AuthProvider(issuer_url=self._auth_issuer, service_key=service_key)
+
+    def _backend_config_callback(self, msg: String) -> None:
+        payload = json.loads(msg.data)
```
theo-michel, line 134: **Bug — unguarded `json.loads` (same class as the brain_client_node issues).** `_backend_config_callback` parses `/brain/backend_config` with no try/except, while the sibling `WebSocketManager.backend_config_callback` wraps it. Under the node's `MultiThreadedExecutor` it's logged rather than fatal, but the config/service-key update is silently dropped on a malformed message. Guard it like the sibling.

axelpey:
Fixed the IK callback starvation issue in commit `90d27758`.

What changed:
- `ManipulationInterface` now owns a private helper ROS node (`<skills node>_manipulation_interface`) with its own `SingleThreadedExecutor` running in a daemon thread.
- IK/FK/arm-state subscriptions plus arm service clients are created on that helper node instead of the `SkillsActionServer` node.
- `skills_server.py` remains single-threaded; we are not changing the global action-server executor model.
- `solve_ik()` now waits while the helper executor can still deliver `/ik_solution`, so agent/action-triggered code skills no longer starve their own IK response callbacks.
- Added shutdown wiring from `SkillsActionServer.destroy()` and serialized IK requests with a lock because `/ik_solution` has no request id.

Validated locally with:
- `python3 -m py_compile ros2_ws/src/brain/brain_client/brain_client/robot/manipulation.py ros2_ws/src/brain/brain_client/brain_client/nodes/skills_server.py`
- `pre-commit run --all-files --config .config/pre-commit-config.yaml --hook-stage pre-commit`

theo-michel:
## Code review — correctness findings

Ran a high-recall review over the diff. Findings below survived adversarial verification against the PR head (line numbers reference the PR branch).

### 🔴 High — affects the physical robot

**1. `reboot_servos()` desyncs torque state → silent no-op arm motion** — `manipulation.py:434`
`reboot_servos()` sets `self._torque_enabled = True` on success, but the firmware `armRebootServosCallback` leaves arm torque **off** (`arm_services.cpp:343-349`: `configureServosLocked(false)`). After a reboot, the new `if self._torque_enabled is False` guard passes, so `move_to_joint_positions` sends the goal and **reports success while the arm is limp and never moves**. Related: `_torque_enabled` inits to `None` and is never synced from actual hardware state (firmware publishes `is_torque_enabled` but it's never read), so the gate only ever reflects locally-issued calls.

### 🟠 Medium — crashes / data loss

**2. Unvalidated `manual_skill_event` status → `KeyError` crashes brain callback** — `skills/lifecycle.py:28`
`PRIMITIVE_LIFECYCLE_MESSAGE_TYPES[status]` indexes with no membership check. `status` is an unconstrained `str` in `ManualSkillEventRequest` (`video_api.py:48`), published verbatim — any value outside `{running, completed, interrupted, failed}` raises. Compounded by `brain_client_node.py:294,325`: the new `_on_set_active_skills` / `_on_manual_skill_event` callbacks do `json.loads(msg.data)` and index `payload["status"]` with **no try/except**, while sibling callbacks (`_on_chat_in`, `_on_custom_input`) in the same file do guard. The exception propagates out of the rclpy subscription callback.

**3. Bare `json.loads` on `/brain/backend_config`** — `innate_uninavid/node.py:134`
Parses `msg.data` with no guard, while the brain's own handler for the *same topic* (`orchestrator.on_backend_config`) defensively guards `json.loads` + `isinstance(dict)`. Malformed config kills the UniNavid subscription callback.

**4. Chat dedup set is never cleared → reconnect shows empty history** — `shared_queues.py`
`enqueue_chat_from_bridge` dedups on `(sender, text, round(timestamp, 3))` against a 300-key window that's never cleared on reset/reconnect. On a `/ws/chat` (re)connect the history re-sync re-enqueues the last 30 entries through the same function — their keys are still in the window, so **every re-synced message is silently dropped**. A page refresh or second client gets a blank chat panel. (Also: two genuinely distinct messages with the same sender+text in the same millisecond collide and one is lost.)

**5. `chat_from_bridge` queue: maxsize 5000→200 + inverted drop policy** — `shared_queues.py`
Old code dropped the *incoming* message when full (and logged it); new code silently *evicts the oldest un-rendered message*. The drain task only starts on the first frontend connect, so before any client attaches, agent chat buffers and the earliest messages are silently evicted once 200 accumulate.

**6. `solve_ik` holds `_ik_lock` across the full timeout** — `manipulation.py:233`
The lock now wraps the publish **and** the entire `while ... < timeout` polling loop. A second concurrent IK caller blocks for the full remaining timeout, so concurrent manipulation skills serialize and multiply latency.

### 🟡 Lower

**7. `parse_available_skills_message` unguarded in `inbound_data_loop`** — `agent_websocket_bridge.py:534`
A non-numeric `episode_count` raises `ValueError`, tearing down the rosbridge data connection. It auto-reconnects (self-recovering) but drops all in-flight cmd_vel/chat/arm traffic until then.

**8. Skill-registration contract change** — `skills/registration.py`
Now sends `primitives: []` (was `None`) and registers only the *active* subset. Deselecting all skills registers zero primitives; the `[]`-vs-`None` distinction may matter to the cloud agent (separate repo — worth confirming on that side).

**9. Placeholder service-key validation weakened** — `transport/ws_config.py`
The `PLACEHOLDER_SERVICE_KEYS` blocklist was removed; any non-empty token now passes. Mostly fine since the new `.env.template` default is empty (still caught), but a literal placeholder like `changeme` now attempts a real auth instead of returning a clear `invalid_config` diagnostic.

### Cleanup (not bugs, worth a look)
- Skill-parsing logic is duplicated inline vs the new `parse_available_skills_message` helper.
- The rosbridge action/service client (status `4/5/6` → success/cancelled/failure) is reimplemented in **both** Python (`video_api.py`) and TS (`rosbridgeService.ts`) — protocol drift risk across the two languages.

---
Suggest treating **#1** and the **#2–#3** unguarded-callback crashes as merge blockers, since they hit the real robot, not just the sim.

PR #417, opened 2026-06-02
Add system-wide brain-loop testing (+ ruff adoption, dead-code removal)

theo-michel:
## Why

The goal is **basic system-wide testing** for the robot OS — a deterministic
integration test that drives the real brain stack end to end. A few enabling
changes were needed to make that test actually pass in CI, so they ride along.

## 1. System-wide test (the point)
- **`FakeCloud`** — scriptable, no-ROS stand-in for the cloud agent's VLM brain,
  speaking the exact robot↔cloud WebSocket protocol.
- **`test_fake_cloud_loop.launch.py`** — launches the real `brain_client_node` +
  `ws_client_node` + `skills_action_server` and drives
  connect → register → dispatch → execute → complete (+ `chat_in`), asserting on
  the wire transcript. Same nodes run on hardware, so green ⇒ high confidence.
- **`test_fake_cloud_selftest.py`** — guards the double (pure Python, no ROS).
- Wired into CI; **`TESTING.md`** documents the gates (format / build / test).

## 2. Make CI green so the test can run
- **ruff**: the format gate was red across the whole repo. Added a pinned
  pre-commit hook, scoped the CI check to the whole repo, reformatted + applied
  safe autofixes everywhere, and grandfathered remaining legacy violations with
  `noqa` (ratchet — rules stay enforced for new code). `.git-blame-ignore-revs`
  keeps `git blame` clean across the reformat.
- **build**: pinned `topic_tools` to `humble` (its default branch doesn't compile
  against Humble's rclcpp) and fixed the `Dockerfile.test` entrypoint heredoc.
- **build coverage**: un-skipped `maurice_cam`/`nav`/`control` — they compile fine
  on non-Jetson (cam's VPI stereo is `if(VPI_FOUND)`-guarded; nav's `python3-cupy`
  is runtime-only). The CI build gate now compiles **all packages** (verified: full
  20-package image build is green), instead of 15.

## 3. Remove dead code
Deleted the unused MuJoCo (`maurice_sim` sim node) and Stage (`stage_ros2`) sim
paths + the orphaned `arm_utils` node — none referenced by any active launch flow
(active sim is innate-sim/Genesis over rosbridge). Kept `maurice_sim` URDF/meshes,
still used by the arm/IK stack.

## Verification
- `test_fake_cloud_loop`: 5/5 in-container; `test_fake_cloud_selftest`: 7/7.
- `ruff check .` / `ruff format --check .`: green.
- Full `colcon build` (20 packages, all formerly-skipped ones included): green.

## Notes for reviewers
- Large diff from the repo-wide reformat — **will conflict with in-flight branches; merge promptly.**
- Coverage is still a thin behavioral slice: `test_fake_cloud_loop` is one happy-path
  thread through the brain. Most packages have no behavioral tests yet — the build
  gate (now all packages) is the broad net for structural refactors; per-subsystem
  behavioral tests are follow-up work.
- `maurice_cam`'s VPI stereo-depth component (Jetson-only) is still not compiled in CI.


vignesh-anand:
## Code Review Notes

Great PR overall — the FakeCloud design is clean, the two-tier test approach (selftest guards the double, integration test guards the stack) is the right structure, and the ruff adoption + dead code removal are solid housekeeping.

Four things worth addressing before merge:

---

### 1. Tests share state across methods — ordering dependency is invisible

All 5 test methods in `TestFakeCloudLoop` write to the same global `FAKE` transcript, which accumulates across tests. They run alphabetically, so `test_brain_sends_perception` (10s image feed, likely triggers primitives) runs before `test_loop_continues_after_completion` (asserts ≥2 `primitive_completed`). The later test may be passing entirely on state produced by the earlier one — not on its own work.

Add a new test that resets state, or change the runner order, and tests fail in confusing ways.

**Fix:** Record a baseline at the start of each test and slice from it:
```python
def test_loop_continues_after_completion(self):
    before = len(FAKE.transcript)
    self._feed_perception(duration=30.0)
    completed = [m for m in FAKE.transcript[before:] if m.get("type") == "primitive_completed"]
    ...
```

---

### 2. `allowable_exit_codes=[0, -2, 1]` masks real crashes

Both `test_fake_cloud_loop` and `test_pose_image` allow exit code `1` in `TestShutdown`. In Python, `1` is the generic unhandled exception exit code — any node that crashes with a traceback exits `1`, and this would be silently accepted.

`-2` (SIGINT on teardown) is expected. `1` is not.

**Fix:**
```python
allowable_exit_codes=[0, -2]
```

---

### 3. `_feed_perception` uses fixed-duration sleeps

`_feed_perception(duration=10.0/20.0/30.0)` spins for a hardcoded wall-clock window regardless of when work actually completes. On fast machines it wastes time; on slow CI machines the window may close before a primitive finishes, causing a false failure.

**Fix:** Use `FAKE.wait_for()` instead — it returns as soon as the expected message arrives and only burns the full timeout when something is actually wrong:
```python
FAKE.wait_for("primitive_completed", timeout=30.0)
```

---

### 4. `selfhosted` runner is a silent single point of failure

`integration-test.yml` runs on `selfhosted` with no fallback. If the runner goes offline the job queues indefinitely — it does not fail, it just never runs. PRs can merge with the check permanently pending if it is not configured as a required status check in branch protection.

**Fix:** Confirm `integration-test` is marked as a required status check on `main` in Settings → Branches → Branch protection rules. That way a stuck runner blocks the merge rather than silently passing.

---

_Reviewed on `theo/ruff-precommit-cleanup` @ 6c229ce2_

vignesh-anand:
**TODO:** Replace `selfhosted` runner with on-demand GCP integration — spin up a GCP VM (e.g. Compute Engine or Cloud Run job) from the workflow, run the Docker-based tests there, then tear it down. Eliminates the single point of failure of a persistent self-hosted runner.

PR #424, opened 2026-06-03
refactor(brain_client): concept-folder layout + decompose god-class nodes

theo-michel:
## Why

`brain_client` had grown into a flat ~25-module package with two god-class nodes (`brain_client_node` 2587 lines, `skills_action_server` 1575). This is an architecture pass — reorganize by concept, small single-responsibility modules, and a testable pure core — with behavior held at parity. Also tried reusing common repeating code.

## What

- **Concept-folder layout**: `core/ perception/ navigation/ skills/ agents/ inputs/ transport/ robot/ common/ nodes/`. The package top level is now just the runnable entry points.
- **All four nodes decomposed** into thin composition roots + focused collaborators (brain loop, lifecycle, vision-output, skill runner/registry/catalog, robot-state, ws transport).
- **Pure, ROS-free core** (`pose`, `payload`, `image_codec`, `registry`, `config`, `messages`) unit-tested without a ROS runtime; an import-rule test enforces the no-`rclpy` boundary.
- **WebSocket transport fused in-process** — cloud/image payloads no longer cross a ROS topic or get re-serialized; incoming messages cross to the executor thread via a guard condition (no handler runs on the socket thread).
- **Dead code removed, duplication collapsed** (skill id↔name resolution, stop command, chat entry, …); READMEs added as signposts.

## Verified

`colcon build` · pure unit tests · `test_fake_cloud_loop` 5/5 (full directive→primitive→completion loop) · `test_pose_image` 3/3 — at parity with the pre-refactor baseline (compared via a worktree at the base commit). Lint/format clean.

also verified running it on the robot, update system is working


vignesh-anand:
## Code Review

Solid refactor — the decomposition is well-executed and the net result is dramatically more readable. The \`DynamicLoader\` base class, \`BrainConfig\` frozen dataclass, in-process WebSocket transport with guard-condition hand-off, and \`import_rules_test.py\` architectural guard are all standout decisions. The composition root is exactly what it should be: wiring with no logic.

Issues below, ordered by priority.

---

### Bugs

**`_svc_get_directives` crashes when no directive is loaded**
\`nodes/brain_client_node.py\`, \`_svc_get_directives\`:
\`\`\`python
response.current_directive = self.state.current_directive.id
\`\`\`
\`current_directive\` is \`None\` if no agents were found at startup (empty \`workspace/innate_agents/\`, all agents fail to load, etc.). The startup code explicitly guards \`if self.state.current_directive:\` but this service handler doesn't — it will raise \`AttributeError\` on the first call to \`/brain/get_available_directives\` in that state.

Fix: \`self.state.current_directive.id if self.state.current_directive else ""\`

---

**Breaking change for robots with custom skills/agents outside this repo**

This PR deletes the old module paths that custom skills and agents must import to function:

\`\`\`
brain_client/skill_types.py    →  brain_client/skills/types.py
brain_client/agent_types.py   →  brain_client/agents/types.py
brain_client/input_types.py   →  brain_client/inputs/types.py
brain_client/logging_config.py →  brain_client/common/logging.py
\`\`\`

Custom skills and agents aren't optional importers — they *must* inherit from \`Skill\`, \`Agent\`, or \`InputDevice\` to be discovered and loaded at all. A skill file on a customer's robot looks like:

\`\`\`python
from brain_client.skill_types import Skill, SkillResult  # this path is now gone

class PickApple(Skill):
    def execute(self, target: str) -> tuple[str, SkillResult]:
        ...
\`\`\`

The \`workspace/\` files in this repo are all updated, but any custom skill or agent living on a robot outside this repo will silently stop loading the moment the package is updated — \`DynamicLoader\` will catch the \`ModuleNotFoundError\`, log a warning, and skip the file. No build error, no crash, the skill just disappears from the available list.

**Fix:** Leave forwarding shims at the old paths for at least one release cycle:
\`\`\`python
# brain_client/skill_types.py
from brain_client.skills.types import *  # noqa: F401, F403
\`\`\`

This keeps the old imports working while customers migrate their files, and can be deleted in a later release.

---

### Inherited hazards (pre-existed, but worth tracking)

**\`time.sleep(0.5)\` on the executor thread** — \`core/lifecycle.py\`, \`reactivate_brain()\`
Stalls all timer and subscriber callbacks for 500ms while waiting for the server to process \`READY_FOR_CONNECTION\`. A one-shot timer would be the correct fix but is out of scope here — worth a follow-on issue.

**\`rclpy.spin_once\` re-entrancy in \`perform_full\`** — \`skills/hot_reload.py\`
Calling \`spin_once\` on a node that's already on the executor thread (inside a service handler) is re-entrant. Also pre-existing; flag as a follow-on.

---

### Design

**\`BrainState.token\` set but unused for reconnection** — \`core/orchestrator.py\`, \`on_backend_config\`
\`state.token\` is updated here, but \`WebSocketManager\` handles actual token/URI updates independently via its own \`backend_config_callback\`. The two code paths don't share state. Either remove \`state.token\` or add a comment explaining what it's for — right now it looks like a bug.

**Private method passed as callback across class boundary** — \`core/vision_output.py\`
\`\`\`python
self._runner.cancel_active_goal(on_done=self._runner._on_cancel_response)
\`\`\`
\`_on_cancel_response\` (which just logs) should be the default \`on_done\` inside \`cancel_active_goal\`, not passed in from the outside.

**\`_log_backend_unavailable_once\` dropped from \`WebSocketManager\`** — \`transport/ws_manager.py\`
The old \`WSClientNode.set_ws_status\` rate-limited the "Brain backend unavailable" error log to once per 30s. The new manager doesn't have this. Repeated disconnect/reconnect cycles will produce chatty logs.

---

### API risk

**Removed \`CameraProvider\` properties** — \`perception/camera_provider.py\`
\`last_main_camera_raw\`, \`last_wrist_camera_raw\`, \`last_main_camera_time\`, \`last_wrist_camera_time\` removed as dead code. Fine if nothing outside this repo reads them — worth a grep across any customer skill repos before merging.

---

### Nits

- \`import json\` inside \`_on_head\` callback (\`perception/camera.py\`) — harmless but inconsistent with the rest of the file, move to module level.
- \`reactivate_brain\` logs "Brain reactivated and reset initiated" but \`perform_brain_reset\` is not called — only \`catalog.register()\`. Misleading message.

---

**Bottom line:** The \`_svc_get_directives\` null-directive crash and the broken import paths for custom skills are both bugs that should be fixed before merging. Everything else is either inherited from the pre-refactor code or a low-priority cleanup. The architecture is sound.

vignesh-anand:
## Suggested test coverage

Based on reading the full diff, here's a breakdown of cases worth covering across each subsystem. Current integration tests (`test_fake_cloud_loop`, `test_pose_image`) cover the happy path end-to-end — these are mostly the edge cases and failure paths that aren't yet tested.

---

### Pure unit tests (no ROS, fast)

**Config**
- All params load with correct defaults
- `horizontal_fov` derived correctly from vertical FOV + aspect ratio
- `camera_info` / `proxy_config` dict shapes

**SkillRegistry** *(partially covered)*
- Resolve by id, resolve by name, resolve unknown → `None`
- Duplicate name: last wins, callback fires
- `__bool__` falsy on empty, truthy when populated
- `name_for` falls back to id when unknown

**Navigation payload** *(partially covered)*
- Map-free vs real map payload structure
- `robot_coords` dict shape

**Image codec** *(partially covered)*
- Encode `None` → `None`
- Video encode with mismatched frame dimensions (bad frames skipped)
- Temp file cleaned up on failure

**Pose math** *(partially covered)*
- `compute_pose_delta` / `adjust_local_nav_command`
- Quaternion-to-yaw: identity, ±90°, 180°

**WebSocket config** *(fully covered)*

---

### WebSocket transport

- Exponential backoff caps at 30s
- Old thread joined before new one starts on backend config update
- `version_mismatch` error → TTS spoken + status update
- Invalid URI → `invalid_config` status, no connection attempt
- Placeholder token for hosted URI → `invalid_config` status
- Live backend config swap: URI changes → old thread stopped, new connection started
- Same URI + same token on config update → no-op
- Invalid JSON on `/brain/backend_config` → logged, ignored
- **WSBridge**: incoming on socket thread → enqueued, not dispatched until executor thread drains
- **WSBridge**: handler raises exception → error logged, bridge continues
- **WSBridge**: unknown message type → warning, no crash
- **WSBridge**: missing `type` field → error logged
- **WSBridge**: invalid JSON → error logged

---

### Brain lifecycle

- Activate when already active → no-op
- Deactivate when already inactive → no-op
- Deactivate: cancels running skill, stops gaze, stops subscriptions, publishes zero velocity, deactivates inputs
- Reactivate with no directive → warns, remains inactive
- Directive switch to unknown name → error in chat, directive unchanged
- Brain reset → unregisters primitives, clears chat, sends RESET, re-registers
- Simulator mode auto-activation after registration

---

### Skill registration (SkillCatalog)

- Registration rejected by cloud → `primitives_registered` stays `False`
- Re-registration on WebSocket reconnect
- Pending re-registration while skill running → deferred until `on_task_finished`
- Duplicate display names filtered before publishing (first kept, error logged)

---

### Perception loop (Orchestrator)

**Image send**
- Brain not active / primitives not registered / `ready_for_image` not set → skip
- No image available → skip
- No pose source → skip
- Image encode failure → clears `ready_for_image`, returns
- Video feed mode with no recent frames → falls back to single image
- `pose_at_image_send` stamped before send

**Pose image**
- Navigation mode → includes AMCL covariance
- Mapfree mode → covariance set to `1e4`
- Mapping mode → skipped (unless simulator mode)
- TF `map→base_link` not ready → skipped gracefully
- `pose_image_started` gate: timer only created once

---

### VisionOutputHandler

- Brain not active → message dropped
- Primitives not registered → message dropped
- `stop_current_task=True` with active goal + pending next task → cancel goal, store pending
- `stop_current_task=True` with no active goal → log only
- `next_task` with `name` field instead of `type` (HOTFIX alias) → works correctly
- Unknown skill type in `next_task` → warning, no crash
- Local nav compensation applied for `navigate_to_position(local_frame=True)`
- Local nav compensation skipped for all other skills
- Local nav compensation skipped when `pose_at_image_send` is `None`

---

### Primitive runner

- Goal rejected → status "failed", `primitive_running` cleared, `on_task_finished` called
- `SUCCESS` → PRIMITIVE_COMPLETED + task_status "completed"
- `CANCELLED` → PRIMITIVE_INTERRUPTED + task_status "interrupted"
- `FAILURE` → PRIMITIVE_FAILED with reason + task_status "failed"
- `abort_running`: fire-and-forget cancel, clears state
- `interrupt_for_deactivation`: cancel + PRIMITIVE_INTERRUPTED + status
- Pending task runs after cancellation completes (CANCELLED or SUCCESS result)
- Pending task dropped if previous task failed
- `_stop_robot()` and `on_task_finished` called on every result path

---

### Hot reload

- Full reload: deactivates brain, waits for primitives, reinitializes directives, re-registers
- Timeout waiting for primitives after full reload → warning, continues
- Selective: skills only → PEAS called, agents skipped
- Selective: agents only → reloaded in-process, PEAS skipped
- Nothing reloaded → no re-register
- Multiple file saves within debounce window → coalesced into one reload
- Multiple saves for different files → names merged

---

### Skill execution (SkillsActionServer)

**Code skills**
- Skill raises exception → `goal_handle.abort()`, error logged
- Skill disappears between dispatch and lookup (concurrent reload) → abort with message
- Invalid inputs JSON → abort immediately

**Physical skills**
- Behavior server unavailable after 5s → abort
- Goal rejected → abort
- Timeout waiting for goal acceptance → cancel-when-ready, abort
- Cancel requested before goal accepted → sent when goal ready
- "cancel" in failure message → treated as CANCELLED
- Behavior server dies mid-execution → "server_unavailable" → abort
- Physical skill disappears during reload → abort with message

**Robot state injection**
- Any required state unavailable → warning, not injected (not a crash)
- Continuous updates stopped in `finally` even on exception
- Subscriptions started before execution, stopped in `finally`

---

### Camera capture

- `start()` idempotent (no double-subscribe)
- `stop()` clears buffers
- Depth encoding `16UC1` → `uint16`, `32FC1` → `float32`, unknown → `uint8` + warning
- Compressed image decode failure → warning, `last_image` unchanged
- `recent_frames` returns only frames within time window
- `clear_after_send` resets depth + arm, keeps RGB

---

### Service handlers

- `get_available_directives` when `current_directive=None` → no crash (currently a bug — see other comment)
- `set_brain_active` True when already active → "already active", `success=True`
- `set_brain_active` False when already inactive → "already inactive", `success=True`

---

### Startup

- Skills server not running → 25s timeout, warning, continues with empty registry
- No agents found → `current_directive=None`, brain stays inactive, no crash
- `activate_directive_inputs` only called when `current_directive` is not `None`

---

### DynamicLoader (shared discovery machinery)

- File with missing dependency → warning, file skipped, others still load
- File with syntax/runtime error → error logged, file skipped
- Class that doesn't subclass the base → skipped
- Base class itself → skipped
- Class re-exported from an imported module (not defined in this file) → skipped
- Multiple valid classes in one file → all loaded
- Name conflict across directories → last wins, warning logged
- Directory doesn't exist → warning, empty result

theo-michel:
TODO: Test everything thuroughly also backwards compatibility, maybe add a test in integrationt test?

PR #427, opened 2026-06-06
Teleop joystick + robot-mic audio; 0.5s motor deadman; brain_client hot-reload & configurable agent/skill dirs

theo-michel:
## Summary

This branch bundles several related changes: teleop feel + audio, brain_client skill/agent tooling, and a few smaller cleanups picked up along the way.

### 1. Manual joystick teleop (`app.cpp`)
Net change is a **refactor plus one opt-in feature — default driving feel is unchanged**.
- Tuning values extracted into named constants in the `joy_tuning` namespace; defaults are identical to the pre-PR feel (15% deadband, quadratic curve, 0.5 m/s / 1.0 rad/s caps).
- New **car-like reverse steering** behind a `reverse_steering` ROS parameter, **off by default**; the app's Dev tab can toggle it at runtime via `set_parameters`. When enabled, `angular.z` inverts while reversing, ramped in over the first 0.15 m/s of reverse speed so steering doesn't snap at zero throttle.
- Removed a per-message INFO log.

> History note: this branch experimented with a softer curve (deadzone 0.04, exponent 1.5), 50 Hz acceleration smoothing, and a speed-dependent turn bonus. All were reverted after testing; the final commits restore the pre-PR defaults. Squashable on request.

### 2. Hot-reload physical-skill `metadata.json` / asset edits (`brain_client`)
Physical skills live in a per-skill subdirectory (`<skills_root>/<skill>/metadata.json` + assets), but the hot-reload watcher watched skill roots **non-recursively** and reacted only to `.py` files — so editing a physical skill's `metadata.json` or assets produced no events and never reloaded. The catalog's reload path was already correct; the gap was purely in the watcher.
- Skills watcher is now **recursive** (agents stay flat / non-recursive — they're single `.py` files).
- A path resolver maps any changed file to the skill it belongs to: `root/foo.py` → `foo` (code skill, `.py` only), `root/foo/<anything>` → `foo` (physical skill, any file type). Roots and event paths are realpath-normalized so symlinked dirs (e.g. macOS `/var` → `/private/var`) still match.
- Filters editor/build noise so reloads can't loop or fire spuriously: `__pycache__` (the `.pyc` written *during* a reload), dotfiles/dot-dirs (`.git`, `.DS_Store`), and `.tmp`/`.swp`/`~` scratch files.
- Handles atomic-save renames via `on_moved` (many editors — and the existing `create_physical_skill` write — rename a temp file onto the target).
- The unused legacy `HotReloadHandler` class at the top of `hot_reload_watcher.py` was **removed** as part of the rewrite.

**Reviewer note:** recursive watching means writing episode data into a `custom_skills/<skill>/` dir during training now triggers a debounced, cheap selective reload (re-reads metadata + recounts episodes — keeps `episode_count` live). Harmless-to-helpful; narrow in `_skill_name_for` if it ever proves noisy.

### 3. Stream robot microphone to teleoperator over WebRTC (`maurice_cam`)
Adds a one-way Opus audio track (`webrtc.sink_2`) to the WebRTC streamer so the manual-control teleoperator can hear the robot's mic over the **same** peer connection that already carries the camera feeds — no new signaling or dependencies; the audio m-line is bundled into the existing offer.
- **Per-connection opt-in:** the client must send `{"audio": true}` in the `/webrtc/start` payload; default is off, so the agent page (and any app build that doesn't send the flag) gets a video-only offer and leaves the OS audio session free for its speech APIs. ⚠️ **Hearing audio requires a controller-app change that sends this flag** — until that ships, behavior is unchanged.
- `enable_audio` node parameter is the master capability switch: **on** in the hardware launch files (Arducam addressed by stable card name `sysdefault:CARD=Light`, not a positional index), **off in the sim launch** (no capture device).
- Mic branch: `alsasrc → opusenc` (mono 48 kHz, voice profile, declared as `opus/48000/2` per RFC 7587 so libwebrtc accepts the m-line) behind a leaky queue, so audio is dropped rather than stalling video under network pressure.
- Pipeline build/start refactored into `build_pipeline_description()` / `start_pipeline_locked()`; element/device parameter values are validated before being spliced into the pipeline string. If the audio branch fails to come up (missing/busy mic, missing opus plugin), `on_start()` retries **video-only** so teleop driving is never blocked by the microphone.
- New 200 ms health poll on the executor thread drains the GStreamer bus (runtime errors now get logged) and tears the pipeline down once the peer connection is gone — immediately on CLOSED, after a ~15 s grace window on FAILED/DISCONNECTED so webrtcbin can recover transient blips via ICE restart. Frees the encoders and the mic on dead sessions.

### 4. Config-driven extra agent/skill directories (`brain_client`)
Let a user point at agent/skill directories **anywhere on the machine** (not just `workspace/`) via a new `[paths]` section in `config/os.toml`, and have those dirs hot-reload like `workspace/`:
```toml
[paths]
agent_dirs = "/home/me/my-agents:/opt/team/agents"   # ":"-separated absolute paths
skill_dirs = "/home/me/my-skills"
```
- Reuses the existing config → env → `script_paths` pipeline: `env_loader._load_os_config` exports `INNATE_EXTRA_AGENT_DIRS` / `INNATE_EXTRA_SKILL_DIRS` (accepts a delimited string — which parses identically with or without `tomllib`, so it works on the robot's Python 3.10 — or a TOML array). Every brain_client launch file already calls it. Nonexistent configured dirs log a warning instead of failing silently.
- `script_paths.get_{agent,skill}_directories()` fold the extra dirs in; provenance is `"user"` (ids → `local/<name>`).
- The loaders **and both hot-reload watchers** already consume those getters, so the extra dirs are watched for free (skills recursively, agents flat).
- **Caveats:** the dir *list* is read at launch (adding/removing a dir needs a restart; hot reload covers files *inside* configured dirs). In a Docker/sim setup, host paths must also be mounted into the container.

### 5. Motor deadman cut from 5 s to 0.5 s (`maurice_bringup` i2c/uart)
Every `/cmd_vel` producer streams continuously (nav smoother 40 Hz, app joystick heartbeat ~7 Hz, skills 25 Hz), so going 0.5 s without a command means the operator/controller is gone and the base should stop rather than coast 5 s on the last latched velocity. ⚠️ Any future producer that publishes one-shot Twists and expects them to latch will now stop after 0.5 s — by design.

### 6. `maurice_sim` slimmed to an asset-only package
The package only ever served URDF/SRDF/meshes to the arm/IK stack; the Python module was an empty `__init__.py`. Dropped the module, `rclpy`/message deps, and lint tests; the share-dir install (`urdf/`, `meshes/`) that all consumers (`maurice_arm`, `maurice_bringup`, `manipulation`) actually use is unchanged.

### 7. Training-manager static bundle rebuild
`training_manager/static/` regenerated from the current frontend source (hash-named JS bundle + matching `index.html` reference). No intended functional change.

## Validation

  - **No unit tests are included, intentionally.** Dev-time tests for the watcher and extra-dirs code were written during development and deliberately removed before merge; validation of those paths was manual (editing a physical skill's `metadata.json`/assets and watching the selective reload fire; configuring `[paths]` dirs and confirming scan + hot reload, including the no-`tomllib` fallback).
  - `ruff check` + `ruff format --check` clean on changed Python files.
  - **Joystick (#1), WebRTC mic (#3), deadman (#5):** C++/hardware-facing; not compiled on this macOS box (no native ROS2 / cross-compile) — relies on CI build. Recommended hardware smoke test before release: drive and release the stick (0.5 s stop), toggle `reverse_steering` and reverse, and check mic audio with an app build that sends `"audio": true`.
- [ ] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate sim up`. — The WebRTC change touches `webrtc_streamer.sim.launch.py` only to pass `enable_audio:=false` (audio off in sim); sim startup was not re-verified in this session.
- [ ] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`. — N/A.


ros2_ws/src/maurice_bot/maurice_cam/maurice_cam/webrtc_streamer.cpp
```diff
@@ -407,12 +461,44 @@ void WebRTCStreamer::on_start(const std_msgs::msg::String::SharedPtr msg) {
...
+    if (with_audio) {
+        // One-way Opus microphone track (sink_2). Mono 48 kHz at a voice
+        // bitrate keeps it light; the leaky queue drops audio rather than
+        // stalling if the network backs up. webrtcbin negotiates this as a
+        // sendonly audio m-line that the phone plays automatically.
+        std::string src = audio_source_element_;
+        if (!audio_capture_device_.empty()) {
+            // Quote the device so ALSA names like hw:1,0 survive gst parsing.
+            src += " device=\"" + audio_capture_device_ + "\"";
+        }
```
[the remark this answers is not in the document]
theo-michel: Done in 733e1049 — validate audio_source_element is a plain element name (alphanum/-/_) before splicing into the pipeline string; on an invalid value we log an error and fall back to video-only. Device stays separately quoted.

ros2_ws/src/brain/brain_client/brain_client/skills/hot_reload_watcher.py
```diff
@@ -172,14 +178,14 @@ def start(self):
...
         )
 
         # Watch all directories
         watched_count = 0
         for directory in self.skills_directories + self.agents_directories:
             if os.path.exists(directory):
-                self._observer.schedule(handler, directory, recursive=False)
+                self._observer.schedule(handler, directory, recursive=self.recursive)
                 self.logger.info(f"👁️ Watching for changes: {directory}")
                 watched_count += 1
```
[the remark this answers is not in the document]
theo-michel: Done in 733e1049 — recursive is now applied only to skills directories; agents directories are always scheduled with recursive=False, matching the documented intent.

ros2_ws/src/maurice_bot/maurice_control/maurice_control/app.cpp
vignesh-anand, line 1: Deadzone discrepancy — PR description says "deadzone 0.15 → 0.08" but the constant in code is 0.04. One of those numbers is wrong. That's a real discrepancy worth resolving — 0.04 vs 0.08 is a meaningful difference for drift sensitivity.

vignesh-anand, line 1: Not necessary but considering making the reverse direction mode a selectable options. also the slow mode (will wait for nicks feedback cause this is his request)

ros2_ws/src/maurice_bot/maurice_cam/maurice_cam/webrtc_streamer.cpp
vignesh-anand, line 1: The audio fallback won't fire for a busy/wrong mic device.

gst_element_set_state for a live source (alsasrc) almost always returns GST_STATE_CHANGE_ASYNC — GStreamer opens the mic on a background thread and returns immediately. The failure check only catches GST_STATE_CHANGE_FAILURE, which alsasrc won't produce here — so start_pipeline_locked returns true, the fallback never triggers, and a bad device just goes silently dead.

Fix: after ASYNC, call gst_element_get_state(pipeline_, nullptr, nullptr, 3 * GST_SECOND) to wait for the state change to resolve, then check the result before returning.

vignesh-anand, line 1: That's a real problem. on_connection_state_changed only logs the state — when the connection goes to DISCONNECTED or CLOSED, it does nothing. The pipeline keeps running, and the mic stays locked.

So the lifecycle is actually:

Teleoperator connects → on_start fires → pipeline starts, mic is opened
Teleoperator disconnects → connection state changes to DISCONNECTED/CLOSED → logged only, pipeline left running, mic still held
Next teleoperator connects → on_start fires again → cleanup_pipeline tears it down and rebuilds

vignesh-anand, line 1: No null check after gst_bin_get_by_name — crash risk

appsrc_main_ = gst_bin_get_by_name(GST_BIN(pipeline_), "src_main");
appsrc_arm_  = gst_bin_get_by_name(GST_BIN(pipeline_), "src_arm");
webrtc_      = gst_bin_get_by_name(GST_BIN(pipeline_), "webrtc");

// immediately used with no null check:
g_object_set(appsrc_main_, ...);   // segfault if null
g_object_set(appsrc_arm_,  ...);   // segfault if null
g_signal_connect(webrtc_,  ...);   // segfault if null
gst_bin_get_by_name returns null if the element isn't found. In normal operation the pipeline string always contains src_main, src_arm, and webrtc, so this is fine. But if the pipeline string ever gets malformed in a way that gst_parse_launch doesn't catch, or an element gets renamed, you get a hard crash with no log, no fallback, nothing. Three lines of null checks would prevent that.

vignesh-anand, line 1: Async fallback — the mic-busy case never triggers the retry, need gst_element_get_state with a timeout after ASYNC
Pipeline not torn down on disconnect — on_connection_state_changed only logs, mic is held between sessions, add cleanup_pipeline() on DISCONNECTED/CLOSED/FAILED
No GStreamer bus handler — runtime audio failures are completely invisible, need gst_bus_add_watch
build_pipeline_description silent fallback — validation failure returns video-only but caller logs "audio=on", misleading
No null check after gst_bin_get_by_name — potential crash if elements aren't found
hw:0,0 is positional — use sysdefault:CARD=<name> instead
webrtc_streamer.launch.py has empty device — opens the wrong card silently, and the async failure means it won't even fall back to video-only

ros2_ws/src/brain/brain_client/brain_client/skills/hot_reload_watcher.py
vignesh-anand, line 1: 1. HotReloadHandler at the top of the file is dead code

The old class is still sitting there — 80+ lines, its own debounce logic, completely different callback signature. It's never instantiated anywhere. The PR description acknowledges it ("left untouched — candidate for a separate cleanup") but it's actively confusing: a reader sees two classes with similar names and has to figure out which one is actually used. _InternalHandler at the bottom is the live one.

2. on_deleted is not handled

_InternalHandler only overrides on_modified, on_created, and on_moved. If someone deletes metadata.json from a physical skill, or deletes a skill's .py file entirely, watchdog fires a deleted event — which goes to the default no-op handler. No reload, no update to the catalog. The skill entry stays stale until something else triggers a reload.

3. getattr(event, "dest_path", None) is unnecessarily defensive

FileMovedEvent always has dest_path — it's a required attribute of the class. The getattr with a default implies it might be missing, which it won't be. Minor, but slightly misleading to a reader.

4. Dotfile check and _is_transient overlap slightly

In _skill_name_for:


if any(part.startswith(".") for part in parts):
    return None
And in _is_transient:


return filename.startswith(".")
Both catch dotfiles in the filename (.DS_Store, .gitignore). The first check catches them as path components, the second as the filename specifically. Not a bug — just redundant for the filename case.

config/os.toml.template
vignesh-anand, line 1: 1. TOML array syntax silently breaks on Python 3.10

The documented format is colon-separated: agent_dirs = "/path/one:/path/two". But a user might reasonably try TOML array syntax: agent_dirs = ["/path/one", "/path/two"]. On Python 3.11+ with tomllib, _join_dirs handles this correctly — it gets a list. But on Python 3.10 (the robot), the fallback parser doesn't understand TOML arrays — it returns the raw string ["/path/one", "/path/two"]. _join_dirs treats that as a string, splits on :, finds no colons, and produces a single path named ["/path/one", "/path/two"] — clearly wrong, never found on disk, silently fails.

No error, no warning, skills from those dirs just don't load.

2. No warning when a configured directory doesn't exist

Extra dirs are in the required list in _scan_dirs. That means they're always included in the scan list whether or not they exist on disk. The loaders handle missing dirs gracefully — they just find nothing. But if you typo a path in os.toml, you get total silence. No log message saying "configured dir /home/me/my-skilss does not exist". The skills just don't load and you have to figure out why.

3. setdefault means a shell environment variable silently wins over the config file


os.environ.setdefault("INNATE_EXTRA_AGENT_DIRS", joined_agent_dirs)
setdefault only sets the value if it's not already there. So if INNATE_EXTRA_AGENT_DIRS is already set in the shell environment when the robot starts, the config file value is silently ignored. This is consistent with how the rest of env_loader.py works, so it's not a new problem — but it's worth knowing that the env var always wins over os.toml.

vignesh-anand (approved):

ros2_ws/src/maurice_bot/maurice_control/maurice_control/app.cpp
theo-michel: Yes!

theo-michel: Fixed — updated the PR description to "deadzone 0.15 → 0.04" to match the constant in code. 0.04 is the intended value (smaller deadzone for less drift); the description was stale.

ros2_ws/src/maurice_bot/maurice_cam/maurice_cam/webrtc_streamer.cpp
theo-michel: Fixed in db92251d — after GST_STATE_CHANGE_ASYNC we now wait via gst_element_get_state(…, 3 * GST_SECOND) so a busy/missing mic surfaces here and trips the video-only fallback instead of going silently dead.

theo-michel: Fixed in f9e890dc — added a null check on src_main/src_arm/webrtc right after gst_bin_get_by_name; on a missing element we log and tear down instead of segfaulting in the g_object_set/g_signal_connect calls.

ros2_ws/src/brain/brain_client/brain_client/skills/hot_reload_watcher.py
theo-michel: Fixed in f9e890dc — (1) removed the dead HotReloadHandler class and now-unused imports; (2) added on_deleted so deleting a skill/agent file reloads and drops the stale catalog entry; (3) use event.dest_path directly; (4) dropped the redundant dotfile check in _is_transient (already filtered by the per-part hidden-path check).

config/os.toml.template
theo-michel: Fixed in f9e890dc — (1) the fallback parser now parses single-line inline TOML arrays so Python 3.10 (no tomllib) matches tomllib for [paths] array syntax; (2) we now warn (stderr) when a configured agent/skill dir does not exist. Left (3) as-is: setdefault/env-wins is intentional and consistent with the rest of env_loader.

ros2_ws/src/maurice_bot/maurice_cam/maurice_cam/webrtc_streamer.cpp
theo-michel: Addressed the safe subset: async ASYNC-wait fallback (db92251d), the misleading audio=on log (f9e890dc — with_audio is now by-ref so the log reflects a dropped audio branch), and the null check after gst_bin_get_by_name (f9e890dc). Intentionally NOT done blind, leaving this thread open: teardown on DISCONNECTED/CLOSED and a bus watch both need careful threading — set_state(NULL) from a webrtcbin signal thread can deadlock, and there's no GMainLoop here so gst_bus_add_watch won't fire (it needs a ROS-timer bus poll + a session-generation guard so a stale disconnect can't kill a fresh session) — plus on-robot testing; and the hw:0,0 → sysdefault:CARD=<name> device choice needs the real card name. Recommend a focused follow-up PR for those.

theo-michel:
We will need to update the app as well

PR #428, opened 2026-06-06
Quiet ROS node logs for a readable `innate view`

theo-michel:
## Why

`innate view` was dominated by repetitive, per-message, and per-step startup logs that buried anything useful. This makes the node logs **clear and relevant from a new developer's perspective**: idle output is just a periodic vitals heartbeat, and startup shows concise summaries instead of step-by-step play-by-play.

Measured against the live `ros_nodes` tmux session (each change rebuilt → restarted → re-observed):

| | Before | After |
|---|---|---|
| **Idle stream** (75s, all 14 panes) | hundreds of lines | **3 lines** (vitals heartbeat) |
| **Startup** (all panes) | ~1040 lines | **~500 lines** |
| `Publisher already registered` | 14 | 0 |
| `kdl_parser` root-inertia / SRDF / `material in collision` errors | many | 0 |
| zenoh SHM watchdog warning (per node, per boot) | ~8 lines each | 0 |

## What changed (no behavior changes — log levels, dedup, root-cause fixes)

- **Pure noise removed:** duplicate unconditional INFO in bringup `uart.py`/`i2c.py`/`bringup.py`; per-message `rws` "Field 'z' not in json" (launch log-level); per-frame joystick→cmd_vel.
- **Telemetry / per-step init narration → DEBUG** (concise summaries kept at INFO): arm LoopTiming/Gains/GainSched/servo+joint config, camera config dumps, recorder subscribe/service listings, manipulation timing, udp stats, grid-localizer progress, goal-checker tiers, dynamic_footprint geometry, skills loading, mode_manager lifecycle-state spam.
- **Consolidated** `logger_node` vitals into one throttled line (cloud telemetry send unchanged).
- **Root-cause fixes (not suppression):**
  - URDF: removed invalid `<material>` inside `<collision>` blocks.
  - SRDF: anchored `virtual_joint` to the real root link + dropped non-existent passive joints → clears MoveIt `base_root`/`base_x` errors.
  - `CameraProvider`: unique node names → fixes `Publisher already registered`.
  - `config/security/limits.d` rtprio drop-in + `post_update.sh` installer so zenoh's SHM watchdog can set RT priority (priority 48) instead of warning every boot. Ceiling capped at 50 to keep RT blast radius minimal.
- **Genuine warnings/errors stay visible.** Only benign `rcl.logging_rosout` and `kdl_parser` root-inertia notices are suppressed; external nav2 nodes set to `WARN` (following the existing `behavior_server` pattern) — real nav warnings still surface.
- **Launcher:** env now sourced from a `0600` file, so the `INNATE_SERVICE_KEY` and source boilerplate no longer print in every pane (also a small secret-hygiene win).

## Not covered

The remaining startup volume is third-party **stdout** (MoveIt `Loading 'move_group/...'` / GStreamer / NvMM) that ROS log levels can't reach — would require redirecting/filtering process stdout in the launcher, which I left out as too blunt.

## Test plan

- Rebuilt each affected package and restarted nodes; verified subsystems still come up healthy (MoveIt loads model, skills load, arm/nav/cam alive) and that summaries + warnings/errors are preserved.
- ⚠️ The rtprio fix's host side (`/etc/security/limits.d/...`) deploys via `post_update.sh` on the next update; it has **not** been applied to this dev host yet (needs sudo).

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/common/dynamic_loader.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/common/dynamic_loader.py
@@ -142,14 +142,14 @@ class DynamicLoader:
             self.logger.warning(f"Path is not a directory: {directory_path}")
             return result
 
-        self.logger.info(f"Scanning for {kind} in: {directory_path}")
+        self.logger.debug(f"Scanning for {kind} in: {directory_path}")
         for py_file in self._iter_candidate_files(directory):
             try:
                 result.update(self.discover_in_file(py_file))
             except Exception as e:
                 self.logger.error(f"Error loading {kind} from {py_file}: {e}")
 
-        self.logger.info(f"Discovered {len(result)} {kind} in {directory_path}")
+        self.logger.debug(f"Discovered {len(result)} {kind} in {directory_path}")
         return result
 
     def load_from_directories(self, directories: list[str]) -> dict:
--- a/ros2_ws/src/brain/brain_client/brain_client/common/dynamic_loader.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/common/dynamic_loader.py
@@ -149,7 +149,7 @@ class DynamicLoader:
             except Exception as e:
                 self.logger.error(f"Error loading {kind} from {py_file}: {e}")
 
-        self.logger.debug(f"Discovered {len(result)} {kind} in {directory_path}")
+        self.logger.info(f"Discovered {len(result)} {kind} in {directory_path}")
         return result
 
     def load_from_directories(self, directories: list[str]) -> dict:
--- a/ros2_ws/src/brain/brain_client/brain_client/skills/catalog.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/skills/catalog.py
@@ -139,7 +139,7 @@ class SkillRepository:
             try:
                 instance = self._instantiate(skill_class, src_path)
                 code_skills[skill_id] = (display_name, instance)
-                self._logger.info(f"Loaded code skill: {skill_id} ({display_name}) [source={instance.source}]")
+                self._logger.debug(f"Loaded code skill: {skill_id} ({display_name}) [source={instance.source}]")
             except Exception as e:
                 self._logger.error(f"Error instantiating skill {skill_id}: {e}")
         return code_skills
@@ -192,7 +192,7 @@ class SkillRepository:
                         )
                     else:
                         physical_skills[skill_id] = skill_data
-                        self._logger.info(
+                        self._logger.debug(
                             f"Loaded physical skill: {skill_id} (type: {metadata.get('type', 'unknown')})"
                         )
                 except json.JSONDecodeError as e:
@@ -211,7 +211,7 @@ class SkillRepository:
         ensure_user_directories()
         directories = [str(p) for p in get_skill_directories()]
         for directory in directories:
-            self._logger.info(f"Scanning skills directory: {directory}")
+            self._logger.debug(f"Scanning skills directory: {directory}")
         return directories
 
     def _apply_sim_swap(self, id_keyed: dict[str, tuple[str, type, Path]]) -> None:
--- a/ros2_ws/src/brain/brain_client/brain_client/skills/catalog.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/skills/catalog.py
@@ -139,7 +139,7 @@ class SkillRepository:
             try:
                 instance = self._instantiate(skill_class, src_path)
                 code_skills[skill_id] = (display_name, instance)
-                self._logger.debug(f"Loaded code skill: {skill_id} ({display_name}) [source={instance.source}]")
+                self._logger.info(f"Loaded code skill: {skill_id} ({display_name}) [source={instance.source}]")
             except Exception as e:
                 self._logger.error(f"Error instantiating skill {skill_id}: {e}")
         return code_skills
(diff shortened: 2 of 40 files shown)


scripts/launch_ros_in_tmux.sh
```diff
@@ -27,9 +27,18 @@ WINDOW_NAMES=(
...
+# sourcing) into one file the panes source. This keeps the command echoed in
+# each pane short and, importantly, keeps the service key off the screen and
+# out of the scrollback. Mode 600 so the key file isn't world-readable.
+PANE_SETUP_FILE="${TMPDIR:-/tmp}/innate_pane_env.zsh"
+{
+    echo "${RUNTIME_ENV_EXPORTS:-true}"
+    echo "source $DDS_SETUP_SCRIPT"
+    echo "source $ROS_WS_PATH/install/setup.zsh"
+} > "$PANE_SETUP_FILE"
+chmod 600 "$PANE_SETUP_FILE"
```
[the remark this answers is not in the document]
theo-michel: Fixed in 9d81dfe — switched to `mktemp`, so the file is created atomically at mode 0600 with an unpredictable name. That closes both the umask race (the plain `>` briefly created it world-readable before `chmod`) and the predictable-path symlink window. Verified on the device: the new file is `-rw------- /tmp/innate_pane_env.HilLHa`.

ros2_ws/src/maurice_bot/maurice_cam/maurice_cam/stereo_depth_estimator.cpp
```diff
@@ -216,8 +216,8 @@ void StereoDepthEstimator::syncCallback(const sensor_msgs::msg::Image::ConstShar
         return;
 
     if (!vpi_initialized_ || !calibration_loaded_) {
-        RCLCPP_WARN_THROTTLE(this->get_logger(), *this->get_clock(), 10000,
-                             "VPI or calibration not ready, skipping frame");
+        RCLCPP_WARN_ONCE(this->get_logger(),
+                         "Depth disabled: waiting for stereo calibration to load (skipping frames until ready)");
         return;
     }
```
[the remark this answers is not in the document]
theo-michel: Good catch for this node specifically. Rather than revert to the 10s throttle (the periodic repeat is exactly the noise this PR set out to remove), I kept the `WARN_ONCE` "Depth disabled" and added an `INFO_ONCE` "Depth enabled" on recovery (9d81dfe). So the current state is always distinguishable — one line when depth goes down, one when it comes back — without going silent or spamming.

For context, `main_camera_driver` already has the equivalent counterpart: it logs "Stereo calibration (re)loaded" on success (~L735), so its `WARN_ONCE` on the file-watch exception is similarly distinguishable. Left as-is intentionally.

ros2_ws/src/maurice_bot/maurice_sim/urdf/maurice.urdf
```diff
@@ -47,9 +44,7 @@
       <origin xyz="-0.0587 0 0.0838" rpy="0 0 0"/>
       <geometry>
         <box size="0.1878 0.182 0.1676"/>
-      </geometry>
-      <material name="collision_material"/>
-    </collision>
+      </geometry>    </collision>
   </link>
 
   <!-- joint1 → link1 -->
```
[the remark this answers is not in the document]
theo-michel: Fixed in 9d81dfe — split `</geometry></collision>` back onto separate lines across all 9 collision blocks.

ros2_ws/src/brain/brain_client/brain_client/common/dynamic_loader.py
vignesh-anand, line 1: This seems like aone time log not a repeated one and seems like useful info?

ros2_ws/src/brain/brain_client/brain_client/skills/catalog.py
vignesh-anand, line 1: this also seems like once per boot and feels like useful debugging to have, if your skill actually registered?

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
vignesh-anand, line 1: Agree with this debug, but maybe there is a simpler message we can put in info once per policy execution with a summary of the policy execution

ros2_ws/src/brain/manipulation/src/recorder_main.cpp
vignesh-anand, line 1: Seems like a one time useful info?

ros2_ws/src/brain/manipulation/src/recorder_node.cpp
vignesh-anand, line 1: Maybe we optimize this by in info or actually in wanr, saying what topics are missing, instead of in info saying what topics are available?

ros2_ws/src/maurice_bot/maurice_arm/maurice_arm/arm_config.cpp
vignesh-anand, line 1: We moved everything to debug, is it possible to replace to in info put a 2 line summary like all joint intialised or joint x failed.Cause failures do happen every now and then.

ros2_ws/src/maurice_bot/maurice_cam/maurice_cam/stereo_depth_estimator.cpp
vignesh-anand, line 1: would be good to replace all the lines of info  with aybe one consolaidated line saying its running

scripts/launch_ros_in_tmux.sh
vignesh-anand, line 1: The tempfile is never cleaned up — INNATE_SERVICE_KEY sits in /tmp until reboot. Add trap 'rm -f "$PANE_SETUP_FILE"' EXIT INT TERM right after the mktemp call.

PR #430, opened 2026-06-08
ci: split test image into a warm base + thin layer for faster CI

theo-michel:
## What
Speeds up the `integration-test` CI (GitHub Actions → Cloud Build) from ~6 min toward ~1–2 min on a typical PR, **without** reducing build coverage.

## Why
Every run rebuilt the whole image from scratch: ~180s to apt/pip-install ~1000 ROS packages **and** a full `colcon build`. The `build-test-image` step had no `--cache-from`, and Cloud Build VMs are ephemeral, so the BuildKit `--mount=type=cache` mounts don't persist between runs.

## How
Mirrors the existing `sim/Dockerfile` + `Dockerfile.ros-prebuilt` + `cloudbuild.sim-images.yaml` pattern:

- **`ci/Dockerfile.test-base`** (new) — warm base: apt/pip deps, rosdep, a full prebuilt `ros2_ws` (`build/`+`install/`), and a populated **ccache**. Pushed to `ghcr.io/innate-inc/innate-os-test-base:{main,buildcache}` **only on main** (`_PUBLISH_BASE=true`).
- **`ci/Dockerfile.test`** — now a thin `FROM`-the-base layer: overlays this revision's source and does an **incremental, ccache-backed** `colcon build`. Only changed packages recompile; the whole tree still builds, so a build error anywhere fails CI.
- **`ci/cloudbuild.integration-test.yaml`** — bash step that logs into ghcr, publishes the base on main / reuses it on PRs (with an inline-build fallback for cold start), then builds the thin image and runs the tests. Reuses the existing `innate-os-sim-ghcr-*` secrets. `colcon` parallelism `2 → $(nproc)`.
- **`ci/run_integration_tests.sh`** — dropped the redundant second `colcon build` (the image already built the workspace).
- Docs (`ci/README.md`, `docs/TESTING.md`) updated for the two-step local build.

Why a *published* base (not just `--cache-from`): on ephemeral VMs a cache-missed `colcon` layer can never see the ccache it would itself produce, so only a base used as `FROM` carries a warm cache across runs.

## Repo size
Unaffected — nothing built is committed. `build/`/`install/`/ccache live only in Docker and ghcr (already covered by `.gitignore`); this PR adds two small text files.

## Rollout / prerequisites
- The base doesn't exist until the **first push to main** after merge; until then PRs build it inline (no slower than today).
- The Cloud Build service account needs **read access to the `innate-os-sim-ghcr-*` secrets**, and the private worker pool needs **egress to ghcr.io**.

## Verification
Real end-to-end needs a Cloud Build run. Locally (linux/amd64): build the base, build the thin against unchanged src and confirm `ccache -s` shows ~100% hits; touch one `.cpp` → only that package recompiles; introduce a deliberate error in `maurice_cam` → image build still **fails** (gate preserved).

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -48,10 +48,17 @@ jobs:
           if [ -n "${{ vars.GCP_CLOUD_BUILD_WORKER_POOL }}" ]; then
             pool_flag="--worker-pool=${{ vars.GCP_CLOUD_BUILD_WORKER_POOL }}"
           fi
+          # Publish (and push) the warm base image only on main; PRs reuse it.
+          # See ci/cloudbuild.integration-test.yaml for the two-image strategy.
+          publish_base="false"
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
+            publish_base="true"
+          fi
           gcloud builds submit \
             --project="${{ vars.GCP_PROJECT_ID }}" \
             --region="${{ vars.GCP_REGION }}" \
             --gcs-source-staging-dir="${{ vars.GCP_CLOUD_BUILD_SOURCE_STAGING_DIR }}" \
             --service-account="projects/${{ vars.GCP_PROJECT_ID }}/serviceAccounts/${{ vars.GCP_CLOUD_BUILD_SERVICE_ACCOUNT }}" \
             ${pool_flag} \
+            --substitutions=_PUBLISH_BASE="${publish_base}" \
             --config=ci/cloudbuild.integration-test.yaml
--- a/ci/Dockerfile.test
+++ b/ci/Dockerfile.test
@@ -1,106 +1,44 @@
 # syntax=docker/dockerfile:1.7
 #
-# ci/Dockerfile.test - Test image for CI on non-Jetson platforms.
-# Skips Jetson-specific packages (nvidia-*, cuda-*, etc.)
+# ci/Dockerfile.test - Thin CI test image (built per PR/commit).
 #
+# Builds on the warm base (ci/Dockerfile.test-base, published as
+# ghcr.io/innate-inc/innate-os-test-base:main): apt/pip deps, rosdep, a full
+# prebuilt ros2_ws (build/ + install/), and a populated ccache are all baked in.
+# This layer overlays the current revision's source and does an INCREMENTAL,
+# ccache-backed colcon build, so only packages that actually changed do real
+# compile work. The whole workspace is still built, so a build error in any
+# package fails CI (the build gate is preserved).
 
-FROM ros:humble-ros-base
-
-ENV DEBIAN_FRONTEND=noninteractive \
-    ROS_DISTRO=humble \
-    INNATE_OS_ROOT=/root/innate-os \
-    UV_SYSTEM_PYTHON=1 \
-    UV_LINK_MODE=copy \
-    PIP_DISABLE_PIP_VERSION_CHECK=1
+ARG BASE_IMAGE=ghcr.io/innate-inc/innate-os-test-base:main
+FROM ${BASE_IMAGE}
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-# Install base prerequisites + enable universe (for gstreamer, libhdf5-dev, etc.)
-RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
-    --mount=type=cache,target=/var/lib/apt/lists,sharing=locked \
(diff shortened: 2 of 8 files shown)


vignesh-anand:
## Code Review

Good architectural idea overall — the warm base + thin layer split is a clean approach to the problem, and the fallback logic is thoughtful. A few issues worth addressing before merging.

---

### Critical

**1. `ccache` may not be installed in the base image.**

`ci/Dockerfile.test-base` sets `CCACHE_DIR=/root/.ccache` and passes `-DCMAKE_C_COMPILER_LAUNCHER=ccache` / `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache` to CMake, but the `apt-get install` block doesn't explicitly install `ccache`. If it's not already in `ros2_ws/apt-dependencies.common.txt`, every compilation will fail with `ccache: command not found`. Worth verifying explicitly and/or adding it to the `apt-get install` block in the base Dockerfile.

**2. Changes to `apt-dependencies.common.txt` or `pip-requirements.txt` in a PR are silently ignored.**

The base bakes those files at main-build time. If a PR modifies either (e.g. adds a new pip package needed for a new test), the thin layer doesn't re-run `apt-get install` or `uv pip install` — it just uses whatever was in the base. CI can pass or fail for the wrong reasons with no diagnostic. This is distinct from the *new ROS system dependency* case that `rosdep install` in the thin layer handles — it specifically affects hand-curated additions to those two files. This case isn't mentioned anywhere in the PR description or comments.

---

### Significant

**3. Deleted source files can produce false CI passes.**

If a PR deletes a `.cpp` or `.h` file, the compiled artifact still exists in the base's `build/`+`install/` directories. Docker `COPY` merges, it does not sync. The incremental `colcon build` in the thin layer may see no changed inputs and skip re-linking, so CI passes with a binary that includes dead code — or silently masks a linker error that would catch a missing symbol. The PR acknowledges this in a comment ("Low risk for a build gate; rebuild the base if it matters") but it understates the severity: a PR that deletes a file used by another package could pass CI while the actual post-merge main build breaks.

**4. `rosdep install ... || true` contradicts the comment's promise of "clear diagnostics."**

The comment in `ci/Dockerfile.test` says:

> *"on a genuinely missing dep, fail with a clear rosdep diagnostic instead of a cryptic compiler error"*

But `|| true` suppresses rosdep's non-zero exit. If rosdep can't find a package, its error is silently discarded and then `colcon build` fails with the cryptic compiler error the comment claims to avoid. The comment is actively misleading — I'd either remove the claim or remove the `|| true` (accepting that rosdep occasionally exits non-zero for non-fatal reasons).

**5. `--parallel-workers "$(nproc)"` risks OOM on `E2_HIGHCPU_8`.**

The original was `--parallel-workers 2`; this PR changes it to `$(nproc)`, which is 8 on that machine type — a 4× increase. `--parallel-workers` controls how many packages build simultaneously. With heavy packages like `maurice_cam`, 8 concurrent C++ link jobs can exhaust the 8 GB RAM on `E2_HIGHCPU_8` with no useful error message. Since ccache handles the compile-time savings anyway, the parallelism mainly matters for the cold base build. Something like `min($(nproc), 4)` or a fixed `4` would be safer.

---

### Minor

**6. All logic in one Cloud Build step hurts observability.**

The `build-test-image` step in `cloudbuild.integration-test.yaml` is ~70 lines of inline bash. Cloud Build only shows pass/fail at the step level. Splitting into a `pull-or-build-base` step and a `build-thin-image` step would make failures much easier to diagnose.

**7. GHCR token scope assumption.**

The PR reuses `innate-os-sim-ghcr-token`/`innate-os-sim-ghcr-username` (originally scoped for sim image writes) to push to the new `innate-os-test-base` repo. If that token is fine-grained to the sim repo only, the push will fail silently on the first main build. Worth confirming the token has write access to `innate-os-test-base` before merging.

**8. Inline fallback base build on PRs is semantically wrong.**

If `docker pull "$base_image:main"` fails on a PR (e.g. first run ever), `build_base()` is called inline using the PR's own `ros2_ws/src`. That means the "base" contains this PR's source, and the thin layer then COPYs the same files on top again. It works, but the base no longer represents a main-branch state. Low impact, but the first PR after this lands will give misleading build time signals.

---

**TL;DR:** #1 (ccache not installed) and #2 (stale apt/pip deps on PR builds) are the ones I'd fix before merging. The rest are worth at minimum a doc note so future contributors understand the constraints.

vignesh-anand (commented):
left comments

theo-michel:
Ok fixed!

PR #432, opened 2026-06-11
Enable disk-backed swap file in post_update (INN-530)

theo-michel:
## What

Adds an idempotent step to `scripts/update/post_update.sh` that creates an **8 GB NVMe-backed swap file** (`/swapfile`), enables it, and persists it in `/etc/fstab`.

Because `post_update.sh` runs after every OTA, **all robots gain swap on their next update with no manual action** — which is the point of [INN-530](https://linear.app/innate/issue/INN-530/enable-swap).

## Why

Jetson ships only **zram** (compressed *RAM*-backed swap). zram adds no real capacity once physical RAM is exhausted — it just compresses pages in the same 7.4 GB. That's the cause of the OOM kills seen during the hackathon (INN-515, duplicated into this ticket) and during the memory-heavy `colcon` rebuild that runs later in this same script.

A disk-backed swap file adds a genuine overflow tier. The kernel fills high-priority zram first (prio 5), then spills to the lower-priority disk swap.

## Details

- **Placement:** runs before the ROS workspace rebuild, so the *same* update run that ships this already benefits from the extra headroom.
- **Idempotent:** detects already-active swap and an existing fstab entry; safe to run on every update.
- **Non-fatal:** swap setup must never block an update, so any failure only logs a warning and the update continues. `fallocate` falls back to `dd` on filesystems that would produce a sparse file.

## Testing

- `bash -n` syntax check passes.
- Isolated harness with stubbed `swapon`/`mkswap`/`fallocate` verified: fresh setup, idempotent re-run (exactly one fstab line, no duplicates), and the failure path (warns + continues, exit 0).

Note: not yet exercised on a live robot through a full `innate update apply` — worth a smoke test on one device before broad rollout.

### DIFF
--- a/scripts/update/post_update.sh
+++ b/scripts/update/post_update.sh
@@ -230,38 +230,44 @@ SWAP_FILE="/swapfile"
 SWAP_SIZE_GB=8
 
 setup_swap() {
-    # Already active? Nothing to do (the common case on every update after the first).
     if swapon --show=NAME --noheadings 2>/dev/null | grep -qx "$SWAP_FILE"; then
+        # Already active (the common case on every update after the first).
+        # Still fall through to the fstab check below: an earlier run interrupted
+        # between swapon and the fstab append leaves swap active but unpersisted.
         log "  Swap file already active ($SWAP_FILE)"
-        return 0
-    fi
-
-    # Not active. Any existing file may be a partial leftover from an interrupted
-    # run (power loss, disk-full, OOM-killed dd); a truncated file would make
-    # mkswap fail on every future update. So never trust an existing file:
-    # rebuild from scratch, and remove it on any failure so a botched run can't
-    # permanently block swap setup on later updates.
-    rm -f "$SWAP_FILE"
-    log "  Creating ${SWAP_SIZE_GB}G swap file at $SWAP_FILE..."
-
-    # Create with mode 600 up front (never world-readable), then allocate.
-    # fallocate works on the ext4 NVMe root; fall back to dd where unsupported.
-    install -m 600 /dev/null "$SWAP_FILE" || return 1
-    if ! fallocate -l "${SWAP_SIZE_GB}G" "$SWAP_FILE"; then
-        log "  fallocate unavailable; falling back to dd"
-        dd if=/dev/zero of="$SWAP_FILE" bs=1M count=$((SWAP_SIZE_GB * 1024)) status=none \
-            || { rm -f "$SWAP_FILE"; return 1; }
-    fi
-
-    if ! mkswap "$SWAP_FILE" >/dev/null || ! swapon "$SWAP_FILE"; then
+    else
+        # Not active. Any existing file may be a partial leftover from an
+        # interrupted run (power loss, disk-full, OOM-killed dd); a truncated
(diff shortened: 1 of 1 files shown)


theo-michel:
Addressed both review findings in d3ee3d0d:

- **P1 (partial file breaks idempotency):** when swap isn't active, an existing `/swapfile` is now never trusted — it's removed and rebuilt from scratch, and removed again on any failure (`dd`/`mkswap`/`swapon`). A run interrupted by power loss / disk-full / OOM no longer leaves a truncated file that makes `mkswap` fail forever; the next update self-heals.
- **P2 (world-readable window):** the file is created with `install -m 600 /dev/null "$SWAP_FILE"` before any allocation, so it's never world-readable. The separate `chmod 600` is dropped.

Verified with a stubbed harness covering fresh setup, idempotent re-run (one fstab line), `mkswap`-failure cleanup (no leftover file), partial-leftover recovery, and the `dd` fallback path.

theo-michel:
Second review pass surfaced three fstab-handling gaps, fixed in 188850f9:

1. **Trailing-newline guard** — if `/etc/fstab` lacks a trailing newline, a bare `echo >>` glues the swap entry onto the previous mount line (empirically confirmed; worst case corrupts the root mount entry). The append now adds a newline first when missing.
2. **Persistence on the already-active path** — a run interrupted between `swapon` and the fstab append left swap active but unpersisted, and every later run early-returned without ever persisting it. The fstab check now runs on both paths.
3. **`nofail` in the entry** — a stale fstab line pointing at a missing `/swapfile` (possible after a failed re-setup removed the file) no longer produces a degraded swap unit at boot; it's tolerated until the next successful run recreates the file.

Harness verified: non-trailing-newline fstab append keeps the previous entry intact, active-but-unpersisted gets its entry added, failure + recovery yields exactly one entry, idempotent re-runs don't grow fstab, and empty-fstab append works.

theo-michel:
**Live-tested on a real Jetson (Orin, JetPack ext4 NVMe root)** — ran the swap block extracted verbatim from this branch, twice in a row as root:

- 1st run: created `/swapfile` (8G, mode 600 from creation), `mkswap` + `swapon` succeeded, fstab entry added.
- 2nd run: detected active swap, skipped creation, no duplicate fstab line.

Post-run state: `swapon --show` lists `/swapfile file 8G prio -2` (below zram's prio 5, as intended), exactly one fstab entry (`/swapfile none swap sw,nofail 0 0`), total swap 11Gi. This closes the last open item from review — the change is verified end-to-end on target hardware.

vignesh-anand (commented):
Looks good overall — the idempotency design is solid and the edge cases (interrupted runs, trailing-newline in fstab, already-active-but-unpersisted) are well handled. A couple of things still worth considering:\n\n**Disk-space check before allocation.** If the NVMe is nearly full, dd will fill it before failing, which could corrupt other state.\n\n**dd status=none silences errors on some versions of dd.** If dd fails silently the swap file is empty and mkswap errors with a confusing message. Consider dropping status=none or keeping stderr visible.\n\nEverything else looks addressed — install -m 600 is intentional and correct, priority falls naturally below zram prio 5, and the live Jetson test closes the remaining open item. LGTM pending the disk-space check.

PR #434, opened 2026-06-11
Surface a clear error when the lidar stops publishing (INN-474)

theo-michel:
## Summary

Fixes [INN-474](https://linear.app/innate/issue/INN-474/if-lidar-isnt-connected-agent-fails-silently-should-degrade-gracefully): when the lidar is disconnected, no `/scan` data flows → AMCL never publishes `map→base_link` → `pose_source()` returns `None` → the brain's `agent_loop()` returns early on every 10 Hz tick with only a log warning. The agent appears frozen with nothing reaching the app.

This adds detection and a clear user-facing error (full graceful degradation can come later, per the ticket):

- **`perception/scan_health.py`** (new): `ScanHealthMonitor` subscribes to `/scan` (sensor-data QoS, compatible with both the best-effort driver and the reliable throttle relay) and records the last arrival time; stale after 10 s without a scan (normal rate ~6 Hz).
- **`core/orchestrator.py`**: `check_lidar_health()` runs every agent tick. On a stale scan it logs an error and emits one system chat message per disconnect episode to `/brain/chat_out` (the app's chat channel), plus a recovery message when scans resume. Suppressed in mapfree mode and the simulator, where the lidar isn't required.
- **`core/config.py` / `nodes/brain_client_node.py`**: new `scan_topic` parameter (default `/scan`) and wiring.

The app now shows:

> ⚠️ LiDAR is not responding (no laser scan data has been received since startup). The robot cannot localize or navigate until it recovers. Check that the LiDAR is connected and spinning.

## Testing

- Unit tests (staleness tracking, emit-once-per-episode, recovery, mapfree/simulator suppression): 9/9 passed. Written for verification and removed before merge per review flow (see reverted commit in history).
- Live on a robot (zenoh RMW, full stack running): monitor on the real `/scan` stayed healthy; monitor on a publisher-less topic went stale as designed.
- End-to-end with real ROS pub/sub: real `Orchestrator` + `ChatManager`, simulated disconnect → exactly one error message on chat-out; scans resumed → exactly one recovery message.
- Full fast CI bucket (88 tests) passes; lint clean.


vignesh-anand (commented):
Doesn't it make more sense to have all the robot sensor monitoring in bot and not brain?

theo-michel:
It's for communication with the app, to show in the chat interface that the lidar is not working and that that is why the agent is stuck

theo-michel:
Will merge I think it's safe

PR #437, opened 2026-06-12
feat: persistent robot microphone toggle (INN-485)

theo-michel:
## Summary

There was no easy way to disable the robot's microphone ([INN-485](https://linear.app/innate/issue/INN-485/robot-microphone-should-be-easily-deactivatable)). This adds a robot-side mute that is the source of truth — it survives robot reboots and works regardless of app state — following the same pattern as the speaker volume setting (INN-467).

- **`app_control`** (`mars_control/app.cpp`): new `/set_microphone` service (`std_srvs/SetBool`) persists `microphone_enabled` in `robot_info.json`, mirrors it in `/robot/info`, and broadcasts it on a latched `/microphone_enabled` Bool topic (also published at startup so the setting is re-applied after reboot).
- **Input manager** (`brain_client`): subscribes to `/microphone_enabled` (TRANSIENT_LOCAL QoS) and gates the `micro` input device. Disabling closes the device immediately — the arecord stream and the OpenAI Realtime connection are torn down, so the robot truly stops listening, not just ignoring transcripts. Re-enabling reopens the mic if the active directive requests it (requested inputs are now tracked across toggles).
- TTS output is unaffected: the robot can still speak while muted.

Companion app PR adds a discrete mute button on the agent screen that calls the service and reflects `robot_info.microphone_enabled`.

## Test plan
- [x] `pre-commit` (ruff + clang-format, `.config/pre-commit-config.yaml`) passes on changed files
- [x] `mars_control` compiles in the CI test image (`colcon build --packages-select mars_msgs mars_control`)
- [x] `brain_client.inputs.manager` / `brain_client.nodes.input_manager` import cleanly in a ROS2 Humble env
- [ ] On robot: `ros2 service call /set_microphone std_srvs/srv/SetBool '{data: false}'` → mic device closes (input_manager logs "💤 Closed input device: micro"), `robot_info.json` gains `"microphone_enabled": false`, state persists across reboot, re-enable reopens mic while a directive is running

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
@@ -139,10 +139,13 @@ class InputDeviceManager:
 
     def set_mic_enabled(self, enabled: bool) -> None:
         """Apply the robot-level microphone toggle."""
-        if enabled == self._mic_enabled:
-            return
+        if enabled != self._mic_enabled:
+            self._logger.info(f"🎙️ Microphone {'enabled' if enabled else 'disabled'}")
+        # Track the user's intent unconditionally (it gates handle_active_inputs
+        # via _is_allowed), and re-apply even when unchanged: _apply is a no-op
+        # when state already matches, so a re-publish of the same value retries
+        # a previously failed open/close instead of being swallowed.
         self._mic_enabled = enabled
-        self._logger.info(f"🎙️ Microphone {'enabled' if enabled else 'disabled'}")
         device = self.input_devices.get(MIC_DEVICE_NAME)
         if device is not None:
             self._apply(MIC_DEVICE_NAME, device, enabled and MIC_DEVICE_NAME in self._requested_inputs)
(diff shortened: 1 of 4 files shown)


ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
vignesh-anand, line 1: When re-enabling the mic, set_mic_enabled(True) calls _apply("micro", device, True). Inside _apply:


device.set_active(True)   # marks the device as active in Python
device.on_open()          # actually starts arecord + OpenAI connection
If on_open() crashes, the device object thinks it's open (is_active() == True) but the actual stream never started. Now the manager's internal state is out of sync with reality.

The consequence: the next call to _apply for this device sees was_active = True, so the if should_be_active and not was_active branch is skipped — it thinks the device is already open and does nothing. The mic is permanently stuck in a broken half-open state until the node restarts.

The fix is one line — roll back the flag if on_open() fails:


except Exception as e:
    if should_be_active:
        device.set_active(False)  # undo the flag
    self._logger.error(...)
    return False

PR #441, opened 2026-06-12
innate update status: warn when installed systemd units drift from the checkout

theo-michel:
## Problem

`innate update apply` re-installs systemd units via `post_update.sh`, but a manual `git pull` on a robot bypasses that — the checkout moves, `/etc/systemd/system` doesn't. Found in the field: a robot whose checkout had been manually merged onto the maurice→mars rename had its **ble-provisioner crash-looping ~4,000 times** (installed unit still ran `maurice_bot/.../simple_bt_service.py`, which no longer exists). The only symptom was the robot silently vanishing from the controller app's BLE scanner; `systemctl is-active` even reported `active` between restarts.

## Fix

`innate update status` now renders each `config/systemd/*.service` with the same `User=`/home templating `post_update.sh` applies and compares it against the installed copy. On any mismatch or missing unit it prints:

```
⚠ Installed systemd units differ from this checkout: ble-provisioner.service
  Services may be running stale paths (a manual `git pull` skips
  unit installation). Run 'innate update reinstall' to fix.
```

Detection only — nothing is changed automatically; the remedy remains the existing `innate update reinstall`.

## Validation

- `python3 -m py_compile scripts/innate` clean
- Helper smoke-tested in isolation: on a machine with no installed units it lists every repo unit as drifted; identical rendered content reports clean

### DIFF
--- a/scripts/innate
+++ b/scripts/innate
@@ -3,6 +3,7 @@
 
 import json
 import os
+import pwd
 import re
 import shlex
 import shutil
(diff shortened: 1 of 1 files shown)


theo-michel:
Addressed in 4fccb97e — home is now resolved via `pwd.getpwnam(user).pw_dir` to match post_update.sh's `eval echo ~$ACTUAL_USER`, with `Path.home()` only as the fallback when the user isn't in passwd.

vignesh-anand (approved):

PR #444, opened 2026-06-16
perf: arm camera VIC offload + headless boot + drop desktop daemons

theo-michel:
Resource optimizations from a performance audit of the robot (Orin Nano, 7.6 GB RAM — memory is the binding constraint). No change to robot behavior.

---

## 1. Arm camera: offload YUY2→BGR conversion to the VIC engine

The arm camera driver used a software `videoconvert` to turn YUY2 (YUYV) frames into BGR for OpenCV — per-pixel YUV→RGB math on the CPU every frame at 640×480@30. This moves it onto the Jetson **VIC hardware engine** via `nvvidconv` (the pattern `main_camera_driver` already uses), leaving only a cheap BGRx→BGR stride drop on the CPU.

```
- videoconvert ! video/x-raw,format=BGR ! appsink
+ nvvidconv ! video/x-raw,format=BGRx ! videoconvert ! video/x-raw,format=BGR ! appsink
```

**Measured (isolated A/B, /dev/video0, 640×480@30 YUYV, 2 passes):**

| Pipeline | CPU (one core) |
|---|---|
| OLD — software `videoconvert` | 18.0% / 17.9% |
| NEW — `nvvidconv` (VIC) | 5.5% / 5.5% |

Confirmed live: the arm `v4l2src` thread in the running `camera_container` dropped **18.3% → 5.8%** — **~69% reduction (~12.5 pts of a core)**, freed continuously. `nvvidconv` YUY2→BGRx validated on-device; camera confirmed publishing after the change. (`GR3D` stays 0% — the VIC is separate fixed-function silicon from the GPU 3D core.)

---

## 2. Headless (no-GUI) boot on display-less robots

Robots ship on the stock Ubuntu **desktop** image and boot to `graphical.target` — a full GNOME session (gnome-shell, Xorg, gdm, ~25 `gsd-*` daemons, gvfs, tracker) costing **~0.8 GB RAM** plus idle CPU, on a machine where no display is attached (`card1-DP-1: disconnected`).

`post_update.sh` now switches the default boot target to `multi-user.target` and stops gdm.

- **Guarded** — only switches when no DRM connector reports a connected display, so an engineering kit on a monitor keeps its GUI.
- **Reversible** — gdm is stopped, never masked; `systemctl set-default graphical.target` restores it.
- **Idempotent & non-fatal** — re-running is a no-op once headless; failures only warn.

Verified: no prior GUI-disable step ever existed in `post_update` history (nothing reverted), nothing in the runtime depends on an X session, and the one incidental `After=graphical.target` in `arducam-audio.service` was already removed in the ALSA migration.

---

## 3. Disable desktop/update daemons with no robot function

The desktop image also runs daemons that serve no purpose headless and only cost RAM + periodic CPU wakeups. `post_update.sh` now disables: `packagekit`, `fwupd`, `ModemManager`, `cups`/`cups-browsed`/`lpd`, `colord`, `switcheroo-control`, `kerneloops`, `ubuntu-advantage`, and `snapd`. (~150–200 MB + polling.)

- **Fleet-safe** — each unit is disabled only if present (idempotent, tolerant of image variants); non-fatal (a failure only warns).
- **snapd guard** — disabled only when no snaps are installed, so a robot shipping a snap is never broken.
- **Reversible** — `systemctl enable --now <unit>`.
- Verified on-device: no snaps, no modem, no NFS. (Session-scoped cruft like gnome-software / update-notifier / tracker is already gone via the headless boot in §2.)

---

## Combined impact

~0.9–1.1 GB RAM reclaimed (§2+§3) + ~12.5 pts of a CPU core (§1), continuously, with no effect on robot performance.

## Still deferred (not in this PR)

- **2 leaked zombie processes** (`gst-play`/`rm`) from a child-reaping gap — cosmetic, worth fixing at the source.


karmanyaahm (approved):
makes sense, i'm surprised we didn't already disable gdm

PR #447, opened 2026-06-16
Speed up local ACT policy: faster loading + optional TensorRT inference

theo-michel:
Speedups + motion smoothing for running learned ACT policies locally on the Jetson Orin, plus the per-robot provisioning, settings, and webapp changes that support it.

## 1. Faster policy loading (~12.6s → ~3.6s)
- Remove `torch.compile` from the load path; it has no Triton backend here and gave 0% speedup while costing ~8s.
- Single host→device transfer: checkpoint to CPU (`weights_only=True, mmap=True`), build on CPU, `load_state_dict(assign=True)`, then one `.to(device)`.
- Single warmup pass.

## 2. TensorRT inference — the only path (fp32)
A fused TRT engine runs the ACT forward far faster than eager; `torch.compile`/fp16-autocast/bf16-autocast all gave no speedup at batch=1.

| Approach | latency | RAM | error vs fp32 |
|---|---|---|---|
| eager fp32 (removed; baseline) | 63.7 ms | ~1174 MB | — |
| **TRT fp32 (default)** | **6.6 ms (~10×)** | ~707 MB | 0.024% RMSE |
| TRT fp16 (optional) | 3.2 ms (~20×) | ~619 MB | 0.077% RMSE |

- **`act_trt.py`**: `build_engine()` (ONNX export → cached engine next to the checkpoint, scoped by TRT version / precision / arch dims) and `TRTACTPolicy` (`select_action`/`reset`, holds no torch weights).
- **The only inference path** in `manipulation_server`: the eager `ACTPolicy` is built solely to export the engine, then released. **TensorRT is a hard requirement** — eager `select_action` was removed, so there is no fallback. If `tensorrt` can't import or the engine build fails, the policy load fails loudly (it does not run eager). This means a unit that hasn't run the `post_update` provisioning below, or a dev/sim box without JetPack/CUDA, cannot run learned policies.
- **Pre-built in the model-download flow** (`innate_training_node`) so the ~1 min build is paid at download time, not first run.

## 3. Smooth motion via temporal ensembling
The model ran once per chunk (~1Hz); without ensembling each replan caused a visible action discontinuity. TRT's per-step headroom (~10 ms incl. ensembler, well under the 40 ms budget) lets us run the model **every step** and blend overlapping chunk predictions (`ACTTemporalEnsembler`, reused verbatim) — removing both the boundary jump and the periodic timing spike.
- Enabled by default on the TRT path (`temporal_ensemble_coeff`, default 0.01; set 0 for the chunked queue path).
- Validated: TRT vs eager temporal-ensemble match within 2e-4 over 120 steps.
- Also resets the policy after warmup so execution starts from a clean state.

## 4. Per-robot provisioning (`post_update`)
- apt (`apt-dependencies.hardware.txt`): `tensorrt`, `python3-libnvinfer`, `nvidia-l4t-dla-compiler`.
- pip (`pip-requirements.txt`): `onnx`.
- `ldconfig` after apt install — `libnvdla_compiler.so` (which libnvinfer dlopens at `import tensorrt`) lands in `/usr/lib/aarch64-linux-gnu/nvidia`; the cache is stale until ldconfig runs. Without it, `import tensorrt` fails and the policy can't load (TRT is required).

## 5. Base-speed scale, replay fallback, and config defaults
- **Independent base-speed tuning** (`config_loader` + `settings.yaml.template`): `replay_base_speed_scale` (default 1.0, plays recorded `cmd_vel` back 1:1) and `learned_base_speed_scale` (default 0.5, de-rates the learned policy). Applied at execution time in `manipulation_server`.
- **Replay raw_data fallback** (`brain_client/skills/catalog.py`): when the background dataset encoder has moved a freshly-saved recording into `../raw_data/`, replay resolution falls back there (same action data), closing the window where the episode isn't yet under `data/`.
- **Template default syncs**: `temporal_ensemble_coeff` 0.01, updated cartesia voice id, `consecutive_stops_to_complete` 30.

## 6. Webapp: collapsible settings page
- Redesign the Settings page as a collapsible divider list — each catalog section is a header that expands its knobs (collapsed by default), with groups holding an override auto-expanded and an unsaved-changes dot on the header. Expands the knob catalog to cover the new settings.

## Test plan
- [x] Load-time, TRT latency/RAM/accuracy, and engine build-time measured on-device.
- [x] TRT (chunked and temporal-ensemble) outputs match eager within fp32 precision over multi-step sequences.
- [x] Per-step TRT+ensemble latency ~10 ms (fits 25 Hz).
- [x] Reviewer: rebuild + restart the node; confirm "Using TensorRT engine for inference (temporal_ensemble_coeff=0.01)" and smooth motion (no per-replan jerk).


ros2_ws/src/brain/manipulation/manipulation/act_config.py
```diff
@@ -0,0 +1,90 @@
+"""ROS-free factory for the ACT inference config.
+
+Kept separate from manipulation_server so the offline engine pre-build
+(`python -m manipulation.act_trt`, spawned by innate_training_node) can build
+the exact same architecture without importing the ROS node (rclpy, cv_bridge,
+mars_msgs, ...). manipulation_server re-exports this for its own use.
+"""
+
```
theo-michel, line 8: I don't like this line

ros2_ws/src/brain/manipulation/manipulation/act_config.py
```diff
@@ -0,0 +1,90 @@
...
+    if n_action_steps is None:
+        effective_n_action_steps = default_n_action_steps
+    else:
+        try:
+            # Reject bools (bool is an int subclass in Python) and anything
+            # that isn't a numeric type outright, so malformed metadata
+            # (e.g. a string or NaN) doesn't crash policy loading.
+            if isinstance(n_action_steps, bool) or not isinstance(n_action_steps, (int, float)):
+                raise TypeError(f"expected int or float, got {type(n_action_steps).__name__}")
+            # int() on NaN -> ValueError, on inf -> OverflowError, both caught below.
```
theo-michel: This is ugly we can make simpler code

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
```diff
@@ -0,0 +1,350 @@
...
+        # would run on the default stream while the engine ran on self._stream with no
+        # ordering between them -- the engine could read the input buffers before the copies
+        # land, yielding actions computed from stale/partial data.
+        self._stream.wait_stream(torch.cuda.current_stream())
+        with torch.cuda.stream(self._stream):
+            self._buffers["image_camera_1"].copy_(batch["observation.image_camera_1"])
+            self._buffers["image_camera_2"].copy_(batch["observation.image_camera_2"])
+            self._buffers["state"].copy_(batch["observation.state"])
+            self._context.execute_async_v3(self._stream.cuda_stream)
+        self._stream.synchronize()
```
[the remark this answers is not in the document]
theo-michel: It's fixed

karmanyaahm (changes requested):
Image

the commits are labeled with the old robot user. you should rebase.

Other than that it looks good to me, this is very big so idk if i would've even caught anything, but if it works it works

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
```diff
@@ -0,0 +1,375 @@
...
+    image_shape = policy.config.input_shapes["observation.image_camera_1"]
+    state_dim = policy.config.input_shapes["observation.state"][0]
+    onnx_path = engine_path + ".onnx"
+    dummy = (
+        torch.zeros(1, *image_shape, device=device),
+        torch.zeros(1, *image_shape, device=device),
+        torch.zeros(1, state_dim, device=device),
+    )
+
+    tmp_path = engine_path + ".tmp"
```
vignesh-anand: **Concurrent builds of the same engine race on shared scratch paths**

`onnx_path` and `tmp_path` are deterministic (`engine_path + ".onnx"` / `+ ".tmp"`), so two builds of the *same* checkpoint write to the same scratch files. At node startup both build triggers can fire at once — the startup sweep (`prebuild_all_engines`) and the on-download pre-build (`_prebuild_trt_engine`) — and the `if os.path.exists(engine_path): return` fast-path does not protect the **first-ever** build, which is exactly that window. One process can `torch.onnx.export` (truncate+rewrite) the `.onnx` while the other is mid-`parser.parse()` on it, so one build spuriously fails with a corrupt-parse `RuntimeError` (swallowed as a warning in `workers.py`).

The final engine write is already atomic via `os.replace`, so the fix is just to make the scratch paths per-process and let last-writer-win (both builds produce an identical engine):

```python
scratch_prefix = f"{engine_path}.{os.getpid()}"
onnx_path = scratch_prefix + ".onnx"
tmp_path = scratch_prefix + ".tmp"
```

(The shared timing-cache write at `_timing_cache_path()` races too, but it is self-healing — a corrupt read is caught and falls back to a full build — so it only costs a slow build, not a failure.)

ros2_ws/src/cloud/innate_training_node/innate_training_node/workers.py
```diff
@@ -294,3 +369,53 @@ def maybe_auto_download(
...
+    Complements the on-download pre-build, which only fires for a fresh download.
+    This covers checkpoints already on disk (downloaded before pre-build existed,
+    shipped skills) and engines invalidated by a TensorRT upgrade (the engine cache
+    key is version-scoped). Best-effort and idempotent -- the builder is a fast
+    no-op when the engine already exists.
+    """
+    root = os.environ.get("INNATE_OS_ROOT", os.path.expanduser("~/innate-os"))
+    scan_roots = [
+        os.path.join(root, "workspace", "innate_skills"),
+        os.path.join(root, "workspace", "custom_skills"),
```
vignesh-anand: **Sweep hardcodes a subset of the recognized skill roots — misses user-configured (and legacy) skill dirs**

`scan_roots` only covers `workspace/{innate_skills,custom_skills}`, but the system loads and executes skills from more locations than that. The canonical loader list is `script_paths.get_skill_directories()`:

```
required:  innate_skills, custom_skills, *get_extra_skill_dirs()   # user-configured, "anywhere on the machine"
optional:  $INNATE_OS_ROOT/skills (legacy), ~/skills               # if present
```

`get_extra_skill_dirs()` comes from `config/os.toml [paths].skill_dirs` (→ `INNATE_EXTRA_SKILL_DIRS`), so a user can put runnable skills in an arbitrary dir. A skill there is fully executed by the robot, and its checkpoint can be downloaded to via `req.skill_dir` (the download dest is caller-supplied and only validated as absolute — it is *not* constrained to these two roots).

Consequence: the sweep — whose whole purpose is to catch engines the on-download path missed (skill predates pre-build, on-download build failed, or a TRT upgrade invalidated the engine) — never sees those skills. For exactly them, after a TRT upgrade you eat the ~1-min lazy build stall on first execution, with no pre-warm. Not a correctness bug, just the slow path permanently for user-configured/legacy skill locations.

Fix: drive the sweep from the same source of truth as the loader instead of hardcoding a subset:

```python
from brain_client.common.script_paths import get_skill_directories
scan_roots = [str(d) for d in get_skill_directories()]
```

(If `innate_training_node` cannot import `brain_client`, read `INNATE_EXTRA_SKILL_DIRS` + the legacy/home dirs the same way — the point is to not hardcode a subset of the recognized roots.)

ros2_ws/src/cloud/innate_training_node/innate_training_node/workers.py
```diff
@@ -256,6 +322,15 @@ def do_download(
         except Exception as e:
             _ros.warning(f"Download OK but activation failed for {sid}/{run_id}: {e}")
 
+        # Pre-build the TensorRT inference engine so the first execution is fast
+        # (the engine is otherwise built lazily on first run, costing ~1 min).
+        # Best-effort: failures here never affect the download/activation result.
+        if result is not None:
+            try:
+                _prebuild_trt_engine(dest_dir, result["checkpoint"])
```
vignesh-anand, line 345: **(Pre-existing, not introduced here — but this PR touches the spot, so flagging) Failed `activate_run` leaves the skill silently unusable, and the new sweep cannot recover it**

To be clear up front: this PR handles its own concern correctly — gating the pre-build on `if result is not None` is exactly right, since a failed activation means there is nothing valid to build. The stuck-state below is older behavior. I am only raising it because the new pre-build/sweep machinery sits right on top of it and could cheaply make it self-healing.

The chain when `activate_run` raises (no `*_step_*.pth` / `*stats*.pt` match, missing run dir, unreadable `metadata.json`):
- files are on disk, but `execution.checkpoint` is never written to `metadata.json` (the metadata write is the last step in `activate_run`);
- pre-build is skipped (correctly);
- `do_download` still logs `Download finished` — so it reports **success**, and the only signal is one `WARNING` line among the INFO logs;
- the skill then **cannot run** (manipulation_server resolves the checkpoint from `execution.checkpoint`), and
- the **startup sweep cannot rescue it** either — `prebuild_all_engines` keys off `execution.checkpoint`, the exact field that was never written, so the skill is skipped;
- and `mark_download_started` is never undone on this path, so there is no in-session auto-retry.

Net: one skill is quietly un-runnable until something external intervenes (app re-registers the dir + node restart, or a manual metadata fix).

Would be nice (follow-up, not blocking this PR):
- call `store.unmark_download(skill_id, run_id)` on the activation-failure path so a retry can fire, and/or
- have the sweep notice "checkpoint files present but no `execution.checkpoint`" and attempt activation, and/or
- surface the activation failure more loudly than a single warning given the download otherwise reports success.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -822,11 +627,15 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim=8, n_action_step
...
+            # n_action_steps is logged below, once we know whether ensembling is active --
+            # it only governs the chunked-queue replan interval and is unused while ensembling.
+            # Build on CPU and adopt the checkpoint tensors directly (assign=True) to avoid
+            # allocating random weights on the GPU and a redundant device transfer. The model
+            # is moved to the device once, after the state dict is loaded.
+            self.current_policy = ACTPolicy(config=policy_config, dataset_stats=dataset_stats)
 
             # Load state dict with strict=False to handle potential mismatches gracefully
-            load_result = self.current_policy.load_state_dict(state_dict, strict=False)
+            load_result = self.current_policy.load_state_dict(state_dict, strict=False, assign=True)
```
vignesh-anand: **Cached-engine path still pays the full eager load — the common case after pre-build does ~3.6s of pure waste**

After pre-build (the intended common case) the engine is already cached, but this load path unconditionally does the entire eager pipeline anyway:
- load the full checkpoint to CPU,
- construct `ACTPolicy` (random-inits resnet18 + the transformer on CPU — discarded),
- `load_state_dict`,
- `.to(self.device)` (the host→device transfer),

…then calls `build_engine(...)`, which on a cache hit returns at its first line without ever touching the policy weights:

```python
engine_path = engine_path_for(...)
if os.path.exists(engine_path):
    return engine_path   # policy never used
```

…and then immediately releases the eager model (the `self.current_policy = None; gc.collect()` block) without ever running it. `TRTACTPolicy` loads the engine directly and needs none of it. So the whole eager-model load is wasted on every cache hit. The PR cut the eager load 12.6s→3.6s, but on the cached path that 3.6s is entirely skippable.

`act_trt._main` already has exactly the right fast-path; the runtime can mirror it:

```python
state_dict = normalize_state_dict(torch.load(checkpoint_path, map_location="cpu", weights_only=True, mmap=True))
chunk_size = infer_chunk_size(state_dict)                 # cheap: one tensor`s shape via mmap
policy_config = create_act_config(action_dim=action_dim, chunk_size=chunk_size, n_action_steps=n_action_steps_override)
engine_path = engine_path_for(checkpoint_path, action_dim, chunk_size, "fp32")
if not os.path.exists(engine_path):
    # cache miss only: build the eager model, load_state_dict(assign=True), .to(device), build_engine(...)
    ...
trt_policy = TRTACTPolicy(engine_path, policy_config, self.device, temporal_ensemble_coeff=ensemble_coeff)
```

Confirmed nothing on the hit path needs the eager model or the loaded stats:
- `TRTACTPolicy(engine_path, config, device, coeff)` takes the **config**, not `dataset_stats`, and holds no torch weights;
- `create_act_config` needs only `action_dim` + `chunk_size`;
- `engine_path_for` → `_dataset_stats_signature` reads `dataset_stats.pt` **from disk** for the hash, so it does not need the `torch.load(stats_path)` object either — the `dataset_stats` load can also be skipped on a hit.

Net: common-case policy load drops from ~3.6s to roughly engine-deserialize + warmup. The eager `ACTPolicy` only needs to exist on a genuine cache miss (to export the engine).

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -773,45 +595,28 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim=8, n_action_step
...
-                    # Remove both prefixes if present (handle nested cases like module._orig_mod.)
-                    cleaned_key = key
-                    if cleaned_key.startswith("module."):
-                        cleaned_key = cleaned_key.replace("module.", "", 1)
-                    if cleaned_key.startswith("_orig_mod."):
-                        cleaned_key = cleaned_key.replace("_orig_mod.", "", 1)
-                    cleaned_state_dict[cleaned_key] = value
-                state_dict = cleaned_state_dict
+            try:
+                state_dict = torch.load(checkpoint_path, map_location="cpu", weights_only=True, mmap=True)
```
vignesh-anand: **Inconsistent `weights_only` handling vs the stats load (nit)**

This checkpoint load is pinned and hardened (`weights_only=True` + a `weights_only=False` fallback), but the `dataset_stats` load a few lines above (line 593) omits `weights_only` entirely, so its behavior rides on the torch-version default (≥2.6 → `True`, <2.6 → `False`). And the *same* `dataset_stats.pt` is loaded with yet a third treatment in the offline builder (`act_trt._main`: explicit `weights_only=False`). Three loads of the same kinds of file, three different policies.

Impact is low and recoverable: `dataset_stats.pt` is normally plain tensors (loads fine either way), and even if the stats load fails → `None`, the normalize/unnormalize buffers are recreated as `inf` placeholders in `create_stats_buffers` and then assign-loaded from the checkpoint `state_dict` (they are persistent `nn.Parameter`s), so normalization is restored from the checkpoint.

The sharp edge is diagnosability: that recovery is silent and conditional on the checkpoint actually containing those buffer keys. If it ever does not, a swallowed stats-load failure resurfaces far away — either as the `is infinity … initialize ACTPolicy with dataset_stats` error during the ONNX export in `build_engine`, or (worse) as `inf` normalization baked into a cached engine → systematically wrong actions, with only a buried warning pointing back to the stats load.

Cheap fix — make the stats load (line 593) mirror this one so all three agree across torch versions:

```python
try:
    dataset_stats = torch.load(stats_path, map_location="cpu", weights_only=True)
except Exception:
    dataset_stats = torch.load(stats_path, map_location="cpu", weights_only=False)
```

Nit-level, not blocking — flagging mainly because the rewrite hardened this load and left its sibling on the version default.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -773,45 +595,28 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim=8, n_action_step
...
-            chunk_size = 50  # default matching most training configs
-            pos_embed_key = "model.decoder_pos_embed.weight"
-            if pos_embed_key in state_dict:
-                chunk_size = state_dict[pos_embed_key].shape[0]
-                self.get_logger().info(f"Inferred chunk_size={chunk_size} from checkpoint")
+            chunk_size = infer_chunk_size(state_dict)
+            self.get_logger().info(f"Using chunk_size={chunk_size} (inferred from checkpoint)")
 
             self.get_logger().info(
                 f"Creating ACTPolicy with action_dim={action_dim}, chunk_size={chunk_size}, "
```
vignesh-anand: **Nit: dead/misleading `action_dim=8` default in the signature (line 573)**

The `action_dim=8` default is never exercised — the sole caller passes `params.action_dim` positionally, which is `LearnedExecCfg.action_dim` (`Field(10)`). The offline builder (`_run_engine_builder`) also defaults to 10. So three places agree on 10 and only this signature says 8.

Harmless today since it is dead, but it is a latent footgun: if a second caller is ever added and relies on the default, it would silently build an 8-dim policy that mismatches the checkpoint `action_head` and fails at the TRT I/O shape check. Suggest `action_dim=10` (match the real default) or drop the default and make it required.

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
```diff
@@ -0,0 +1,375 @@
...
+
+        serialized = builder.build_serialized_network(network, config)
+        if serialized is None:
+            raise RuntimeError("TensorRT engine build returned None")
+        # Write atomically: a process killed mid-write (OOM, disk full, SIGKILL) must not
+        # leave a truncated engine at engine_path, which would poison the cache (the next
+        # load sees the file, skips the rebuild, and fails to deserialize forever).
+        with open(tmp_path, "wb") as f:
+            f.write(serialized)
+        os.replace(tmp_path, engine_path)
```
vignesh-anand, line 187: **Note (not blocking): superseded engines are never cleaned up → slow unbounded disk growth**

The cache key is scoped by TRT version / precision / dims / stats-hash, so a TRT upgrade, a precision change, or a stats regen builds a *new* `.engine` and leaves the old one on disk forever. Nothing ever prunes superseded engines.

Sizing it so it is not hand-wavy: from the PR’s own RAM numbers (~707 MB fp32 vs ~619 MB fp16, ~88 MB delta), the baked inference weights are ~176 MB, so a serialized fp32 engine file is roughly **~200 MB** (fp16 ~100 MB). So each orphan is a few hundred MB, and on a unit with N skills one TRT upgrade orphans N × ~200 MB — a few GB over a unit’s lifetime, stacked on disks already heavy with 300 MB–1 GB checkpoints. On an Orin’s eMMC that is the kind of thing that silently fills a disk months later.

Not crucial for this PR — just flagging. If it is easy, the cheap fix is to prune sibling engines for the same checkpoint right after the successful `os.replace`: glob `f"{checkpoint_path}.*.engine"`, keep the one just built, remove the rest (keeps exactly one engine per checkpoint). Could also be a tiny step in the startup sweep instead. Worth doing only if low-effort.

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
```diff
@@ -0,0 +1,375 @@
...
+
+    device = torch.device("cuda")
+    checkpoint_dir = os.path.dirname(checkpoint_path)
+
+    # The engine path depends on action_dim (CLI arg) and chunk_size, and chunk_size is
+    # only knowable from the checkpoint. The state_dict load below is mmap'd, so reading
+    # one tensor's shape is cheap -- do it first so the fast path can skip the expensive
+    # work (dataset_stats load, model construction, device transfer, ONNX export + build).
+    # Normalize first so wrapped/compiled checkpoints resolve like they do in the server.
+    state_dict = normalize_state_dict(torch.load(checkpoint_path, map_location="cpu", weights_only=True, mmap=True))
```
vignesh-anand: **Minor: `_main` lacks the `weights_only=False` fallback the server has → pre-build silently fails for some checkpoints**

The server load (`manipulation_server.py:604`) tries `weights_only=True` and falls back to `weights_only=False`; this builder load does not. `weights_only=True` rejects a checkpoint whose outer container holds non-basic types (a training-config dataclass/enum, numpy scalar, custom class) — not hypothetical, since `normalize_state_dict` already unwraps wrapped checkpoints and the server team added the fallback for a reason.

For such a checkpoint:
- on-download pre-build and the startup sweep both fail (`subprocess.run(check=True)` → `CalledProcessError`, swallowed as a `pre-build skipped` warning) — and since the engine is never created, the sweep re-attempts and re-fails this build **on every boot**;
- yet the server loads it fine at runtime (its fallback), hits a cache miss, and builds the engine inline — so it works, but you eat the full ~1 min cold build on first run, the exact stall pre-build exists to avoid.

Not a correctness bug (runtime fallback rescues it), but it silently defeats the pre-build benefit for that class of checkpoints and burns a failing subprocess every boot. Trivial fix — mirror the server:

```python
try:
    raw = torch.load(checkpoint_path, map_location="cpu", weights_only=True, mmap=True)
except Exception:
    raw = torch.load(checkpoint_path, map_location="cpu", weights_only=False, mmap=True)
state_dict = normalize_state_dict(raw)
```

ros2_ws/src/brain/manipulation/manipulation/act_config.py
```diff
@@ -0,0 +1,90 @@
...
+    Falls back to ``default`` when the key is absent so the runtime and the offline
+    pre-build behave identically (both build/load an engine for the same chunk_size)
+    instead of one raising and the other silently defaulting.
+    """
+    if _CHUNK_SIZE_KEY in state_dict:
+        return state_dict[_CHUNK_SIZE_KEY].shape[0]
+    return default
+
+
+def create_act_config(action_dim=10, chunk_size=30, n_action_steps=None):
```
vignesh-anand: **Nit: the two `chunk_size` defaults disagree (`create_act_config`=30 vs `infer_chunk_size`=50)**

`create_act_config(..., chunk_size=30, ...)` here defaults to 30, while `infer_chunk_size(state_dict, default=50)` defaults to 50. Harmless in practice — both real callers (server + `_main`) always pass the inferred `chunk_size`, so neither default is exercised — but two functions in the same module having different fallback chunk sizes is a smell that could mislead a future caller that relies on one of them. Suggest aligning them (or dropping the `create_act_config` default so the value is always explicit). Nit only.

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
```diff
@@ -0,0 +1,375 @@
...
+        self.config = config
+        self.device = device
+        action_dim = config.output_shapes["action"][0]
+        image_shape = config.input_shapes["observation.image_camera_1"]
+        state_dim = config.input_shapes["observation.state"][0]
+
+        self._runtime = trt.Runtime(trt.Logger(trt.Logger.WARNING))
+        with open(engine_path, "rb") as f:
+            self._engine = self._runtime.deserialize_cuda_engine(f.read())
+        self._context = self._engine.create_execution_context()
```
vignesh-anand, line 237: **Note: a corrupt/incompatible cached engine gives an opaque error.** `deserialize_cuda_engine` returns `None` (it does not raise) on a bad engine — e.g. one built for a different GPU arch, or a TRT-version mismatch that somehow slipped the filename. The very next line then does `None.create_execution_context()` → a bare `AttributeError: 'NoneType' object has no attribute ...`, which surfaces as "Failed to load policy" with no hint that the fix is to delete the cached `.engine`. The atomic write guards against *truncated* engines but not against a *complete-but-incompatible* one. A one-liner would make it self-explanatory:

```python
self._engine = self._runtime.deserialize_cuda_engine(f.read())
if self._engine is None:
    raise RuntimeError(f"Failed to deserialize TRT engine at {engine_path} (incompatible/corrupt cache — delete it to rebuild)")
```

Minor / robustness only.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -844,24 +653,65 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim=8, n_action_step
...
+            else:
+                self.get_logger().info(
+                    f"Resolved n_action_steps={policy_config.n_action_steps} "
+                    f"(chunked-queue replan interval; chunk_size={chunk_size})"
+                )
+
+            # Imported lazily: tensorrt is hardware-only, so importing at module top would
+            # crash the whole node (incl. poses/replay) on a sim/dev box. Here it raises
+            # only on the learned path, failing this load loudly -- TRT stays required.
+            from manipulation.act_trt import TRTACTPolicy, build_engine  # noqa: PLC0415
```
vignesh-anand: **Question, not a blocker — is the no-fallback acceptable?** We noticed that with the eager `select_action` removed, TensorRT is now a hard requirement with no fallback: if the `tensorrt` import fails *or* the engine build fails at runtime (e.g. a transient CUDA OOM during build), the learned policy can't run at all — the load just fails. Poses/replay are fine (lazy import), but the learned path has no degraded mode.

That seems like a reasonable simplification (one path, not two), and the failure is at least loud. But before we sign off: how reliably has TRT behaved for you on-device in practice — have you seen import or build failures (OOM, ldconfig/driver hiccups, JetPack quirks) that would take a skill out, and if so how do you want that handled (retry? a clear operator-facing error? accept it)? Just want to make sure the hard-requirement is a deliberate call rather than an implicit one.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -844,24 +653,65 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim=8, n_action_step
...
-                self.get_logger().info("Model compilation setup completed")
-                self.get_logger().info("Note: First inference will trigger actual compilation")
-                # Verify compilation was applied
-                if hasattr(self.current_policy, "_orig_mod"):
-                    self.get_logger().info("✓ Compilation verified: model has _orig_mod attribute")
-                else:
-                    self.get_logger().warn("⚠ Compilation may not have been applied (no _orig_mod attribute)")
-            except Exception as e:
-                self.get_logger().warn(f"Model compilation failed: {e}, falling back to uncompiled model")
+                ensemble_coeff = float(os.environ.get("INNATE_TRT_ENSEMBLE_COEFF", "0.01"))
```
vignesh-anand: **Should the ensemble coeff live in `metadata.json` / `LearnedExecCfg` like the other learned params, instead of a global env var?**

Every other learned-behavior tunable is a validated, per-skill field on `LearnedExecCfg` (in the skill's `metadata.json`, checked by `validate_behavior_config`): `action_dim` (`Field(10, ge=1, le=64)`), `n_action_steps` (`ge=1`), `duration` (`gt=0`), `progress_threshold`, the pose times, etc. — all bounded and rejected loudly if malformed.

`INNATE_TRT_ENSEMBLE_COEFF` is the lone exception: a global, undocumented env var with no schema and no bounds, parsed ad-hoc here. It is the one motion-shaping knob that sits outside the validated per-skill surface that governs everything else — and it is arguably the most safety-relevant of them (it directly blends the joint commands).

Proposal: make it a `LearnedExecCfg` field, e.g. `temporal_ensemble_coeff: float = Field(0.01, ge=..., le=...)`. That:
- puts it on the same validated surface as its siblings (consistency),
- makes it **per-skill** (different skills may want different smoothing),
- and closes the NaN gap for free: the existing `isfinite` guard only rejects nan/inf *coeffs*, but a finite-but-large **negative** coeff overflows the float32 ensemble weights (`exp(|coeff|·i)`) to `inf` → `inf/inf` = NaN joint commands (reachable around `coeff ≈ -3` for chunk_size 50). A pydantic `ge`/`le` bound prevents that the same way `action_dim`'s `le=64` does, with no hand-rolled check.

@theo-michel — does the app currently let a user adjust the other `LearnedExecCfg` params (`n_action_steps`, `action_dim`, etc.) per skill? If so, should it expose this one the same way? And if you instead intend the coeff to stay a global dev-only knob (the "set 0 for the chunked path" hint reads that way), can we at least bound its magnitude so a bad value can't produce NaN commands?

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -699,15 +526,10 @@ def _execute_replay_behavior(self, goal_handle, skill_dir, params: ReplayExecCfg
...
-                self.arm_state_pub.publish(arm_msg)
+                self._publish_arm(action[0:6])
 
-                # Extract cmd_vel commands (elements 6-7 = linear.x, angular.z)
-                twist_msg = Twist()
-                twist_msg.linear.x = float(action[6]) / 2.0
-                twist_msg.angular.z = float(action[7]) / 2.0
-                self.cmd_vel_pub.publish(twist_msg)
+                # Extract cmd_vel commands (elements 6-7 = linear.x, angular.z).
+                self._publish_base(float(action[6]) / 2.0, float(action[7]) / 2.0)
```
vignesh-anand: **(Not introduced by this PR — but the refactor surfaced it) the `/2` on the replay base looks wrong: replay plays the base back at half the recorded speed.**

The recorder stores the **raw** `cmd_vel` straight into the action vector, with no scaling at record time (`recorder_node.cpp`):

```cpp
action_data.push_back(latest_cmd_vel_->linear.x);   // index 6
action_data.push_back(latest_cmd_vel_->angular.z);  // index 7
```

So `action[6:8]` in a replay file *is* the exact `cmd_vel` that was commanded during the demo. Replay then does `_publish_base(action[6] / 2.0, action[7] / 2.0)` — i.e. it reproduces the base at **half** the recorded speed. For a path whose whole job is faithful reproduction, that seems off; I would expect replay to send the recorded `cmd_vel` 1:1 (no `/2`).

The learned path also applies `/2`, but that one is more defensible: the policy is trained to predict those same raw `cmd_vel` targets, so its `/2` could be an intentional global base-speed de-rate. Replay has no such justification — the source is the literal recorded command.

To be clear this is **pre-existing** (the old inline replay code had the same `/2`), not something this PR changed — flagging only because pulling both paths through the shared `_publish_base` puts the two `/2`s side by side and makes the replay one look suspicious. Question: is the replay `/2` intentional (a deliberate half-speed playback), or should replay reproduce the recorded base velocity 1:1?

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -914,29 +765,27 @@ def _start_sensor_subscriptions(self):
...
         self.latest_image1 = None
         self.latest_image2 = None
         self.latest_joint_state = None
         self.latest_image1_timestamp = None
         self.latest_image2_timestamp = None
         self.latest_joint_timestamp = None
-        self.get_logger().info("Sensor subscriptions stopped")
 
     def image1_callback(self, msg: Image):
+        if not self.execution_running:
```
vignesh-anand, line 811: **Nit: `joint_state_callback` lacks the idle early-return this and `image2_callback` now have.** Both image callbacks early-return on `if not self.execution_running` while idle, but `joint_state_callback` (line 795) always stores `latest_joint_state`. Functionally harmless (a JointState is tiny and there is no cv_bridge conversion, so the idle cost is negligible), but it is an asymmetry the surrounding docstrings imply is uniform ("the image callbacks early-return while idle"). Either add the same guard there for consistency, or leave it — just flagging the inconsistency.

vignesh-anand (commented):
Left comments

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
theo-michel: Ok that explains the extra time I was seeing thanks!

theo-michel: Addresed

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
theo-michel: Doesn't matter

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
theo-michel: True

theo-michel: Fixed

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
theo-michel: Works well! And it's the same jetson everywehre cleaner to have just one path

ros2_ws/src/cloud/innate_training_node/innate_training_node/workers.py
theo-michel: Fixed

ros2_ws/src/cloud/innate_training_node/innate_training_node/workers.py
theo-michel: https://linear.app/innate/issue/INN-606/failed-activate-run-leaves-skill-silently-unusable-make-downloadsweep

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
theo-michel: fixed

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
theo-michel: fixed

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
theo-michel: fixed

ros2_ws/src/cloud/innate_training_node/innate_training_node/workers.py
theo-michel: Partially fixed

ros2_ws/src/brain/manipulation/manipulation/act_trt.py
theo-michel: fixed

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
theo-michel: fixed

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
theo-michel: Will create a ticket!

theo-michel: Created ticket

karmanyaahm (approved):
lgtm

PR #448, opened 2026-06-17
fix: restore brain client executable bit

axelpey:
## Summary
- restore executable mode on the brain client node entrypoint

## Why
Merged main currently gets through Docker/ROS/Genesis startup, but ROS cannot exec the brain client node because the installed entrypoint is a symlink to a non-executable source file.

## Test plan
- ./innate sim setup
- ./innate sim up --once


axelpey:
@theo-michel can you sanity-check this? Axel asked why this executable-bit fix keeps being necessary.

The confusion/question:

> Why is that necessary? We keep doing it again and again?

My current understanding:

ROS is not running `brain_client_node.py` by explicitly calling `python3`. It asks the OS to execute the installed entrypoint directly. Since this package installs Python node files via `install(PROGRAMS ...)`, those source files need both:

- a Python shebang
- executable mode (`100755`, not `100644`)

Because this workspace uses symlink-style installs, the installed entrypoint is a symlink back to the source file. If the source file loses its executable bit, the installed ROS entrypoint becomes non-executable too and launch fails with `PermissionError`.

This PR is the immediate repair: restore executable mode on `brain_client_node.py`.

Longer-term options to stop rediscovering this:

1. Add a CI or launcher validation that checks every Python file listed in `install(PROGRAMS ...)` has a shebang and executable mode.
2. Prefer moving Python ROS nodes to proper Python `console_scripts` entrypoints where possible, so packaging generates executable wrappers and we do not depend on source file mode.

Do you agree with this diagnosis, and do you think we should follow up with option 1, option 2, or both?

theo-michel:
It's good, created a ticket to change that later option 2 is the best long term

PR #450, opened 2026-06-18
Webapp in innate-os + live training observability, MP4 encoder & console bridge

DavidDobas:
Brings the **operator web cockpit into innate-os** and launches it on boot with the ROS nodes, plus the robot-side infrastructure that powers it. After this, a single innate-os deploy gives you a browser-based cockpit for the robot — teleop, data collection, dataset replay, live training, and a full debugging console — with no separate app to install or start.

Companion PRs in the other repos (mobile app, ACT, innate-cloud) are linked at the bottom.

## The web cockpit

Vendored the webapp into **`webapp/`** (a zero-build static app — `index.html` + ES modules, no build step) and launch it alongside the ROS nodes in a `console-webapp` tmux window: one process serves the **HTTPS app on `:4443`** (+ the `wss` rosbridge proxy and episode media) and a cleartext **`:4080`** media listener for the mobile app. All five pages are new to this repo and share the same modules:

- **Teleop** *(new page)* — live WebRTC video, joystick/keyboard drive, head tilt, robot speech, telemetry, and leader-arm USB follow (WebSerial). On top of the carried-over base, this work adds arm **Reboot** + live **Torque ON/OFF** controls (parity with the mobile app), cuts **drive latency** (~60 Hz trailing-edge `/joystick` publish instead of waiting on the 150 ms heartbeat), and hardens **video reliability** (phase-aware handshake retry, and holding the stream until real frames actually flow instead of flashing black).
- **Collect** *(new page)* — a data-collection station: the teleop cockpit reused with a recording HUD. Record **learned-skill episodes** into a training dataset, or capture a one-shot **recorded movement** — ported from the mobile app's add-skill flow, with the same record gate (leader arm engaged) and fixed-viewpoint head lock during learned recording.
- **Datasets** *(new page)* — browse a skill's episodes and replay them (per-camera **MP4** with native scrubbing + a synced joint graph), with ready vs **preparing** (still-encoding) states, a `N preparing` count, and ‹ / › episode navigation.
- **Training** *(new page)* — start and watch cloud training runs: a **"Running now"** hero card, a determinate **step progress bar + ETA**, **live logs**, the **W&B** link, and a run-detail pane with **hyperparameters**.
- **Debugging** *(new page)* — a live, structured view of the robot's *entire* console (powered by the bridge below): chronological log with severity/source columns, filtering and search, and a "Log Sources" navigator that re-pivots by launch file → process → node → logger.

### Unified logging + a structured console bridge → the Debugging page

Previously each node and library logged in its own format, and much of the robot's real output — library lines, launch-system lines, the rws `asio.system:104` errors — **never reached `/rosout`**, so it was invisible off-box (debugging meant SSHing in to read tmux panes). Now:

- **One log format, pinned once.** `config/dds/setup_dds.zsh` exports `RCUTILS_CONSOLE_OUTPUT_FORMAT='[{severity}] [{time}] [{name}]: {message}'`, so every line is self-describing (severity, timestamp, node/logger name) — the contract the observability tooling relies on. Plus **node-attributed loggers**: dropped detached `get_logger("…")` loggers in `app.cpp` and the camera `main()` fallbacks, and stopped `brain_client` from stripping the node name, so lines attribute to the owning node.
- **`innate_console`** *(new package)* taps every tmux pane's full stdout via `tmux pipe-pane` (seeding from scrollback so quiet panes aren't empty), parses each line **once** into a structured record `{ window, process, node, logger, level, msg, … }`, and republishes over rosbridge. Backfill is buffered **per node** (LRU-capped) so one flooding node can't evict everyone else's history. `/rosout` is an incomplete subset; pane stdout is the complete superset.

### Background MP4 encoding → Datasets & replay

- **`dataset_encoder`** *(new node)* converts freshly-recorded episodes (raw HDF5) to seekable H.264 **MP4** the moment they're saved, reusing the exact idempotent converter cloud-sync runs — so both apps replay episodes as ordinary video and a later sync becomes a pure upload. It runs **continuously** at high `nice` / idle I/O / few threads (no gating — an earlier gated design left episodes stuck un-encoded whenever a live view was open).
- **Encoder output fixes** for mobile playability: 4:2:0 chroma (4:4:4 showed black), ~1 s GOP (so seeks don't decode from frame 0), faststart.
- **Atomic metadata writes** — `dataset_metadata.json` is written via `.tmp` + `rename` in the save/outcome/delete paths, so a power cut mid-write can't truncate it to invalid JSON and break every later reader.
- New `brain_messages` services: `EncodeEpisode` / `DeleteEpisode` / `SetEpisodeOutcome`.

### Live training observability → the Training page

The full vertical slice from GPU to browser: **step progress, ETA, live logs (5 s cadence), the W&B link, and hyperparameters**, plumbed through `training-client`, `innate_training_node`, and `innate_cloud_msgs` (`GetRunLogs` / `CancelRun` / `StartTraining`; new `current_step`/`total_step`/`training_params_json` on `TrainingRunStatus`). Dataset uploads are **content-deduped by MD5** (re-running a skill with unchanged videos only re-uploads changed metadata; a missing `Content-Length` falls back to "not uploaded", the safe side), with an accurate verify-progress bar.

## Related PRs (other repos)
- **Mobile app** — MP4 episode replay (replaces the WebRTC replay path, which couldn't coexist with the browser cockpit and had no scrubbing): innate-inc/innate-controller-app#211
- **ACT** (trainer) — emits the `@@INNATE_PROGRESS step=N total=M` marker that drives live progress: innate-inc/ACT#20, innate-inc/ACT#21
- **innate-cloud** (orchestrator + GPU daemon) — the cloud end of live training observability: innate-inc/innate-cloud#52, innate-inc/innate-cloud#53, innate-inc/innate-cloud#54

## Deploy
- Rebuild the changed packages: `brain_messages`, `innate_cloud_msgs`, `innate_training_node`, `cloud_clients` (training-client), `dataset_encoder`, `innate_console`, `manipulation`, `mars_cam`, `mars_control`, `brain_client`.
- **Restart rosbridge** — the `TrainingRunStatus` and `brain_messages` schemas changed.
- **Relaunch the `ros_nodes` tmux session** — picks up the new launch layout, the `console-webapp` / webapp window, *and* re-sources `setup_dds.zsh` so the unified log format takes effect.
- Open `4443` and `4080` on the LAN firewall. The webapp itself is zero-build (just reload).
- ACT/cloud deploy via their own PRs: the orchestrator ships the GPU daemon, and the next training run picks up the ACT marker (preset pins `ref: main`). All training fields are optional, so un-updated robots/daemons just show the indeterminate bar.


ros2_ws/src/brain/dataset_encoder/dataset_encoder/dataset_encoder_node.py
```diff
@@ -0,0 +1,239 @@
...
+
+    # ---- worker (timer advances one work item per tick) ------------------
+    def _tick(self):
+        if self._gen is None:
+            if not self._queue:
+                return
+            self._current_dir = self._queue.popleft()
+            self._queued.discard(self._current_dir)
+            data_dir = os.path.join(self._current_dir, DATA_SUBDIR)
+            self._gen = convert_episodes_to_h264(
```
vignesh-anand, line 184: **Potential cross-process race on `dataset_metadata.json`.**

This generator is the encoder process; the recorder is a separate process that also writes `dataset_metadata.json`. The two are not synchronized:

- The recorder now writes the file **atomically** (tmp + `rename`, via `write_json_atomic`).
- But the converter's `_update_metadata` (`episode_converter.py`) does a plain `meta_path.write_text(...)` of the metadata snapshot it loaded **at the start of the conversion**.

So if a new episode is saved *during* a long conversion of the same skill, the recorder appends episode N+1 to the file, then the converter finishes and overwrites it with its stale in-memory copy — dropping episode N+1 from the index (the `episode_N+1.h5` on disk survives, but it falls out of `episodes[]`). The `_enqueue` dedup guard prevents double-queuing but does not prevent this read-modify-write overlap; `_scan_skills` won't recover it either, since the episode is no longer in the metadata.

In practice episodes encode quickly so the window is small, but it's a real edge. Options to harden: have `_update_metadata` re-read the on-disk metadata and merge (only set `video_files`/`dataset_type`, never drop entries), and write it atomically like the recorder does.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
```diff
@@ -8,6 +8,25 @@
...
+void write_json_atomic(const std::string& path, const Json& j) {
+    const std::string tmp_path = path + ".tmp";
+    {
+        std::ofstream out(tmp_path);
+        if (!out.is_open()) {
+            throw std::runtime_error("Failed to open " + tmp_path + " for writing");
+        }
+        out << j.dump(4);
+    }  // flush + close before the rename
+    fs::rename(tmp_path, path);
```
vignesh-anand, line 49: **Atomic, but not durable — no `fsync` before/after the rename.**

This gives readers atomicity (a concurrent reader sees old-or-new, never a torn file), which is the goal. But neither the `ofstream` close nor `fs::rename` forces anything to stable storage — at return, the tmp file's contents *and* the rename can both still be only in the page cache.

On a hard power cut (a normal event for a robot), there's a window where the **rename's metadata reaches disk but the tmp file's data blocks do not**, leaving `dataset_metadata.json` present with the right name but garbage/zero content — which breaks every later reader (encoder, uploader, all webapp pages). The very corruption this helper exists to prevent, just moved to the power-cut path.

In practice ext4's default `data=ordered` (+ its rename-flush heuristic) makes this unlikely, so this is a "worth-knowing" note, not a must-fix — but it is an **undocumented dependency on the filesystem/mount options**: on `data=writeback`, a different FS, or flaky eMMC/SD flush ordering, it silently degrades to occasionally bricking the index.

Fully durable version is `fsync(tmp_fd)` before the rename + `fsync(dir_fd)` after (cost: a couple of forced I/Os, tens of ms). Two options: (a) add the fsyncs if power-cut metadata loss is something we cannot tolerate, or (b) at minimum add a one-line comment stating the ext4-ordering dependency. Note the converter's `_update_metadata` (`episode_converter.py`) is more exposed — it does a bare `write_text` with neither atomicity nor fsync.

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
```diff
@@ -374,6 +413,115 @@ def _on_create_run(self, req: CreateRun.Request, res: CreateRun.Response) -> Cre
...
+            if skill is None:
+                res.success, res.message = False, "submit returned no skill"
+                return res
+
+            self._store.put_skill(skill)
+            self._store.register_dir(skill.skill_id, req.skill_dir)
+            self._store.mark_upload_pending(skill.skill_id)
+
+            threading.Thread(
+                target=self._upload_then_run,
```
vignesh-anand: **Daemon upload thread → an in-flight training run can be silently abandoned.**

`_upload_then_run` is two phases: (1) `do_upload` (potentially multi-minute), then (2) `create_run` — and the run is only created *after* the upload returns. The thread is `daemon=True` and nothing joins it (`destroy_node` only stops the poller), so on **any** process exit it is killed abruptly wherever it is: no exception, no `finally`, no cleanup.

"Shutdown" here is routine, not rare — Ctrl-C, crash, reboot, or **relaunching the `ros_nodes` tmux session**, which this PR's own deploy notes instruct. So the likely trigger is: operator clicks Train → upload is still running → box gets redeployed/rebooted → thread dies.

What's lost vs. recovered:
- **Upload (phase 1): recovered.** `do_upload` is idempotent + MD5/HEAD-deduped, so a re-run finishes only the remainder. The cloud dataset self-heals.
- **Run (phase 2): silently abandoned.** If the thread dies before `create_run`, no run is ever created. The caller already got `success=True` ("run will start after upload"), and the **intent is not persisted anywhere** — it lived only in the thread's captured `training_params` + in-memory store (wiped on restart). `_register_local_skill_dirs` re-registers dirs on startup but has no "resume pending start_training" logic, so nothing retries. Net: the skill re-uploads fine, **no run appears, no error shown** — only (maybe) a log line, which an abrupt kill usually preempts.

This is the same silent-failure class as the optimistic `success=True` return: "I clicked Train and nothing happened, no idea why." The data is safe; the *action the user requested* vanishes with no signal.

Hardening (most→least robust):
1. **Persist the intent** (skill_id + params marker on disk); on startup, resume `_upload_then_run` for any whose run was never created. Survives kill-9 / power loss — best fit for a robot.
2. Chain upload→create-run on the **cloud side** so a robot restart can't lose it.
3. Track the thread and `join(timeout=…)` in `destroy_node` — cheap, but only covers graceful shutdown, not crashes/reboots.

(The `_on_submit` path at line ~368 uses the same detached-upload pattern, but it doesn't chain a `create_run`, so only the recoverable upload is at risk there — not a dropped follow-up action.)

ros2_ws/src/brain/dataset_encoder/dataset_encoder/dataset_encoder_node.py
vignesh-anand: **Lost-update race: `_update_metadata` overwrites concurrent metadata edits with a stale snapshot.** (Companion to the note on `dataset_encoder_node.py:172`; this is where the fix goes.)

`convert_episodes_to_h264` snapshots `meta = json.loads(meta_path.read_text())` at the *start* and never re-reads. The `write_text` a few lines below then overwrites the whole file from that stale `meta` at the *end* — seconds-to-minutes later. The recorder is a separate process that also writes this file (episode save, `set_episode_outcome`, `delete_episode`). Its writes are atomic, but **atomicity prevents torn reads, not lost updates** — the converter already holds a complete-but-stale copy, so its overwrite silently reverts anything written during the encode.

Timeline:
```
T0    converter reads meta = [0..N]
      ... long encode ...
T1    recorder saves episode N+1 (atomic write) -> disk = [0..N+1]
T_end converter writes its stale meta           -> disk = [0..N]   (N+1 dropped)
```

Conversions now run **continuously** (gating removed) and recording episodes back-to-back into the same skill is the normal collection workflow, so this interleaving is ordinary, not exotic.

Recovery via `_scan_skills`/`_needs_conversion` is unreliable:
- **Dropped episode:** after the clobber the index is `[0..N]`, all with `video_files`, `dataset_type=h264` -> `_needs_conversion` returns False, so the scan will not re-add N+1. `episode_N+1.h5` is orphaned on disk. (It may ping-pong back if the recorder writes again, since its in-memory copy still has N+1 — but you cannot rely on that.)
- **Dropped `outcome`:** `_needs_conversion` never inspects `outcome`, so a label set during a conversion is **permanently, silently lost** with no path to recovery.

Fix (smallest first):
1. **Re-read the current on-disk metadata right before writing** here, and merge only the fields the converter owns — set `dataset_type` and per-episode `video_files` keyed by `episode_id`/`file_name`, never dropping/reverting entries it did not author — then write **atomically** (tmp + `os.replace`, matching the recorder's `write_json_atomic`).
2. Or make metadata single-writer (route encoder updates through a recorder service).
3. Or drop `video_files` from the JSON and have clients infer "has video / preparing" from MP4 presence on disk — removes the second writer.

(1) closes both the dropped-episode and dropped-outcome cases.

ros2_ws/src/brain/manipulation/src/recorder_node.cpp
```diff
@@ -749,6 +757,56 @@ void RecorderNode::handle_get_task_metadata(
...
+    }
+}
+
+void RecorderNode::handle_delete_episode(const std::shared_ptr<brain_messages::srv::DeleteEpisode::Request> request,
+                                         std::shared_ptr<brain_messages::srv::DeleteEpisode::Response> response) {
+    RCLCPP_INFO(this->get_logger(), "Delete episode %d in %s", request->episode_id, request->task_directory.c_str());
+
+    // Guard: don't delete out from under an in-flight episode of this task.
+    // (TASK_ACTIVE between episodes is fine; only refuse mid-episode.)
+    const bool mid_episode = state_ == State::EPISODE_ACTIVE || state_ == State::EPISODE_STOPPED;
```
vignesh-anand, line 793: **This guard covers recording, but not encoding — delete-during-encode is a genuine corruption bug.**

This refuses a delete only while the skill is *recording* (`EPISODE_ACTIVE`/`EPISODE_STOPPED`). It says nothing about whether the `dataset_encoder` is *converting* the same skill. The two run in separate processes with no coordination (the encoder has no delete/cancel awareness; `delete_episode` has no encoder awareness), so a delete issued while a conversion of that skill is in flight sails right through and produces two independent bad effects.

The Datasets page surfaces "preparing" (still-encoding) episodes specifically, and encoding now runs continuously (gating removed), so deleting an episode while it — or a sibling in the same skill — is encoding is a normal interaction, not an edge.

**Effect A — the deleted entry gets resurrected** (a special case of the lost-update race noted on `episode_converter.py` / `dataset_encoder_node.py:172`):
1. The converter snapshotted `meta` into memory at conversion start; that snapshot still contains episode 5.
2. `delete_episode` erases episode 5 from the file (atomic write) and unlinks its files.
3. At the end of the run, `_update_metadata` does a full-file `write_text` from the **stale snapshot** — writing episode 5's entry back. `video_files` is empty (files were deleted) so it isn't set, but the **entry returns**: a ghost episode 5 with no `.h5` and no MP4s. The delete is silently undone in the index.

**Effect B — unlink-while-open file chaos** (timing-dependent): the converter moved `episode_5.h5` to `raw_data/` and gst/h5py may have it open at delete time. On Linux the inode stays alive for the open fd, so:
- the in-progress camera encode finishes and writes `data/episode_5_cameraX.mp4` **after** the delete removed it → orphaned MP4 with no metadata, no raw source;
- the remaining cameras + the `strip` step re-open `raw_data/episode_5.h5` **by path** (now gone) → raises → the converter aborts the **whole skill's** conversion partway, leaving episodes 6..N un-encoded until the next 60s scan.

End states (some mix of): a ghost entry that won't play, orphaned `episode_5_*.mp4` on disk, and an aborted skill conversion — for a destructive action the **user explicitly requested**. Silent partial-undo of a user-requested delete is the worst-surprise category.

**Fix — needs real cross-process coordination** (metadata handling alone can't fix Effect B):
1. **Make this guard also refuse/defer while the encoder is converting that skill** — e.g. the encoder owns a "converting `<dir>`" state the recorder checks, mirroring the existing mid-recording guard. (Coordination belongs here.)
2. Encoder honors a cancel/skip: check a flag before each work item, and treat a vanished source as "skip this episode," not "abort the whole skill."
3. The re-read+merge fix from the lost-update thread independently kills Effect A (the converter would never write back an entry the on-disk file no longer has).

(1) + (3) together close it: (3) stops the resurrection, (1) stops the unlink-while-open file mess.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
```diff
@@ -258,4 +279,122 @@ std::tuple<bool, std::string, std::string> TaskManager::get_task_metadata_by_dir
...
+            found = true;
+            break;
+        }
+    }
+    if (!found) {
+        return {false, "Episode " + std::to_string(episode_id) + " not found"};
+    }
+
+    try {
+        write_json_atomic(metadata_path, meta);
```
vignesh-anand, line 443: **This atomic write is silently revertible by a concurrent conversion — outcomes set during an encode are lost permanently.**

`set_episode_outcome` correctly writes the metadata atomically, but atomicity only prevents torn reads — it does nothing against the cross-process lost-update race with the `dataset_encoder` (see the thread on `dataset_encoder_node.py:172` / `episode_converter.py`). The converter snapshots `meta` into memory at conversion start and overwrites the whole file from that stale snapshot at the end (`_update_metadata` → bare `write_text`). So:

```
T0    converter reads meta (episode 5 has no/old outcome)
      ... long encode ...
T1    you set episode 5 outcome = "success"  (atomic write)  -> disk has it
T_end converter writes its stale meta                        -> outcome reverted
```

This is worse than the dropped-episode case because it is **unrecoverable**: `_needs_conversion` only checks `dataset_type` and `video_files`, never `outcome`. After the revert the skill looks fully encoded, so the 60s scan never re-touches it, and nothing restores the label. A curation label set while that skill is converting is **permanently, silently lost** — and labels can change which episodes get uploaded for training, so a "success" silently reverting to unlabeled has real downstream effect.

Trigger is ordinary: encoding runs continuously (gating removed), and labeling episodes on the Datasets page while a sibling encodes is normal curation.

Fixed by the same **re-read + merge + atomic write** in `_update_metadata` (merge only the converter-owned fields — `dataset_type`, per-episode `video_files` — and never overwrite `outcome` or other entries it did not author). No change needed here; flagging so this method is not assumed safe on its own.

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
vignesh-anand: **No dedup — near-simultaneous `start_training` calls create duplicate (paid) GPU runs.**

`_on_start_training` calls `mark_upload_pending` but never *checks* any in-flight state before spawning. There's no registry keyed by `skill_id`/`skill_dir`, so every call unconditionally spawns its own `_upload_then_run` thread, and `create_run` is **not idempotent** — each call mints a new `run_id`.

Two near-simultaneous calls are ordinary, and primed by the synchronous-`submit` issue (this handler drains the `submit` generator on the single-threaded executor before returning, so the service feels slow):
- operator double-clicks / re-clicks Train because it seems unresponsive;
- `rosClient` times out the service at 10s and retries, but the server can't cancel an in-flight ROS call, so the original completes too;
- webapp + mobile both trigger for the same skill.

Result (no data corruption — `JobStore` is locked, so this is wasted *work*, not memory unsafety):
- two concurrent `do_upload`s of the same dataset — mostly wasted bandwidth (content-deduped, atomic per-object PUTs), but both write transfer progress so the published progress bar jitters;
- **two `create_run` calls → two GPU training runs for the same skill** — double GPU cost, two dashboard entries, both training on identical data, and the user has no idea they started two.

Severity is medium (it spends money rather than corrupting data), but the triggers are likely and it compounds with the optimistic `success=True` / silently-abandoned-run behavior in `_upload_then_run` (this same service is both too eager to start duplicate work and unable to report when promised work never happened).

Fix:
1. **Guard on in-flight state** before spawning — track active `skill_id`/`skill_dir` (or actually *check* the pending flag); if one's in flight, no-op and return "already in progress."
2. **Make `create_run` conditional** — skip if a run for this skill was just created / is queued/running, or pass an idempotency key the orchestrator dedups on.
3. Fixing the synchronous-`submit` blocking removes the re-click/timeout-retry pressure that triggers this.

(1) is the smallest fix and closes the duplicate-run path. (`_on_submit` shares the no-dedup pattern but doesn't chain `create_run`, so its only cost is redundant uploads.)

ros2_ws/src/brain/dataset_encoder/dataset_encoder/dataset_encoder_node.py
```diff
@@ -0,0 +1,239 @@
...
+                Path(data_dir),
+                nice_level=self.nice_level,
+                threads=self.encode_threads,
+                idle_io=self.idle_io,
+            )
+            self._cur_index = self._cur_total = 0
+            self._cur_file = ""
+
+        try:
+            update = next(self._gen)
```
vignesh-anand, line 194: **Single-threaded executor: this blocks the whole node for the full encode, despite the "stays responsive" framing.**

`rclpy.spin()` is a `SingleThreadedExecutor` — one thread runs one callback at a time, to completion, with no preemption. `_tick` is a timer callback on that thread, and this `next(self._gen)` runs **one full camera encode**: `_encode_camera_to_mp4` opens the h5, spawns gst, writes every frame to its stdin, then `proc.wait()`s. gst is out-of-process, but the Python thread is fully blocked for seconds (more for long episodes).

While that one `next()` runs, nothing else on the node can:
- **`encode_status` freezes** — it's published from inside `_tick` after `next()` returns; there's no separate status timer, so the webapp progress bar sits frozen mid-item;
- **`EncodeEpisode` can't be served** — `_on_encode_request` is on the same thread, so "Prepare video" waits and can hit the rosbridge service timeout → looks failed while encoding is actively happening;
- **inbound `"saved"` buffers** against the depth-10 sub queue → overflow drops (recovered by the 60s scan, but the "encode immediately on save" path silently misses);
- **`_scan_skills` is delayed**.

The docstring's "one work item per tick … stays responsive … no locking" is half-right: *no locking* is real (single thread = no races), but `tick_period_sec` only spaces the gap *between* `_tick` calls — it does nothing about the *duration* of one. A 4s encode = 4s unresponsive regardless of the 0.5s tick.

Fix: move the blocking encode off the executor thread —
1. run `convert_episodes_to_h264` in a dedicated worker thread driven by the queue, and let the timer just poll/publish its status (callbacks stay tiny); or
2. `MultiThreadedExecutor` with the encode timer in its own `MutuallyExclusiveCallbackGroup` and service/scan/sub in a `ReentrantCallbackGroup` (reintroduces the light locking the docstring was avoiding, so (1) is usually cleaner).

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
```diff
@@ -374,6 +413,115 @@ def _on_create_run(self, req: CreateRun.Request, res: CreateRun.Response) -> Cre
...
+        if params_err:
+            res.success, res.message = False, params_err
+            return res
+
+        try:
+            gen = self._mgr.submit(req.skill_dir)
+            skill: SkillInfo | None = None
+            try:
+                while True:
+                    next(gen)
```
vignesh-anand, line 486: **Synchronous `submit` drain blocks the executor — dashboard + all services stall during it.**

This `while True: next(gen)` drains the `submit` generator **on the service-callback thread**, which is the node's single `rclpy.spin()` executor thread. `submit` talks to the cloud orchestrator (network I/O of unknown latency), and nothing else on the node can run until it returns:
- the **`_publish` timer** (which feeds the webapp training dashboard) can't fire — the dashboard stalls;
- **every other service** — `cancel_run`, `get_run_logs`, `get_training_status`, `create_run`, `download_results` — is blocked and can time out.

(The `Poller` runs on its own thread, so cloud polling continues — but the rclpy publish/services are stuck.) The *upload* is correctly offloaded to a thread just below; the `submit` is not. This is also what primes the duplicate-run problem (separate comment): a slow `submit` makes the call look hung, so the operator re-clicks / the client retries, and the server can't cancel the in-flight call.

Fix: move the whole `submit → upload → create_run` chain into the background thread. Validate params synchronously (already done above), then return immediately; the handler should never block on network.

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
```diff
@@ -374,6 +413,115 @@ def _on_create_run(self, req: CreateRun.Request, res: CreateRun.Response) -> Cre
...
+            self.get_logger().error(f"start_training: upload failed for {skill_id[:8]} — run not created")
+            return
+        try:
+            data = self._mgr.client.create_run(skill_id, training_params=training_params)
+            rid = int(data["run_id"])
+            self._store.put_job(self._mgr.run_status(skill_id, rid))
+            self._store.register_dir(skill_id, skill_dir)
+            self.get_logger().info(f"start_training: run {skill_id[:8]}/{rid} created after upload")
+        except Exception as e:
+            self.get_logger().error(f"start_training: create_run failed for {skill_id[:8]}: {e}")
```
vignesh-anand: **`create_run` failure here is silent — log-only, no status pushed back. The upload path already does better; make this match it.**

`start_training` returns `success=True` the instant the skill is submitted, then does the real work in this detached thread. The two phases report failure asymmetrically:

- **Upload failure is surfaced.** `do_upload`'s `except` writes a `STAGE_ERROR` transfer into the `JobStore`, which the periodic `_publish` ships in `TrainingJobList` — so the webapp can show "Upload failed: …".
- **`create_run` failure (this `except`) only logs.** It writes nothing to the store — no error transfer, no failed-run placeholder. So when the upload succeeds but `create_run` throws (orchestrator down, quota, params that passed local validation, a network blip), the dashboard shows **upload DONE and simply no run**, and the user already got `success=True` ("run will start after upload"). No signal except a robot log line they'll never see.

This is the "I clicked Train and nothing happened, no idea why" failure: the *action the user requested* (a run) silently never happens, with no path to discover it. (Distinct from the daemon-thread shutdown case noted at line 449 — that's the thread dying mid-flight; this is a clean in-process failure that just isn't reported.)

The asymmetry is the tell — `do_upload` already established the pattern; this branch just doesn't follow it.

Fix: in this `except`, write a failure the dashboard renders — mirror `do_upload` (a `STAGE_ERROR` transfer for this skill), or put a placeholder `TrainingRunStatus` in an error state with the message. Cleaner still: a dedicated "start_training outcome" status so the whole submit→upload→run attempt has one legible state instead of the user inferring success from a run appearing.

(Worth confirming on the webapp side that an upload `STAGE_ERROR` is actually rendered as "training didn't start," not just a transfer widget turning red — otherwise even the surfaced case is illegible.)

webapp/js/webrtcSession.js
```diff
@@ -0,0 +1,399 @@
...
+    this.#watchdog = setTimeout(() => {
+      this.#watchdog = null;
+      this.#handshakeAttempts += 1;
+      if (this.#handshakeAttempts < MAX_HANDSHAKE_ATTEMPTS) {
+        console.warn(`[webrtc] no media yet (attempt ${this.#handshakeAttempts}), rebuilding`);
+        this.#handshake();
+      } else {
+        console.error("[webrtc] no media after repeated handshakes");
+        this.#closePc();
+        this.#patch({ status: "error" });
```
vignesh-anand, line 387: **No auto-recovery from the latched `error` state when rosbridge stays connected.** (Low/medium — video-only, visible, one-click manual recovery exists. Flagging as a robustness/test-debt item, not a blocker.)

After `MAX_HANDSHAKE_ATTEMPTS` (6) failed handshakes, this latches `status:"error"` and `#closePc()` clears both the watchdog and the ICE-degrade timer — so the session goes fully idle and nothing re-arms. `#started` stays `true`.

The three paths that reset `#handshakeAttempts` and get out are:
1. manual `start()` (the "Retry video" button in `videoStage`),
2. `onStateChange` → `"connected"` — but only on an actual *transition*, i.e. rosbridge had to drop and come back,
3. `goLive()` (media flows), unreachable without a handshake.

So when the failure is purely **media/ICE/robot-pipeline while rosbridge is healthy**, none of the automatic paths fire — it sits in `error` indefinitely until a human clicks Retry. Fine for an attended operator; for an unattended/wall-mounted cockpit it's a black screen that won't self-heal even after the network/robot recovers.

Subtle dead-end worth noting: `setAudio` triggers a rebuild via `#handshake()`, but `#handshake()` does **not** reset `#handshakeAttempts`. From `error` (attempts already 6) that gives exactly one more attempt before re-latching — i.e. some rebuild triggers reset the counter, `#handshake()` itself doesn't. That asymmetry is invisible by reading and is exactly what a test would pin.

Time-to-error varies: ~30s when offers never arrive (5s `OFFER_TIMEOUT` each), up to ~2min when offers arrive but media stalls (20s `MEDIA_TIMEOUT` each).

Suggestions (it's the most complex async module here, with no tests):
1. A slow background retry even in `error` (e.g. re-handshake every 30–60s) so an unattended cockpit self-heals once conditions recover.
2. Or reset `#handshakeAttempts` on a *successful offer* (proof the path is partly alive) so transient media stalls don't march to the hard cap.
3. Unit tests against a fake `ros` for dropped-offer / offer-but-no-media / ICE-flap, asserting the attempt-increment and reset behavior.

(The design comment intends "surface an error and wait for manual retry" — so this is deliberate; the note is that it assumes an *attended* operator.)

webapp/js/rosClient.js
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    if (typeof data.topic === "string") {
+      const sub = this.#subs.get(data.topic);
+      if (!sub) return;
+      sub.retryCount = 0;
+      // rws payload quirk: the message normally rides in `frame.msg`, but
+      // some paths deliver fields on the frame itself — pass the frame
+      // through so `payload.data` keeps working either way.
+      const payload = data.msg ?? data;
```
vignesh-anand, line 292: **Message-envelope shape is normalized here, but consumers still re-handle it — fold the fallback into one place.** (Low — works today; consistency/footgun.)

This passes `data.msg ?? data` to handlers, but it doesn't fully normalize the rws frame quirk, so downstream consumers re-implement their own fallback. e.g. `webrtcSession.js`:

```js
const sdp = payload?.data ?? payload?.msg?.data;   // line 250
const raw = payload?.data ?? payload?.msg?.data;   // line 292
```

and similar dual-pathing in `recordPanel.js` / others. Every consumer that reads a `StringMsg`-style payload re-derives "is the field on `payload` or on `payload.msg`?", which means:
- the rws-frame-shape knowledge is duplicated across modules instead of owned by `rosClient`;
- a new consumer that forgets the `?? payload?.msg?.data` half silently breaks on whichever path it didn't handle;
- if the rws envelope ever changes, it's an N-site fix instead of one.

Suggestion: normalize once here so handlers always receive a single, predictable payload shape (e.g. resolve to the inner message consistently), and drop the per-consumer `?? payload?.msg?.data` fallbacks. Not urgent, but it removes a quiet footgun and centralizes the one piece of rws-specific knowledge.

webapp/js/teleop/armPanel.js
```diff
@@ -0,0 +1,391 @@
...
+  }
+
+  // ---- wiring -----------------------------------------------------------
+
+  /** @param {SerialPort} port */
+  async function openPort(port) {
+    connectBtn.disabled = true;
+    connectBtn.textContent = "Opening…";
+    try {
+      await leader.open(port);
```
vignesh-anand, line 170: **`openPort` multi-trigger race — the `!leader.state.connected` guard is TOCTOU.** (Low — self-converges; brief port thrash.)

`openPort` → `leader.open()` can be invoked from three places: the Connect button (no guard), the `serial 'connect'` event, and `getPorts().then()` on mount. The latter two guard on `!leader.state.connected` — but `connected` only flips `true` *inside* `open()`, after it resolves (`#patch({connected:true})`). So two triggers firing close together both read `connected === false`, both pass the guard, and both call `open()`; the window spans `await port.open()` + stream setup.

How it fires:
- **Cold load:** if the arm is already plugged in, `getPorts()` returns it *and* (depending on timing) the `connect` event also fires — both see `!connected`, both `openPort`.
- **Button vs auto:** the button disables itself only *inside* `openPort`, i.e. after `requestPort()`'s chooser resolves; during the chooser an auto-trigger can start an open, then the chosen port starts a second.

`DynamixelLeader.open()` does `await this.close()` first, so on the same instance the second open's `close()` tears down the streams the first just set up (cancels its reader, releases its writer, closes the port) before re-opening; on the same `SerialPort` you can also get an `InvalidStateError` ("already open") surfacing through `openPort`'s catch as a spurious error flash. It converges to one open port, so impact is brief thrash — hence low.

Fix: serialize opens in `DynamixelLeader` with an `#opening` flag (or chain a promise) so concurrent `open()` calls coalesce/queue, instead of relying on a flag that only becomes true after the operation finishes.

webapp/js/rosClient.js
```diff
@@ -0,0 +1,410 @@
...
+          handler(payload);
+        } catch (err) {
+          console.error(`[ros] handler for ${data.topic} threw:`, err);
+        }
+      }
+      return;
+    }
+
+    if (data.op === "status" && data.level === "error") {
+      const match = typeof data.msg === "string" && data.msg.match(/^Topic (.+?) not found/);
```
vignesh-anand, line 304: **Subscription auto-retry is coupled to a diagnostic error string — silently stops working if the wording drifts.** (Low, but cold-boot-load-bearing.)

This keys the "subscribed before the publisher existed" retry off the rws error text matching exactly `^Topic (.+?) not found` (right prefix, capitalization, unquoted name as group 1). That retry is what makes a subscription to a not-yet-started node eventually deliver instead of silently never arriving.

If the upstream rosbridge/rws ever rephrases the message (`"topic X does not exist"`, a different prefix, a quoted name, …), the regex misses → the code falls into the `else` branch → `console.error` only, **no retry scheduled** → that subscription never retries and its data never arrives, with no user-visible signal (just a console line). A human-readable diagnostic isn't a stable API, so this can break on a rosbridge upgrade with zero changes here.

Why it matters beyond cosmetic: this PR launches the webapp on boot *alongside* the ROS nodes (the `console-webapp` tmux window), so a cold boot is exactly the case where the webapp races node startup and this retry path is load-bearing for panels populating. If the wording drifts, the failure is "after a fresh boot some panels stay empty until you manually reload" — intermittent, timing-dependent, and silent.

Fix options:
1. Match loosely — trigger retry on any `status`-error whose text *contains* a topic name currently in `#subs`, rather than an exact regex.
2. Add a publisher-independent safety net: for any subscription that has received zero messages within a window, re-send `subscribe` with backoff regardless of error text. Robust to both wording changes and dropped status frames.

webapp/js/datasets/episodePlayer.js
```diff
@@ -0,0 +1,505 @@
...
+    ["success", "Successful"],
+    ["failure", "Unsuccessful"],
+  ]) {
+    const o = document.createElement("option");
+    o.value = val;
+    o.textContent = text;
+    if ((val === "failure") === (episode.outcome === "failure")) o.selected = true;
+    label.appendChild(o);
+  }
+  label.addEventListener("change", () => {
```
vignesh-anand, line 91: **Outcome label is optimistic with no revert — UI can show a label the robot never stored.** (Medium — silent state divergence.)

The `change` handler mutates `episode.outcome` and recolors the select *before* the service call, and the call's only failure handling is `.catch((err) => console.error(...))`:

```js
episode.outcome = outcome;
applyLabelClass(outcome === "failure");
ros.callService(SET_EPISODE_OUTCOME_SERVICE, { ... })
  .catch((err) => console.error("[datasets] set outcome failed:", err));
```

If `SetEpisodeOutcome` fails (disconnect, timeout, or — see the converter lost-update race on `task_manager.cpp:320` / `episode_converter.py` — gets silently reverted server-side), the UI keeps showing the new label with no revert and no user-visible signal. The operator believes a curation label is set that the robot doesn't have — which matters because the label gates which episodes get uploaded for training.

Fix: follow the torque-toggle pattern in `armPanel.js` (`buildArmServices`) — optimistic update, then **revert + flash an error on failure**, and ideally re-confirm from the authoritative episode list rather than trusting local state. (Note this is the *display* half; the backend lost-update race is the other half — both need fixing for the label to be trustworthy.)

webapp/js/training/main.js
```diff
@@ -0,0 +1,219 @@
...
+    // "Starting run…" hint. The robot owns the actual submit→upload→create_run
+    // (start_training service), so this is cosmetic and the page can close.
+    // Value = run count at request time, used to clear the hint once it lands.
+    /** @type {Map<string, number>} */
+    pending: new Map(),
+    /** @param {string} dir @param {TrainingParams} runParams @param {number} _episodeCount */
+    startTraining(dir, runParams, _episodeCount) {
+      const st = this.statusFor(dir);
+      this.pending.set(dir, st?.runs?.length ?? 0);
+      ros.callService(START_TRAINING_SERVICE, { skill_dir: dir, run_params: runParams }).catch((err) => {
```
vignesh-anand, line 78: **Training-start failure has no UI surface — the "Starting run…" hint clears the moment upload begins.** (Medium — pairs with the backend `create_run` silent-failure on `node.py:478`.)

`startTraining` sets a `pending` "Starting run…" hint, then fires `START_TRAINING_SERVICE` and `.catch`es **only the synchronous service rejection** (console + clear hint). `_prunePending` then clears the hint as soon as `has_active_transfer` becomes true — i.e. the instant the upload *starts*:

```js
if (st && (st.has_active_transfer || (st.runs?.length ?? 0) > base)) this.pending.delete(dir);
```

So once upload begins, the UI has no "training is starting / failed to start" state at all. Combined with the backend behavior:
- upload succeeds but **`create_run` fails** server-side → hint already gone, no new run appears, dashboard shows nothing wrong → **completely silent** to the operator;
- an upload *error* renders only as a generic `transferBar` stage label, not a clear "training didn't start."

The frontend has no place for a start failure to land. The backend fix (surface the `create_run` failure as a status — see `node.py:478`) is necessary, and the dashboard needs to actually render that failed state (a failed-run/error row, or keep the pending hint until a run *or* an error is observed — not clear it on upload-start).

Fix here: don't clear `pending` on `has_active_transfer` alone; clear it only when a new run appears *or* a failure status is observed, and render the latter.

ros2_ws/src/brain/dataset_encoder/dataset_encoder/dataset_encoder_node.py
DavidDobas: Fixed in 86d155c4 — `_update_metadata` re-reads `dataset_metadata.json` fresh under a shared `flock` and merges only converter-owned fields (`dataset_type`, per-episode `video_files`), written atomically. A save/delete/outcome edit during the encode is no longer clobbered by the stale snapshot.

ros2_ws/src/brain/manipulation/src/recorder_node.cpp
DavidDobas: Fixed in 86d155c4 — the recorder delete and the converter now take the same `flock`; the converter re-reads fresh (no resurrection), MP4s stage to `.part` + atomic rename, and a source deleted mid-pass is skipped instead of aborting the whole skill.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
DavidDobas: Fixed in 86d155c4 — same `flock` + fresh re-read/merge in the converter, so an outcome set during a conversion is no longer reverted.

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
DavidDobas: Fixed in cb29f6f5 — `start_training` takes an in-flight claim per `skill_dir` (released when the background upload+run finishes) and rejects a duplicate with "already starting".

webapp/js/teleop/armPanel.js
DavidDobas: Fixed in af87fd00 — `DynamixelLeader.open()` now coalesces concurrent opens with an `#opening` flag (first caller wins, the rest no-op), so the cold-boot getPorts + 'connect' double-trigger can't tear down each other's streams.

webapp/js/datasets/episodePlayer.js
DavidDobas: Fixed in be86b038 — the label reverts (outcome + value + class) on a failed `set_episode_outcome` instead of showing a label the robot never stored.

ros2_ws/src/brain/dataset_encoder/dataset_encoder/dataset_encoder_node.py
```diff
@@ -0,0 +1,239 @@
...
+        # Distinguish "no metadata to check yet" from "already encoded" — both
+        # make _needs_conversion return False, but only the latter is a success.
+        if not os.path.isfile(os.path.join(task_dir, DATA_SUBDIR, DATASET_METADATA)):
+            response.success = False
+            response.message = f"No {DATASET_METADATA} yet under {task_dir}/{DATA_SUBDIR}"
+            return response
+        if not self._needs_conversion(task_dir):
+            response.success = True
+            response.message = "Already encoded; nothing to do"
+            return response
```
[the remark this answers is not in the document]
DavidDobas: Fixed in 86d155c4 — `_on_encode_request` now validates the JSON explicitly (`json.load`) before `_needs_conversion`, so a present-but-corrupt `dataset_metadata.json` returns a real error instead of "already encoded; nothing to do".

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
```diff
@@ -374,6 +413,115 @@ def _on_create_run(self, req: CreateRun.Request, res: CreateRun.Response) -> Cre
...
+            threading.Thread(
+                target=self._upload_then_run,
+                args=(skill.skill_id, req.skill_dir, training_params),
+                daemon=True,
+                name=f"train-{skill.skill_id[:8]}",
+            ).start()
+
+            res.success = True
+            res.message = f"Skill {skill.skill_id} submitted — run will start after upload"
+            self.get_logger().info(f"start_training: {skill.skill_id[:8]} submitted — syncing then creating a run")
```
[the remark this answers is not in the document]
DavidDobas: Fixed in cb29f6f5 — `start_training` takes an in-flight claim per `skill_dir` (released when the upload+`create_run` finishes) and rejects a duplicate with "already starting". The claim spans only the upload+create_run window, not the GPU run, so sequential runs with different params are unaffected.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
DavidDobas: Leaving as-is for now. As you note, ext4 ordered-mode makes the power-cut window unlikely, and it's non-blocking — we'll add `fsync(tmp)`+`fsync(dir)` if we decide the index's power-cut durability is worth the forced I/O.

webapp/js/webrtcSession.js
DavidDobas: Intentional for now: surface the error and wait for a manual Retry, which fits our attended-operator use. We'll add a slow background retry if/when we run an unattended cockpit — flagging rather than changing the tuned retry semantics now.

webapp/js/rosClient.js
DavidDobas: Leaving the per-consumer fallback. It's defensive and correct today, and centralizing it touches every consumer for no behavior change — low-value churn mid-PR. Noted for a later cleanup.

webapp/js/rosClient.js
DavidDobas: Low priority — only bites if rosbridge rewords its status string. Noting it; a loose match (any error referencing a subscribed topic) is a reasonable future hardening, but it's guarding a hypothetical, not a current bug.

ros2_ws/src/brain/dataset_encoder/dataset_encoder/dataset_encoder_node.py
DavidDobas: Should be fine to leave as-is: the encode runs in a separate gst process, so it never blocks other nodes — only this node's own status publish / EncodeEpisode service, and only for the few seconds of each camera encode. Both self-recover (auto-encode-on-save + the 60s rescan), and progress still steps per camera. Low priority; we'll move the encode to a worker thread if the responsiveness ever bites. (Fair point that the "stays responsive" wording oversells it.)

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
DavidDobas: Fixed in cb29f6f5 — the whole submit → upload → create_run chain now runs on a background thread, so start_training returns immediately and the publish timer + other services stay live. (`_on_submit` still drains its submit inline since it returns the `skill_id` by contract, but it only blocks on the submit API call, not the upload — that was already threaded.)

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
DavidDobas: Fixed in cb29f6f5 — start_training writes a small on-disk marker before the upload and clears it on any normal completion, so a process killed between upload and create_run resumes the run on next boot (the upload is HEAD-deduped, so it just finishes the remainder) instead of silently dropping it.

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
DavidDobas: Fixed in cb29f6f5 — a create_run (or upload) failure is now recorded as a synthetic "rejected" run carrying the error message, so the dashboard surfaces that the run didn't start instead of only logging it. Cleared on the next attempt.

webapp/js/training/main.js
DavidDobas: Handled from the backend in cb29f6f5 — a create_run/upload failure is now recorded as a synthetic "rejected" run carrying the error, so a failed start shows in the dashboard instead of vanishing once the hint clears. We left the hint logic as-is (the rejected run is the signal); if that row reads oddly we can revisit keeping the "Starting…" hint until a run/error lands.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
```diff
@@ -137,11 +193,8 @@ void TaskManager::save_metadata() {
...
     std::string metadata_path = data_dir + "/dataset_metadata.json";
 
-    std::ofstream file(metadata_path);
-    if (!file.is_open()) {
-        throw std::runtime_error("Failed to open metadata file for writing: " + metadata_path);
-    }
-    file << metadata_.dump(4);
+    MetadataLock lock(metadata_path);
+    write_json_atomic(metadata_path, metadata_);
 }
```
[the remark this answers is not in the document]
DavidDobas: Fixed in 86d155c4 — save_metadata() now merges the encoder-owned fields from a fresh disk read (under the same lock) before writing: it won't downgrade dataset_type from h264 back to h5, and it carries over video_files for episodes that have them on disk but not in the in-memory snapshot. So an add_episode / end_task during a concurrent convert can't strip the encoder's markers (the no-op-requeue → stuck-"preparing" path). Mirrors the converter's re-read on its side.

ros2_ws/src/cloud/innate_console/innate_console/console_node.py
```diff
@@ -0,0 +1,317 @@
...
+        # bucket key -> deque of recent records; OrderedDict for LRU eviction.
+        self._buckets = OrderedDict()
+        self._live = self.create_publisher(String, "/innate/console", 50)
+        self._backfill = self.create_publisher(String, "/innate/console/backfill", 1)
+        self.create_subscription(String, "/innate/console/request", self._on_request, 1)
+
+        # pane_id -> {"window", "pane", "path", "fh", "carry"}
+        self._panes = {}
+        self._wire_panes()
+        self.create_timer(1.0 / POLL_HZ, self._drain_panes)
```
vignesh-anand, line 123: **TODO / question: how resource-intensive is this node, and what's the aggregate footprint of all webapp-related processes on the robot?**

This PR adds several *always-on* processes to what's a resource-constrained on-board machine (Jetson). Before merge, can we get a full **CPU + memory** characterization (idle and under load) of the webapp-related stack, so we know the steady-state cost of having the cockpit running on boot?

For this node specifically (`innate_console`): it drains every tmux pane's capture file at 5 Hz and runs the regex ladder + JSON-encode **per line**, holds per-node ring buffers (200 buckets × 120 lines), and shells out to `tmux` on a 10 s rewire. What's its CPU at idle vs under a flooding node (e.g. a stuck reconnect loop hitting the 400-lines/drain cap), and its RSS with the buffers full?

Please profile the whole webapp-related set, since they run concurrently:
- [ ] `innate_console` (this node) — pane scraping, per-line regex/JSON at 5 Hz, ring buffers
- [ ] `dataset_encoder` — background gst/x264 encodes (nice/ionice), continuous
- [ ] `webapp/proxy/https_server.py` — static serve + `wss` rosbridge relay + media (multi-MB range reads, `cv2` thumbnail decode)
- [ ] `innate_training_node` — polling + upload threads
- [ ] rosbridge / rws — relays all of the above to the browser

Asks:
1. Per-process CPU% and RSS at **idle** (cockpit open, nothing happening) and under representative **load** (live teleop video + an encode running + a debugging client streaming).
2. Aggregate headroom on the target board — does the cockpit-on-boot stack leave enough for the actual robot control/perception nodes?
3. Confirm the encoder's nice/ionice + the proxy's `to_thread` offload actually keep these off the critical path under contention.

Not a blocker on the code itself — but "we added a browser cockpit that runs on boot" deserves a measured power/thermal/CPU budget on the real hardware before it's always-on.

webapp/proxy/https_server.py
```diff
@@ -0,0 +1,507 @@
...
+    rid = (qs.get("id") or [""])[0]
+    rel = (qs.get("file") or [""])[0]
+    if base is None or not rid or not rel:
+        return _plain(404, "Not Found", "not found")
+    run_dir = (base / rid).resolve()
+    target = (run_dir / rel).resolve()
+    if not run_dir.is_relative_to(SKILLS_ROOT) or not target.is_relative_to(run_dir) or not target.is_file():
+        return _plain(404, "Not Found", "no such log file")
+    try:
+        data = target.read_bytes()
```
vignesh-anand: **Reads the whole file into memory before truncating — OOM vector on large files.** (Medium)

`read_bytes()` loads the *entire* file into RAM, and the `MAX_LOG_BYTES` (8 MB) cap below only trims what's *sent* — it doesn't bound the *read*. So peak memory is the full file size, not 8 MB.

Two things make this real on the Jetson, not theoretical:
- This route serves **any file under the run dir**, not just `.log` text (the guard only checks containment + `is_file()`), and run dirs contain **multi-GB `.pth` checkpoints** — `/run/log?...&file=<checkpoint>.pth` pulls the whole checkpoint into RAM.
- Even normal use bites: a long run's log can be hundreds of MB, all loaded just to return 8 MB.

It's in `to_thread` so it won't block the event loop, but `to_thread` doesn't help memory — it allocates the full file on the heap, reachable by any LAN client on `:4443`.

One-line fix — bound the *read*:
```python
with open(target, "rb") as fh:
    data = fh.read(MAX_LOG_BYTES + 1)   # never hold more than ~8 MB
truncated = b""
if len(data) > MAX_LOG_BYTES:
    data = data[:MAX_LOG_BYTES]
    truncated = b"\n\n[truncated]\n"
```
The `+1` lets you still detect "file was longer than the cap" without reading the rest. Same client-visible behavior; peak memory ~8 MB regardless of file size.

(Optional hardening: reject obviously-binary/oversized targets up front so a `.pth` returns a clean 400 instead of 8 MB of binary as `text/plain` — but the bounded read is the actual fix.)

webapp/proxy/https_server.py
```diff
@@ -0,0 +1,507 @@
...
+
+
+def episode_response(request, qs: dict) -> Response:
+    """GET /episode?dir=<skill_dir>&id=<n>&camera=<cam> → episode MP4 (Range)."""
+    base = _resolve_under_root((qs.get("dir") or [""])[0])
+    eid = (qs.get("id") or [""])[0]
+    cam = (qs.get("camera") or [""])[0]
+    if base is None or not eid or not cam:
+        return _plain(404, "Not Found", "not found")
+    mp4 = (base / "data" / f"episode_{eid}_{cam}.mp4").resolve()
```
vignesh-anand: **Minor: unhandled `ValueError` on bad bytes in `id`/`camera` (and `file`/`id` in the run routes).**

These routes embed query params straight into a path and call `.resolve()` *outside* any try/except (unlike `_resolve_under_root`, which wraps it). A NUL or otherwise illegal byte in `id`/`camera` makes `Path(...).resolve()` raise `ValueError`, which propagates out of the `to_thread` task as a 500 / dropped request instead of a clean 404.

Not a traversal — the `is_relative_to(SKILLS_ROOT)` guard still contains everything; just sloppy error handling. Same pattern in `joints_response`, `run_info_response`, `run_log_response`. Wrap the resolves (or validate the params) so malformed input returns 404.

webapp/proxy/https_server.py
```diff
@@ -0,0 +1,507 @@
...
+        return _plain(404, "Not Found", "no such episode video")
+    # Cache beside data/ (not inside it) so thumbnails are never uploaded to the cloud.
+    cache = base / "thumbs" / f"episode_{eid}_{cam}.jpg"
+    try:
+        if not cache.is_file():
+            lock = _thumb_locks.setdefault(str(cache), asyncio.Lock())
+            async with lock:
+                if not cache.is_file():  # another request may have generated it while we waited
+                    await asyncio.to_thread(_make_thumb, mp4, cache)
+            _thumb_locks.pop(str(cache), None)
```
vignesh-anand: **Minor: `_thumb_locks` entry leaked when generation fails.**

The `_thumb_locks.pop(...)` runs *after* the `async with lock` block, so if `_make_thumb` raises inside the block the exception propagates to the outer `except` and this `pop` is skipped — leaving a stale lock in the dict. Functionally harmless (a later request reuses it and retries), but a persistently-failing thumb leaks one entry per cache path. Move the `pop` into a `finally` around the `async with`.

webapp/proxy/https_server.py
```diff
@@ -0,0 +1,507 @@
...
+            str(cert),
+            "-days",
+            "3650",
+            "-nodes",
+            "-subj",
+            f"/CN={hostname}.local",
+            "-addext",
+            f"subjectAltName={','.join(sans)}",
+        ],
+        check=True,
```
vignesh-anand: **Minor: hard dependency on `openssl` is masked by the restart loop.**

`check=True` means if `openssl` is missing or fails, `ensure_cert` raises → `main()` exits → the tmux `while true; do …; sleep 2; done` wrapper respawns the process every 2s. The result is a silent crash loop serving nothing, with no alert — the root cause (no openssl / cert-gen failure) never surfaces. Consider catching cert-generation failure and logging a clear one-time error (or failing loudly) rather than letting it loop invisibly.

webapp/proxy/https_server.py
```diff
@@ -0,0 +1,507 @@
...
+    if not run_dir.is_relative_to(SKILLS_ROOT) or not run_dir.is_dir():
+        # Not downloaded yet (or never will be).
+        body = json.dumps({"downloaded": False, "has_checkpoint": False, "files": []}).encode()
+        return Response(
+            200, "OK", Headers({"Content-Type": "application/json", "Content-Length": str(len(body))}), body
+        )
+    files = []
+    has_ckpt = False
+    try:
+        for p in sorted(run_dir.rglob("*")):
```
vignesh-anand, line 359: **Minor: unbounded `rglob("*")` over the run dir.**

This walks and lists *every* file under the run directory into the JSON response. Run dirs can hold many checkpoint shards / artifacts, so the listing can get large and slow. It's in `to_thread` so it won't block the event loop, but there's no cap on file count or response size. Consider a sane limit (or only listing the entries the UI actually needs).

config/dds/setup_dds.zsh
```diff
@@ -17,3 +17,10 @@ export ZENOH_SESSION_CONFIG_OVERRIDE='transport/shared_memory/enabled=true;trans
 export ZENOH_ROUTER_CONFIG_OVERRIDE='transport/shared_memory/enabled=false;transport/shared_memory/transport_optimization/pool_size=4194304;transport/link/tx/queue/congestion_control/drop/wait_before_drop=900000;transport/link/tx/queue/congestion_control/drop/max_wait_before_drop_fragments=900000'
 export ZENOH_CONFIG_OVERRIDE="$ZENOH_SESSION_CONFIG_OVERRIDE"
 # export RUST_LOG=debug
+
+# Standardize every node's console output so each line is self-describing:
+# severity, timestamp, and logger/node name. This is the contract the
+# observability tooling relies on to attribute logs by node (see
+# innate_console bridge + the webapp Debugging page). It matches rcl's default
+# layout; pinning it explicitly guards against drift and per-node overrides.
+export RCUTILS_CONSOLE_OUTPUT_FORMAT='[{severity}] [{time}] [{name}]: {message}'
```
vignesh-anand, line 28: **Minor (placement): a console log-format export doesn't belong in the DDS/Zenoh config file.**

`setup_dds.zsh` is, by name and contents, the **middleware** config — RMW selection and the Zenoh session/router overrides. `RCUTILS_CONSOLE_OUTPUT_FORMAT` is a **logging/observability** concern, unrelated to DDS, so it's surprising to find here and easy to miss (someone editing logging behavior wouldn't think to look in the DDS setup).

It works (anything sourcing this file inherits it), but the coupling is incidental — it's here because this file happens to be sourced everywhere, not because it's the right home. Suggest moving it to a dedicated logging/observability env snippet (or a general runtime-env file) that's sourced alongside this one, so the "log format is a contract the console bridge depends on" intent lives where someone would actually look for it. At minimum, if it stays, the comment already here is good — but the file name still mislabels it.

Not blocking — purely organization.

scripts/launch_ros_in_tmux.sh
```diff
@@ -83,21 +87,23 @@ process_command_group() {
...
+    # |-delimited command gets its own pane (the split makes it active, so we
+    # send to the window's active pane rather than a fixed pane index).
+    tmux send-keys -t $SESSION_NAME:"$window_name" "$PANE_SETUP_CMD && ${commands[1]}" C-m || return 1
+
+    local idx
+    for (( idx = 2; idx <= ${#commands[@]}; idx++ )); do
         tmux split-window -h -c ~ -t $SESSION_NAME:"$window_name" || return 1
         sleep 0.1
-        tmux send-keys -t $SESSION_NAME:"$window_name".1 "$second_cmd_full" C-m || return 1
+        tmux send-keys -t $SESSION_NAME:"$window_name" "$PANE_SETUP_CMD && ${commands[$idx]}" C-m || return 1
```
vignesh-anand: **`send-keys` now targets the *active* pane (no index), so it's timing-dependent — introduced by this rewrite.**

The previous version addressed panes by explicit index — `send-keys -t "$window_name".0` and `".1"` — which has no active-pane dependency. Generalizing to N panes, this rewrite switched to `send-keys -t "$window_name"` (the *active* pane) and leans on `split-window` + `sleep 0.1` having made the intended pane active. tmux commands are server-synchronous so the new pane *is* active when `split-window` returns and it usually works — but correctness now rides on an implicit global ("active pane") + a fixed sleep, which a slow/loaded box, a tmux hook, or a layout side-effect can perturb. (To be clear: the old code wasn't perfect either, but it didn't depend on active-pane state — this specific fragility is new here.)

Clean fix — address panes by their stable ID instead of "whatever's active". `new-session`/`new-window`/`split-window` all print the created pane's id with `-P -F '#{pane_id}'`:

```sh
local pane_id
if [ $group_index -eq 1 ]; then
    pane_id=$(tmux new-session -d -s $SESSION_NAME -n "$window_name" -c ~ -P -F '#{pane_id}') || return 1
else
    pane_id=$(tmux new-window -t $SESSION_NAME -n "$window_name" -c ~ -P -F '#{pane_id}') || return 1
fi
tmux send-keys -t "$pane_id" "$PANE_SETUP_CMD && ${commands[1]}" C-m || return 1

for (( idx = 2; idx <= ${#commands[@]}; idx++ )); do
    pane_id=$(tmux split-window -h -c ~ -t $SESSION_NAME:"$window_name" -P -F '#{pane_id}') || return 1
    tmux send-keys -t "$pane_id" "$PANE_SETUP_CMD && ${commands[$idx]}" C-m || return 1
done
```

`send-keys -t "$pane_id"` hits exactly the pane you just created regardless of timing/active state, and the `sleep 0.1`s stop being load-bearing for correctness (can drop them). Strictly more robust than both the old index approach and the current active-pane one.

(Bigger alternative: pass the command straight to `split-window "zsh -c 'source … && cmd'"` — no keystroke injection at all — but the pane then closes when the command exits, losing the drop-to-shell-on-crash ergonomics. Pane IDs keep the current UX.)

ros2_ws/src/cloud/clients/training-client/training_client/src/episode_converter.py
```diff
@@ -228,18 +275,32 @@ def _encode_camera_to_mp4(
             "!",
             "videoconvert",
             "!",
+            # Force 4:2:0 chroma. Without this, BGR input negotiates to H.264
+            # 4:4:4 (profile High 4:4:4 Predictive), which hardware decoders —
+            # phones and the Jetson's own nvv4l2decoder — cannot decode and
+            # render black. I420 yields a universally decodable 4:2:0 stream.
+            "video/x-raw,format=I420",
```
vignesh-anand, line 342: **MP4 format changed, but pre-existing episodes aren't re-encoded — old-format files persist in their broken state.**

This PR changes the H.264 output (`_encode_camera_to_mp4`): forces 4:2:0 (`video/x-raw,format=I420`, was negotiating to undecodable-on-phones **4:4:4**), `key-int-max={fps}` (~1s GOP for cheap seeks), and `faststart=true`. Good fixes — but the converter is **idempotent: this skips any episode whose MP4 already exists**. So:

- episodes encoded **after** this PR get the new, decodable format ✅
- episodes already converted by the **old** converter (pre-PR cloud-sync) keep their **old** MP4s — they're never regenerated just because the pipeline changed.

Net: any episode in the field from before this PR will **still render black on phones / scrub poorly**, and nothing here upgrades it. The only way to force a re-encode is to delete the existing `episode_*_*.mp4` so it's rebuilt from the raw in `raw_data/`.

Is there a migration path for old-format episodes (a version stamp on the MP4 / a one-time re-encode pass), or is the assumption that there are no pre-PR H.264 episodes to worry about? If the latter, worth stating explicitly.

**Consumer-compatibility (for reference, not a problem):** the new format is *more* compatible, not less — nothing that decoded the old 4:4:4 will fail on 4:2:0 (it's the reverse), and the consumers that were broken (phones, Jetson `nvv4l2decoder`) now work. The only subtle change: anything using the MP4 as a *training-pixel* source now sees 4:2:0-subsampled chroma instead of 4:4:4 — negligible for manipulation, and full-fidelity frames remain in `raw_data/`, but episodes differ subtly pre/post.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
```diff
@@ -258,4 +374,130 @@ std::tuple<bool, std::string, std::string> TaskManager::get_task_metadata_by_dir
...
+        if (it->value("episode_id", -1) == episode_id) {
+            episodes.erase(it);
+            break;
+        }
+    }
+    if (episodes.size() == before) {
+        return {false, "Episode " + std::to_string(episode_id) + " not found"};
+    }
+
+    // number_of_episodes doubles as the next-id counter (see add_episode); leave
```
vignesh-anand, line 490: **Serious: per-episode delete silently breaks a previously-structural invariant — episode ids were *guaranteed* contiguous until this PR.**

Leaving `number_of_episodes` as a monotonic next-id counter is correct for *assigning* ids (it stops a new episode reusing a deleted slot), but it makes `episodes[]` ids **sparse**, and that's a behavior change with wide blast radius — because contiguity wasn't just a convention, it was guaranteed *by construction* until now:

- Recording is append-only: `add_episode` always uses `id = number_of_episodes` then bumps it.
- Pre-PR there was **no per-episode delete at all** — only whole-skill `DeleteSkill`. The curation workaround was to recreate the skill and copy the keepers, which starts fresh at `number_of_episodes = 0` and **re-bases ids to a clean `0..N-1`**.

So for the entire history of this format, `episode_id ∈ 0..N-1` and `number_of_episodes == len(episodes)`, with no gaps — an invariant **no code path could violate**. Every downstream reader (the trainer/ACT, the mobile app, uploader, analytics scripts) was *correct* to rely on it. This PR's in-place delete is the first thing that can produce a gap, so consumers that were correct yesterday can be wrong today with no change on their side:

- **Counting via `number_of_episodes`** now overcounts (field is named like a count but is really "next id").
- **`for i in range(number_of_episodes): open(f"episode_{i}.h5")`** hits a deleted id → FileNotFound / crash, or a silently skipped/duplicated sample in a training dataset.
- **Re-deriving the next id as `len(episodes)`** (instead of the counter) now collides with an existing file → overwrite/corruption. (The robot's own `add_episode` is safe; external tools that do this are not.)

The root footgun: `number_of_episodes` now means *next id*, but its name says *count*, and it *used* to be a valid count — so the wrong reading is the natural one.

Asks:
1. Rename to `next_episode_id` (keep both for back-compat if needed) and add an explicit `episode_count = len(episodes)`, so nothing is tempted to read the counter as a count.
2. Document in the dataset-format spec that **episode ids are now sparse/non-contiguous** — readers must iterate `episodes[]` by stored `episode_id`, never by range.
3. Audit the known consumers (ACT loader, mobile app, uploader) for range/`number_of_episodes`-based iteration before this ships, since they predate the invariant being breakable.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
DavidDobas: Added in d682b599 — `write_json_atomic` now `fsync`s the tmp file before the rename and `fsync`s the parent directory after, so the rename can't reach disk ahead of the file's data. (The same rewrite also adds the write-error check greptile flagged.)

ros2_ws/src/brain/manipulation/src/task_manager.cpp
```diff
@@ -1,13 +1,120 @@
...
+void write_json_atomic(const std::string& path, const Json& j) {
+    const std::string tmp_path = path + ".tmp";
+    {
+        std::ofstream out(tmp_path);
+        if (!out.is_open()) {
+            throw std::runtime_error("Failed to open " + tmp_path + " for writing");
+        }
+        out << j.dump(4);
+    }  // flush + close before the rename
+    fs::rename(tmp_path, path);
```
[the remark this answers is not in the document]
DavidDobas: Fixed in d682b599 — rewrote `write_json_atomic` to write via POSIX `write()` and throw on a short/failed write **before** the `rename`, so a disk-full / I/O error can no longer promote a partial `.tmp` over the good file. It also `fsync`s the tmp file and the parent dir (vignesh's durability note on the same helper).

webapp/proxy/https_server.py
DavidDobas: Fixed in d682b599 — `run_log_response` now does `open(...).read(MAX_LOG_BYTES + 1)` instead of `read_bytes()`, so peak memory is bounded regardless of file size (a multi-GB `.pth` request reads at most the cap, not the whole file). The truncation marker still fires.

webapp/proxy/https_server.py
DavidDobas: Fixed in d682b599 — added a `_safe_resolve()` helper (`resolve()` → `None` on `ValueError`/`OSError`) and routed `episode`, `thumb`, `joints`, `run_info`, and `run_log` through it, so a NUL/illegal byte in `id`/`camera`/`file` returns 404 instead of a 500.

webapp/proxy/https_server.py
DavidDobas: Fixed in d682b599 — the `_thumb_locks.pop(...)` is now in a `finally` around the `async with lock`, so a failed `_make_thumb` no longer leaks the lock entry.

webapp/proxy/https_server.py
DavidDobas: Fixed in d682b599 — the openssl call is wrapped: `FileNotFoundError` (missing) and `CalledProcessError` (failed) now print a loud `FATAL …` to stderr — including openssl's own stderr, which `capture_output` had swallowed — before re-raising, so the cause is visible on each restart instead of a silent respawn.

webapp/proxy/https_server.py
DavidDobas: Fixed in d682b599 — the run-dir listing is capped at 2000 files with a `truncated` flag in the JSON. It still scans for the checkpoint marker but won't serialize an unbounded file list.

config/dds/setup_dds.zsh
DavidDobas: Addressed in d682b599 — rather than spin up a one-line file that both the launch script and `.zshrc` would have to source (more surface than it removes), I made the header honest: it's the shared ROS runtime env (DDS/Zenoh transport **plus** the console-logging format the observability tooling depends on), and the export keeps its justification comment. Happy to split it into a dedicated logging snippet if you'd still prefer that.

scripts/launch_ros_in_tmux.sh
DavidDobas: Fixed in d682b599 — each pane is now addressed by its unique pane-id: the first via `display-message -p '#{pane_id}'`, each split via `split-window -P -F '#{pane_id}'`, then `send-keys -t "$pane_id"`. No active-pane assumption, and the `sleep` is gone.

ros2_ws/src/cloud/clients/training-client/training_client/src/episode_converter.py
DavidDobas: Fixed in f113d647 — old 4:4:4 episodes are now detected by an `encoder_version` stamp and transcoded in place to 4:2:0. Their raw frames are already gone (the H5 was stripped), so it transcodes from the MP4 itself: software `avdec_h264` (the only decoder that reads 4:4:4) → I420 → `x264enc` + faststart, `.part` + atomic rename. No pixel-format probe — the 4:2:0 fix and this stamp ship together, so 'below version' is exactly 'old 4:4:4' (and the robot has no ffprobe). The stamp is written once the MP4s are confirmed current, so it's a one-time pass folded into the existing converter scan, not the update step. Verified end-to-end on a synthesized 4:4:4 file.

ros2_ws/src/brain/manipulation/src/task_manager.cpp
DavidDobas: Re-scoped after a consumer audit — I don't think this is Serious in practice. Nothing reconstructs filenames from `range(N)` or indexes `episodes[id]`: the converter iterates `episodes[]` keyed by `file_name`; the uploader enumerates real files on disk (`_enumerate_files`, not `range(count)`); the webapp builds from `meta.episodes`, sorts by numeric id, and calls services with the actual `episode_id`; the mobile app finds episodes by id-match and builds MP4 URLs from the actual id. So sparse ids already work everywhere. The one real fallout was `number_of_episodes` being read as a *count* — fixed in d682b599 (`read_local_episode_count` now uses `len(episodes)`), so `uploaded_episode_count` no longer overcounts after a delete. If you'd still rather re-densify ids on delete I can, but I don't think any consumer needs it.

ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py
```diff
@@ -374,6 +418,211 @@ def _on_create_run(self, req: CreateRun.Request, res: CreateRun.Response) -> Cre
...
+    def _run_training_flow(self, skill_dir: str, training_params: dict | None) -> None:
+        """Background worker: submit → (idempotent) upload → create_run, off the
+        executor thread. Surfaces a failure as a synthetic 'rejected' run so the
+        dashboard shows it. Clears the resume marker on any normal completion, so
+        only an abrupt kill (which skips the finally) leaves it for a boot-resume.
+        """
+        skill_id: str | None = None
+        try:
+            gen = self._mgr.submit(skill_dir)
+            skill: SkillInfo | None = None
```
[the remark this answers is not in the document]
DavidDobas: Fixed in 2bcdfd30 — the resume marker is now cleared the **instant** `create_run` returns success, *before* the `run_status` round-trip that previously sat inside the kill window. So a crash/power-cut after the run is created no longer leaves the marker for the next-boot resume to re-run (which is what created the duplicate billed run). The `finally` still clears it for the upload-failed / create_run-failed / early-return paths.

A residual single-call window remains (a kill *between* `create_run` committing on the cloud and it returning); fully closing that needs a cloud-side idempotency key on `create_run`, which I left as a follow-up.

PR #451, opened 2026-06-18
Single config/settings.yaml file for tunable parameters (.env for cloud/env config)

theo-michel:
## Why

Tunable parameters were scattered across six mechanisms — ROS 2 `declare_parameter` defaults, per-package YAML, launch args, `.env`, hardcoded constants, and directive prompts. Someone who wanted to change *"how fast the robot drives"* had no single place to look or edit.

## What this PR does

Two files, two jobs:

| File | What goes here |
|---|---|
| **`config/settings.yaml`** | How the robot **behaves** — tunable ROS parameters: driving speed, camera, arm, manipulation, navigation, brain-runtime knobs, TTS voice |
| **`.env`** | How it **reaches the cloud** + environment — secrets (`INNATE_SERVICE_KEY`), all cloud endpoint URLs (`BRAIN_WEBSOCKET_URI`, `TELEMETRY_URL`, `UNINAVID_WS_URL`, `INNATE_PROXY_URL`, …), and extra agent/skill dirs |

`config/settings.yaml` is a **native ROS 2 params file** (no custom parser). Each launch file layers it last so it wins over the package defaults:

```python
Node(parameters=[package_defaults.yaml, *settings_params()])   # last wins
```

- `/**` sections apply to every node that declares the parameter (one `motion_control.max_speed` reaches the joystick, the keyboard, and the app at once); named sections target one node.
- `config/settings.yaml.template` is committed with every knob commented out + its default; `config/settings.yaml` (gitignored) is auto-created from it on update / sim setup, and user edits are never overwritten.
- An all-commented file = "no overrides", so default behavior is unchanged until you uncomment something.

## What's tunable

| Stanza (node) | Keys |
|---|---|
| `/**` `motion_control` | `max_speed`, `max_angular_speed` |
| `/**` `inflation_layer` | `inflation_radius`, `cost_scaling_factor` |
| `joystick_controller` | `joystick.slow_mode_factor` |
| `bringup` | `battery.warning_percentage` / `critical_percentage`, `safety.max_speed` / `safety.max_angular_speed` (hard `/cmd_vel` clamp) |
| `mars_arm` | `max_jerk`, `stress_enabled` |
| `main_camera_driver` | resolution, fps, jpeg quality, auto-exposure knobs |
| `manipulation_server` | `inference_hz`, `speed`, `n_action_steps`, `temporal_ensemble_coeff` |
| `navigation_grid_localizer` | `max_score_threshold`, `max_range`, `auto_localize_timeout` |
| `brain_client_node` / `input_manager_node` | camera fov, pose-image cadence, jpeg quality, TTS voice, STT/realtime models |
| `uninavid_node` | VLN cadence + completion tuning |

## The one bit of magic: the nav2 remap

Native layering matches params **by name**, and our drive nodes all use `motion_control.*` — so one `/**` stanza reaches the joystick, the keyboard, and the app together. nav2 names the same concept per-plugin (`InnateFollowPath.vx_max`, `FollowPath.max_vel_x`, the velocity smoother's `[x, y, theta]` lists, costmap `inflation_layer`), so a thin remap (`load_motion_limit_overrides` / `load_costmap_rewrites` in `config_loader.py`) translates the single knob onto each schema at launch.

Underneath both, bringup owns the hard `/cmd_vel` clamp before the motors — its own `safety.max_speed` / `safety.max_angular_speed` parameter (default `0.4` / `2.5`), **separate from `motion_control`** so tuning driving feel can never lower the safety backstop. Even if a teleop or planner limit were missed, that clamp still enforces the real ceiling. Keep `safety` ≥ `motion_control` and ≥ `nav`.

## Knob-delivery fixes (knobs that silently did nothing)

Wiring this up surfaced several "turn the knob, nothing happens" bugs, now fixed:

- **App drive joystick** (`app.cpp`) scaled to hardcoded caps and never received the override file → now declares + reads `motion_control.max_speed`/`max_angular_speed`, and `app.launch.py` / `app.sim.launch.py` layer `settings.yaml`.
- **`input_manager_node`** received no override file at all, and its voice / OpenAI-model knobs were filed under the wrong node section → now wired with `settings_params()` and given an `input_manager_node:` stanza, with `cartesia_voice_id` sourced from `.env` so both speech paths stay in sync.

## Removed: `config/os.toml`

Earlier commits in this PR routed overrides through `config/os.toml` + a hand-rolled TOML parser. That's fully removed in favor of the native `settings.yaml` path:

- Deleted the TOML parser, `_load_os_config`, `load_param_overrides`, and the `os.toml` template/readers (`env_loader.py`, sim launcher, `print_runtime_env.py`).
- `.env` is now the real source for `BRAIN_WEBSOCKET_URI` / `TELEMETRY_URL` / `CARTESIA_VOICE_ID` (the old skip-list that funneled them through os.toml is gone).
- **Hard cutover:** a value set only in a robot's old `os.toml` falls back to the built-in default until moved to `.env` / `settings.yaml`.

## Migration (existing robots)

`post_update.sh`:
- creates `config/settings.yaml` from the template **only if missing**, so an existing robot's tuning is never overwritten (there is no `overrides.yaml` rename — that intermediate name never shipped; create-if-missing is the whole migration);
- seeds `/etc/innate.env` with `INNATE_SERVICE_KEY` (write-once, root-owned `600`) so the key survives a repo reset/reinstall; the repo `.env` still wins when it sets the key;
- stamps a leftover `config/os.toml` as deprecated in-place and warns (idempotent);
- migrates deprecated `.env` URLs to the new endpoints (existing behavior), and warns once when a previously-ignored `.env` key (`BRAIN_WEBSOCKET_URI` / `TELEMETRY_URL` / `CARTESIA_VOICE_ID`) is now honored, so a stale value can't silently reactivate.

## Testing

- `env_loader` smoke-tested: `.env`-only load, `settings_params()` resolves `config/settings.yaml`, the nav remap produces the correct per-plugin keys; missing file → `[]` / `{}` (launch unaffected).
- All edited Python compiles; `settings.yaml(.template)` is valid YAML; node-name keys are unique so PyYAML can't silently drop a duplicate `/**` block.
- `pre-commit` passes (ruff, ruff-format, clang-format, and the custom "launch files layer settings.yaml" hook).
- **Not run:** live in-sim `ros2 param get` after uncommenting a value (needs a full `colcon build` in the sim Docker).


ros2_ws/src/mars_bot/mars_bringup/mars_bringup/env_loader.py
```diff
@@ -233,6 +60,169 @@ def get_env(key: str, default: str = "") -> str:
...
+    return Path(innate_root) / "config" / "settings.yaml"
+
+
+def _load_settings_yaml() -> dict:
+    """Parse config/settings.yaml (a native ROS 2 params file). Returns ``{}`` when the
+    file is missing, empty/all-commented, or unreadable."""
+    try:
+        import yaml
+    except ImportError:
+        return {}
```
theo-michel: Why would we not be able to import yaml?

config/settings.yaml.template
karmanyaahm, line 1: love this

docs/PARAMETERS.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Going deeper (advanced parameters)
+
+Not surfaced in `settings.yaml` (edit the package YAML directly): per-joint arm PID gains (`mars_arm/config/arm_config.yaml`), full nav2 planner/controller/AMCL config (`mars_nav/config/`), stereo depth filters (`mars_cam`). Per-skill/model-coupled manipulation params (`chunk_size`, per-skill `n_action_steps`) live in each skill's `behavior_config`.
+
+## Adding a knob
+
+1. Add a commented stanza to `config/settings.yaml.template` targeting the node + parameter name.
+2. Make sure that node's launch loads overrides last: `parameters=[pkg_yaml, *settings_params()]`.
+   For a nav2 differently-named param, add a remap case in `env_loader.load_motion_limit_overrides`.
```
karmanyaahm: does what is exposed on the web app depend on these knobs? would be nice. we can have an advanced mode to show everything, and in normal mode just the onees defined inthe config.

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
```diff
@@ -233,6 +62,161 @@ def get_env(key: str, default: str = "") -> str:
...
+    overrides: dict = {}
+    if lin is not None:
+        for key in mapping.get("max_speed", ()):
+            overrides[key] = float(lin)
+    if ang is not None:
+        for key in mapping.get("max_angular_speed", ()):
+            overrides[key] = float(ang)
+        if schema == "dwb":
+            overrides["FollowPath.min_vel_theta"] = -float(ang)
+    return overrides
```
karmanyaahm, line 339: i'm a little concerned that the whole nav2 rewrite is DTM, but it's probably fine. we can remove if necessary.

ros2_ws/src/mars_bot/mars_control/config/motion_control.yaml
```diff
@@ -1,7 +1,7 @@
 /**:
   ros__parameters:
     motion_control:
-      max_speed: 0.6
+      max_speed: 0.4
       max_angular_speed: 1.0
```
karmanyaahm, line 5: why

ros2_ws/src/mars_bot/mars_nav/launch/mode_manager.launch.py
```diff
@@ -90,12 +91,17 @@ def generate_launch_description():
     )
 
     # Shared velocity smoother node
+    # config/settings.yaml's /** motion_control overrides the smoother's [x, y, theta] limit lists,
+    # keeping the lateral/reverse components from the YAML default.
+    smoother_limit_overrides = load_motion_limit_overrides(
+        "smoother", defaults=load_yaml_param_defaults(smoother_params_file)
+    )
     velocity_smoother_node = Node(
```
karmanyaahm, line 97: needs to be tested thoroughly (just a nav integration test)

karmanyaahm (approved):
Do you think it's better to merge it now or after the release, just so we have more time to test?

docs/PARAMETERS.md
theo-michel: Yeah in the future!

ros2_ws/src/mars_bot/mars_control/config/motion_control.yaml
theo-michel: It was already like this everywhere

theo-michel: Just it was defined in multiple places, but the value that it was clamped to was 0.4 but we can bump it back later I guess

.env.template
vignesh-anand, line 1: We had discuessed remioving the service key from the repo. I think thats a good idea and we should do it in this PR

vignesh-anand, line 1: .env file is hidden by default right, stuff likle extra skills etc i think should be more easily exposed?

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
vignesh-anand, line 1: I think think this fle should be in bringup. Bring up is like the lowe level robot hardware level code, talks to pcbs etc etc. Seems out of place

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
```diff
@@ -8,16 +8,7 @@
...
-    import tomllib
-except ModuleNotFoundError:
-    tomllib = None
-
-ENV_KEYS_MOVED_TO_OS_CONFIG = {
-    "BRAIN_WEBSOCKET_URI",
-    "TELEMETRY_URL",
-    "CARTESIA_VOICE_ID",
-}
+import yaml
```
vignesh-anand, line 18: Now that this module reads `settings.yaml` too, the module name and top-of-file docstring (`"""Environment loader... Loads .env file"""`) are stale — the bulk of the file is now ROS-param plumbing (`settings_params`, the nav remap, `load_yaml_param_defaults`), not env loading.

Not a bug, but consider renaming to `config_loader.py` or at least updating the module docstring to mention `settings.yaml`. Otherwise the next person greps for the settings.yaml logic and never thinks to look in a file called `env_loader`.

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
```diff
@@ -233,6 +62,161 @@ def get_env(key: str, default: str = "") -> str:
...
+
+
+def _load_settings_yaml() -> dict:
+    """Parse config/settings.yaml (a native ROS 2 params file). Returns ``{}`` when the
+    file is missing, empty/all-commented, or unreadable."""
+    path = _settings_yaml_path()
+    if not path.exists():
+        return {}
+    try:
+        data = yaml.safe_load(path.read_text())
```
vignesh-anand, line 208: **A malformed `settings.yaml` silently wipes all tuning, with no warning** — this is the big one.

`_load_settings_yaml` catches `yaml.YAMLError` and returns `{}`. So a single YAML typo *anywhere* in the file → `settings_params()` returns `[]`, `load_motion_limit_overrides` returns `{}`, `load_costmap_rewrites` returns `{}` → every override the operator set is silently dropped and the robot reverts to package defaults.

And unlike `_load_key_value_env` (which prints to stderr on error), this path is completely silent. Someone tunes their robot, makes one indentation mistake, and all their tuning vanishes with zero feedback — which is exactly the class of bug this PR set out to kill.

At minimum it should `print(..., file=sys.stderr)` here like the `.env` path does, so a broken settings.yaml is loud instead of silent.

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
```diff
@@ -233,6 +62,161 @@ def get_env(key: str, default: str = "") -> str:
...
+        return {}
+
+    if schema == "smoother":
+        return _smoother_limit_overrides(lin, ang, defaults or {})
+
+    mapping = _NAV_MOTION_SCALAR_KEYS.get(schema, {})
+    overrides: dict = {}
+    if lin is not None:
+        for key in mapping.get("max_speed", ()):
+            overrides[key] = float(lin)
```
vignesh-anand, line 328: **The int→double type crash is asymmetric**, which makes it a debugging trap rather than a clean fail-fast.

The same `/**` value travels two paths. The nav remap path is **safe** — here (and `load_costmap_rewrites`) we explicitly `float(value)` before injecting, so `max_speed: 1` becomes `1.0` for nav2. But the **native layering** path (`settings_params()` hands the raw file straight to the node) does no coercion, so the same `max_speed: 1` crashes bringup/keyboard/joystick (they declare `Parameter.Type.DOUBLE`) and the C++ arm node (`max_jerk`).

So one line in one file crashes half its consumers and silently works for the other half. The error surfaces as a deep rclcpp `InvalidParameterTypeException` from a random node, never pointing back at the config line the operator edited — and which nodes crash depends on what is launched. A symmetric "any int crashes everything" would actually be easier to diagnose.

We can't simply `float()` the native path (it deliberately hands the raw file to ROS — the whole no-custom-parser premise). Realistic fix: a launch-time validation pass that scans settings.yaml for known-double keys typed as int and raises one clear message (`settings.yaml: motion_control.max_speed must be a decimal, e.g. 0.4 not 4`) instead of the stack trace.

config/settings.yaml.template
```diff
@@ -0,0 +1,118 @@
...
+#
+# `/**` applies to every node that declares the parameter; a named section targets one node.
+# Use decimals for float values (e.g. 0.4, not 4) so ROS keeps them as floats.
+#
+# config/settings.yaml is auto-created from this template on update; your edits are never
+# overwritten. Secrets, cloud endpoint URLs, and extra agent/skill dirs live in .env (they
+# aren't ROS params, so not here); TTS voice is a ROS param and lives in this file.
+
+# ── Driving speed ─────────────────────────────────────────────────────
+# One knob for the whole robot: bringup clamps every /cmd_vel to it, joystick & keyboard
```
vignesh-anand: I do not think one knob for the whole robot is the right model here. Teleop and the autonomous nav2 controller should have **separate** speed params.

These are genuinely different regimes: nav2 controller limits are usually tuned *down* for safety and reliability (planner stability, obstacle margins, predictable stopping), whereas manual control can and often should be more aggressive — an operator driving by hand wants the full speed envelope. Collapsing both onto a single `motion_control.max_speed` means you cannot turn nav2 down for safety without also throttling manual driving, and you cannot give teleop a more aggressive cap without also speeding up autonomy.

I would rather see two params — e.g. a teleop max speed (bringup clamp + joystick/keyboard/app) and a separate autonomous/nav2 max speed — so the two can be tuned independently. The nav remap could still exist, but it should map from the *autonomous* param, not the shared one.

Open to discussion, but as designed this couples two things that we deliberately want decoupled.

ros2_ws/src/brain/manipulation/config/recorder.yaml
```diff
@@ -12,3 +12,8 @@
     odom_topic: "/odom"                    # Topic for odometry data (used to verify robot is alive)
     image_size: [640, 480]
     max_timesteps: 1800                    # Maximum number of timesteps before automatically stopping episode
+    # Learned-policy inference (manipulation_server)
+    inference_hz: 25.0                      # policy inference loop rate (Hz)
+    speed: 1.5                             # action execution speed multiplier
+    n_action_steps: 0                      # replan horizon; 0 = auto (min(40, chunk_size)), per-skill config wins
+    temporal_ensemble_coeff: 0.0           # ACT temporal-ensemble smoothing; 0 = disabled
```
vignesh-anand: These are inference / policy-execution params, not data-collection params — they should not live in `recorder.yaml`.

`recorder.yaml` is loaded by two nodes: the actual data-collection `recorder_node` (recorder.launch.py) and `manipulation_server` (behavior.launch.py loads the whole file). Because the file is under `/**`, these four knobs get dumped onto **both** nodes — `manipulation_server` declares and uses them, but `recorder_node` (data collection) gets them too and they are meaningless there. They only landed here because behavior.launch.py reuses recorder.yaml as the manipulation server's param source (it needs `data_directory`).

Conceptually inference-loop-rate / action-speed / replan-horizon / ACT smoothing are manipulation-server runtime knobs, not recording config, and do not belong in a file named `recorder.yaml`.

This is also internally inconsistent with the rest of the PR: settings.yaml.template files these overrides under a `manipulation_server:` section, but their defaults live here under `/**`. The clean version is a dedicated `manipulation_server.yaml` (with a `manipulation_server:` section) loaded by behavior.launch.py, leaving recorder.yaml as data-collection-only.

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
```diff
@@ -39,9 +39,9 @@
       
       # Velocity limits (from existing DWB config)
       vx_min: -0.2                      # matches min_vel_x
-      vx_max: 0.45                       # matches max_vel_x
+      vx_max: 0.4                        # matches max_vel_x
       vy_max: 0.0                       # differential drive - no lateral motion
-      wz_max: 0.6                       # matches velocity_smoother max
+      wz_max: 1.0                       # matches velocity_smoother max
```
vignesh-anand: Changing autonomous controller limits without explicit testing is dangerous, and these changes go out by default (not gated behind an override).

`wz_max` is being **raised** 0.6 → 1.0 — a 67% increase in the autonomous controller's top angular velocity — and `vx_max` lowered 0.45 → 0.4. These are not just "expose a knob" edits; they retune how the robot plans and executes motion out of the box. Controller velocity limits interact with planner stability, obstacle clearance/inflation, lookahead, and stopping distance — raising angular speed in particular can change cornering behavior, overshoot, and collision margins.

The PR description itself notes the in-sim path was not run ("Not run: live in-sim `ros2 param get` after uncommenting a value"). Retuning the controller and shipping it as the new default without a nav integration test on hardware (or at least sim) is the kind of change that should be isolated and validated on its own, not bundled into a config-plumbing PR.

Recommendation: either revert the controller/smoother/sim-params value changes to their current defaults (keep this PR purely about the config mechanism), or split them into a separate, explicitly-tested nav-tuning PR with before/after behavior.

ros2_ws/src/mars_bot/mars_bringup/config/robot_config.yaml
```diff
@@ -10,7 +10,7 @@
       critical_percentage: 10
     motion_control:
       max_speed: 0.4
-      max_angular_speed: 2.5
+      max_angular_speed: 1.0
```
vignesh-anand: This is a 60% cut to the robot's default **teleop** turn rate (`max_angular_speed` 2.5 → 1.0), shipping by default — and it is a different subsystem from the nav2 controller change. This is the bringup `/cmd_vel` clamp / manual-drive ceiling, not the autonomous planner. Same caution as the controller.yaml comment applies: retuning the default turn rate should be deliberate and tested, not bundled into a config-plumbing PR.

It also is the clearest evidence for keeping teleop and autonomy as **separate** params (see my comment on settings.yaml.template). Before this PR the values were deliberately ~4x apart: teleop/bringup angular = 2.5, nav2 controller `wz_max` = 0.6. That gap is not an accident — manual control was tuned to be aggressive while the autonomous controller was tuned down for safety/reliability, exactly the distinction we want to preserve. This PR collapses both toward 1.0, erasing an intentional difference. A single shared knob cannot represent "teleop 2.5, nav2 0.6."

ros2_ws/src/mars_bot/mars_nav/config/nav2_navigation_params_sim.yaml
```diff
@@ -62,13 +62,13 @@ controller_server:
...
+      max_vel_x: 0.4
       min_vel_y: 0.0
       max_vel_y: 0.0
-      min_vel_theta: -1.5
-      max_vel_theta: 1.5
+      min_vel_theta: -1.0
+      max_vel_theta: 1.0
       min_speed_xy: 0.0
-      max_speed_xy: 0.45
+      max_speed_xy: 0.4
```
vignesh-anand: Same untested-retune concern as controller.yaml (flagged there). These are the sim DWB limits being changed by default: `max_vel_x`/`max_speed_xy` 0.45 → 0.4 and `max_vel_theta` 1.5 → 1.0 (a 33% cut to autonomous angular). Retuning the controller and shipping it as the new default belongs in an explicitly-tested nav-tuning PR, not bundled into config plumbing. Either revert to current defaults or split + validate.

ros2_ws/src/mars_bot/mars_nav/config/velocity_smoother.yaml
```diff
@@ -7,8 +7,8 @@ velocity_smoother:
     velocity_timeout: 0.2           # Seconds. If no commands received for this duration, output zero velocity.
 
     # Limits using flat list structure [x, y, theta]
-    max_velocity: [0.45, 0.0, 0.6]   # Based on controller max_vel_x, max_vel_y, max_vel_theta
-    min_velocity: [-0.2, 0.0, -0.6]  # Based on controller min_vel_x, min_vel_y, symmetric min_vel_theta
+    max_velocity: [0.4, 0.0, 1.0]    # Based on controller max_vel_x, max_vel_y, max_vel_theta
+    min_velocity: [-0.2, 0.0, -1.0]  # Based on controller min_vel_x, min_vel_y, symmetric min_vel_theta
```
vignesh-anand: Same untested-retune concern as controller.yaml (flagged there). The smoother limits change by default: `max_velocity` [0.45, 0, 0.6] → [0.4, 0, 1.0] and `min_velocity` theta -0.6 → -1.0 — i.e. the autonomous angular ceiling is **raised** 0.6 → 1.0 here while it is cut elsewhere. The smoother is the last stage before `/cmd_vel`, so this directly shapes executed motion. Should be validated on hardware/sim, not shipped silently in a config PR.

ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
```diff
@@ -377,6 +378,11 @@ class AppControl : public rclcpp::Node {
         // Joystick: invert steering while reversing (car-like). Off by default;
         // the app's Dev tab toggles it at runtime through set_parameters.
         this->declare_parameter("reverse_steering", false);
+        // Drive-joystick speed caps. Declared with the pre-tuning defaults so behavior is
+        // unchanged with no override; config/settings.yaml's /** motion_control block layers
+        // on top and wins (see app.launch.py), sharing the one knob with the other teleop nodes.
+        this->declare_parameter("motion_control.max_speed", 0.5);
+        this->declare_parameter("motion_control.max_angular_speed", 1.0);
```
vignesh-anand: Default mismatch: the app drive joystick defaults to `max_speed` **0.5**, while every other path this PR unified uses **0.4** (motion_control.yaml for keyboard/joystick, robot_config.yaml for the bringup clamp, nav2 configs). And app.launch.py loads only `*settings_params()` — no package YAML — so with no override set, this node's only source for the cap is this hardcoded C++ literal.

So in the unoverridden default case: physical joystick / keyboard / bringup / nav = 0.4, but the mobile-app drive joystick = 0.5. The mismatch disappears only once someone sets the `/**` knob (then all read the same value) — i.e. it breaks exactly the "behavior is unchanged / one knob for the whole robot" case the PR advertises.

Severity depends on topology: if the app's /cmd_vel flows through the bringup clamp (0.4), the extra 0.1 is silently clipped and this 0.5 is a misleading dead default; if it bypasses the clamp, the robot genuinely drives ~25% faster from the app by default. Either way it should be reconciled.

This is also the one node whose motion default lives in **code, not a config file**, which is why it drifted — concrete evidence that defaults are not actually single-sourced. Either align to 0.4, or (per my settings.yaml.template comment) make a deliberate separate teleop param rather than a silently-divergent literal.

config/settings.yaml.template
theo-michel: ❤️

ros2_ws/src/brain/brain_client/brain_client/perception/camera.py
```diff
@@ -123,7 +123,7 @@ def depth_block(self) -> dict | None:
 
     def arm_block(self) -> dict | None:
         if self._config.send_arm_camera_image and self.last_arm_camera is not None:
-            arm_b64 = image_codec.encode_jpeg_b64(self.last_arm_camera)
+            arm_b64 = image_codec.encode_jpeg_b64(self.last_arm_camera, self._config.jpeg_quality)
```
vignesh-anand: I know this PR did not introduce this, but exposing `jpeg_quality` here surfaces what looks like a massive waste, and I think it is worth fixing in this PR rather than codifying it behind a knob.

**The problem: we double-compress (transcode) every image we send to the agent.**

brain_client subscribes to `CompressedImage` (line 52) — i.e. the frames arrive **already JPEG-compressed** at the camera driver's `jpeg_quality` (80). Then `_on_image` / `_on_arm` `cv2.imdecode` them back to raw BGR, and `latest_image_b64` / `arm_block` re-encode to JPEG@70. So every frame to the agent goes **JPEG@80 → decode → JPEG@70**: wasted CPU on the Jetson and a second lossy pass (the agent sees a doubly-degraded image). It also makes the new knob misleading — setting it above 80 is pure waste, since you cannot recover detail the source already threw away.

Per path:
- **Arm camera (worst): the decode is pure waste.** `last_arm_camera` is decoded and then immediately re-encoded in `arm_block`, with nothing touching the pixels in between (its only other reference is `clear_after_send`). We could forward the original compressed `msg.data` bytes and delete both the decode and the re-encode.
- **Main single image:** same double-compression in `latest_image_b64`.
- **Video clip:** the raw frames are buffered (`image_buffer`, ~55 MB at the default 60-frame buffer) solely to feed `encode_video_b64`, which uses `cv2.VideoWriter(MJPG)` — and MJPEG is just JPEG frames in an AVI. Source JPEG → decode → buffer raw → re-encode JPEG. Nothing in the codebase ever uses the raw pixels for any actual CV op (no resize/transform) — every consumer ultimately wants JPEG, and the input is already JPEG.

Net cost: extra CPU per frame, ~55 MB of raw frames in RAM, and compounded compression artifacts to the agent — all to satisfy OpenCV encoder APIs.

Suggested fix: forward the original compressed bytes where possible. Arm + single-image are trivial (just base64 `msg.data`). Video needs muxing the original JPEGs into an MJPEG AVI without re-encoding (imageio/ffmpeg or manual AVI muxing) — more involved, but it removes the decode, the re-encode, and shrinks the buffer ~25x. At minimum, the arm path should stop decoding for nothing.

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
```diff
@@ -36,172 +27,12 @@ def _load_key_value_env(path: Path) -> None:
             if line and not line.startswith("#") and "=" in line:
                 key, value = line.split("=", 1)
                 key = key.strip()
-                if key in ENV_KEYS_MOVED_TO_OS_CONFIG:
-                    continue
                 value = value.strip()
                 if (value.startswith('"') and value.endswith('"')) or (value.startswith("'") and value.endswith("'")):
                     value = value[1:-1]
                 os.environ[key] = value
```
vignesh-anand, line 53: **Dead `.env` keys can now silently activate on upgrade — subtle, but worth guarding.**

Before this PR, `.env` values for `BRAIN_WEBSOCKET_URI` / `TELEMETRY_URL` / `CARTESIA_VOICE_ID` were ignored — the `ENV_KEYS_MOVED_TO_OS_CONFIG` skip-list dropped them here so they would be funneled through os.toml. This PR removes that skip-list, so `os.environ[key] = value` now applies to those keys too: `.env` is read for them.

Consequence: a robot with a stale/leftover value for one of these in `.env` — left there precisely *because* it was being ignored — will have it silently take effect after the upgrade. Someone's old `BRAIN_WEBSOCKET_URI=...` that has been inert in `.env` for months could suddenly redirect the robot to a different cloud. Nobody expects a value that was dead to activate on an update, so this will be hard to diagnose.

This is the exact inverse of the os.toml hard-cutover risk (values silently lost) — here values are silently *gained*. Suggest `post_update.sh` detect when `.env` already contains any of these keys and log a clear warning (e.g. "BRAIN_WEBSOCKET_URI in .env is now active; it was previously ignored — confirm it is correct"), so a surprise activation is surfaced rather than silent.

scripts/update/post_update.sh
```diff
@@ -218,6 +219,43 @@ fi
...
+# its values now live in .env (cloud URLs, agent/skill dirs) and settings.yaml (TTS voice).)
+# -----------------------------------------------------------------------------
+SETTINGS_CONFIG="$REPO_DIR/config/settings.yaml"
+SETTINGS_TEMPLATE="$REPO_DIR/config/settings.yaml.template"
+
+if [ ! -f "$SETTINGS_CONFIG" ] && [ -f "$SETTINGS_TEMPLATE" ]; then
+    log "Creating config/settings.yaml from template..."
+    cp "$SETTINGS_TEMPLATE" "$SETTINGS_CONFIG"
+    chown "$ACTUAL_USER:$ACTUAL_USER" "$SETTINGS_CONFIG"
+    log "  Created $SETTINGS_CONFIG (uncomment a stanza to tune a parameter)"
```
vignesh-anand, line 256: The PR description documents a migration that is not in this code. The body says post_update "renames an existing config/overrides.yaml -> config/settings.yaml (preserving a robot's tuning) before the create-from-template step" — but there is no such rename here (and no reference to overrides.yaml anywhere in source). The git history shows it was added (`fix: migrate config/overrides.yaml -> settings.yaml on update`) and then removed (`removed mentions of config/overrides.yaml`), but the description was not updated to match.

I do not think `overrides.yaml` ever reached a deployed robot — it looks like an intermediate name purely within this PR's own development — so the lost-tuning risk is most likely moot. Flagging anyway for two reasons:

1. The description is now stale/misleading to reviewers: it implies existing tuning is preserved across the rename, which this code does not do.
2. If any dev/test robot ever ran the `overrides.yaml` phase, it has `config/overrides.yaml`, gets no migration, and this block creates `settings.yaml` fresh from the all-commented template (create-only-if-missing does not help — the filenames differ) -> that robot silently loses all its tuning.

Could you confirm overrides.yaml never shipped? If so, just fix the PR description. If it might have, restore the rename/migration before the create-from-template step.

tests/test_launch_param_overrides.py
```diff
@@ -0,0 +1,131 @@
...
+            elif isinstance(func, ast.Attribute):
+                names.add(func.attr)
+    return names
+
+
+def _spawns_node(called: set[str]) -> bool:
+    return bool(called & _NODE_CALLS)
+
+
+def _is_override_aware(called: set[str]) -> bool:
```
vignesh-anand: Not major, can be ignored — but worth knowing the limit. This is a single boolean for the whole file: it returns true if the file mentions **any** override helper **anywhere**. It does not verify that the override is applied to *every* node, that the *right* override is used, or that it is even in a `parameters=[...]`.

Concrete consequence: `navigation_sim.launch.py` calls `load_motion_limit_overrides("dwb")`, so it passes — even though it never calls `load_costmap_rewrites`, so `inflation_layer.*` overrides silently do nothing in sim. The test built to catch "a launch path forgot the override" cannot catch "a launch path forgot *one of* the overrides." Same for a multi-node file where only one node is wired.

So the green check means "this file knows the override system exists," not "this file is correctly wired" — easy to over-trust. Not asking to fix it (data-flow checking is a lot more work); just flagging so the guard is not read as stronger than it is.

tests/test_launch_param_overrides.py
```diff
@@ -0,0 +1,131 @@
...
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+_LAUNCH_GLOB = "ros2_ws/src/**/launch/*.launch.py"
+
+# Calls that instantiate a node which can take ROS parameters.
+_NODE_CALLS = {"Node", "ComposableNode"}
```
vignesh-anand: Not major, can be ignored — coverage caveat. Detection is name-based AST over a glob, so it only sees a literal `Node(...)` / `ComposableNode(...)` call that is textually present in a `*.launch.py` under `launch/`. Two blind spots:

1. Nodes built by a helper/factory are invisible. `make_remote_throttle_nodes()` constructs `ComposableNode`s, but it lives in `mars_cam/mars_cam/remote_throttle_nodes.py` (a plain module, not a `*.launch.py`) — outside the glob, and the launch file that calls it sees the name `make_remote_throttle_nodes`, not `ComposableNode`. So those nodes are never checked for override-awareness.
2. The glob misses `.launch.xml`, non-`*.launch.py` names, and nodes started from scripts/C++.

Net: anything indirected through a helper or outside the glob is unguarded (there is already one real instance in-tree). Fine to leave as-is — just noting the guard is a useful heuristic, not full coverage.

vignesh-anand (dismissed):

scripts/update/post_update.sh
theo-michel: yeah sorry need to update the PR description

.env.template
theo-michel: Ok interesting

ros2_ws/src/brain/brain_client/brain_client/perception/camera.py
theo-michel: Intersting

.env.template
theo-michel: Okay!

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
theo-michel: Fixed

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
theo-michel: fixed

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
theo-michel: Fixing

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
theo-michel: Yeah reverting those changes

ros2_ws/src/mars_bot/mars_bringup/config/robot_config.yaml
theo-michel: Yeah reverting those changes

ros2_ws/src/mars_bot/mars_nav/config/nav2_navigation_params_sim.yaml
theo-michel: Yeah reverting those changes

ros2_ws/src/mars_bot/mars_nav/config/velocity_smoother.yaml
theo-michel: Yeah reverting those changes

ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
theo-michel: Yeah reverting those changes

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
theo-michel: Good instinct — there's a real cohesion smell, but I'd reframe it before moving anything:

- `config_loader` is **hardware-agnostic shared glue** (`.env` + `settings.yaml` parsing), imported by ~19 launch files across nearly every package (`mars_cam`, `mars_arm`, `mars_nav`, `mars_control`, `brain_client`, `cloud/uninavid`, `manipulation`). It's a foundational utility, not bringup-specific.
- On ROS2 convention, a `*_bringup` package is the top-level launch/config aggregator, so shared launch-time config glue living here is actually idiomatic. The genuinely out-of-place residents are the *opposite* — `uart.py` / `i2c.py` / `battery.py` are the low-level hardware drivers that don't fit a "bringup" package.
- The real wart isn't the file's name, it's the **dependency direction**: because it lives in `mars_bringup`, every package now hard-depends on a heavyweight package (`rplidar_ros`, `mars_sim`, `tf2`, the UART/I2C drivers) just to parse a YAML file. `brain_client`/`cloud/uninavid` pulling in the lidar package for config is the smell.

So the fix isn't moving it toward hardware — it's extracting a thin, dependency-free `mars_config`/`mars_common` package that everyone depends on instead. That's pure churn (~19 import sites + package wiring) and unrelated to this PR's settings logic, so I'd rather not mix it in here.

Tracked as a follow-up: https://linear.app/innate/issue/INN-602/extract-config-loader-into-a-dependency-free-shared-package-mars

ros2_ws/src/brain/brain_client/brain_client/perception/camera.py
theo-michel: Fixed

.env.template
theo-michel: Done

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
theo-michel: There was no os config in the previous release this was just a change from the prevous main pre release branch

config/settings.yaml.template
theo-michel: Decoupled them

ros2_ws/src/mars_bot/mars_bringup/mars_bringup/config_loader.py
theo-michel: Fixed

PR #452, opened 2026-06-19
fix(sim): transport reliability, WebRTC streaming, and dev-env polish

karmanyaahm:
Gets the Innate simulator stack working end-to-end and moves the web UI off the sim's HTTP API onto the rosbridge connection it already shares with the brain.

## Web UI ↔ brain now over rosbridge
Directives, active skills, execute/cancel skill, brain-active, reset-brain, reset-position, set-environment, the agent list, chat, and brain-backend status all talk **directly over rosbridge** instead of the sim's FastAPI routes. Each switch was audited against the bridge's actual relay format before collapsing (which caught two real bugs: a missing `updated_at` that broke the connection-stability debounce, and `brain_backend_config` not omitting empty fields — "clear" vs "unchanged").

New sim-control rosbridge services on the bridge:
- `/sim/reset_position` — `std_srvs/Trigger`
- `/sim/set_environment` — `brain_messages/ChangeMap`-typed, applied on its own async task so the ~30s apply never starves topic publishing

The sim's HTTP server is reduced to the tiny surface the **launcher** still needs: `/video_feeds_ready`, `/stack_metrics`, `/sim_log_config`. `chat_api.py` and all the brain/skill/directive routes are deleted.

## Camera video now over WebRTC
Camera feeds move from MJPEG to **WebRTC** (aiortc, VP8), streamed peer-to-peer from the sim process to the browser. Only the SDP/ICE **signaling** rides rosbridge (the `/webrtc/*` topics, the same handshake the robot streamer uses) — the media itself never touches ROS.

- **Lazy encoding:** a camera is only VP8-encoded while a browser is actually watching it, driven by an additive `/webrtc/active_streams` signal. Idle cameras cost nothing.
- **Dynamic cameras:** each track is labelled with its camera name in the SDP msid, and only cameras that are actually rendering are offered. The frontend derives the camera set + identities straight from the offer SDP — no hardcoded list — so robot and sim share one viewer code path and differ only by `config.json`'s rosbridge URL.
- The MJPEG `/video_feed*` endpoints and the frontend's `` path are removed.
- The conflicting C++ `mars_cam` WebRTC streamer is disabled in the sim tmux launch (the Python server replaces it for the sim).

## Also on this branch
- Containerized web frontend with runtime-configurable URLs
- CLI: `innate sim setup` / `clean` / `sh`, skill-asset download on `up`; removed the top-level `innate setup` alias
- Deps: `requirements.in` + compiled lock (now incl. `aiortc`)
- Sim depth published as 16UC1 millimeters to match the real robot (mars_cam) — halves the wire payload vs 32FC1
- Costmap view: fix a WebGL context leak (the Three.js renderer was recreated on every map-view toggle → eventually "Map renderer unavailable")
- ROS/CI: restore exec bits on ROS node scripts, symlink-install test pass, teleop + camera switching in the web UI

## Notes
- Sim WebRTC targets a public-IP deployment: STUN only, no TURN (the host candidate is directly reachable); open the negotiated UDP ports for remote viewing.
- The robot's C++ streamer is unchanged — the frontend already drives it through the same dynamic-camera path (generic names until its offer carries msid labels).
- `requirements.macos.txt` still needs regenerating on macOS to pick up `aiortc`.
- The launcher keeps its stdlib HTTP calls (it's a zero-dependency bootstrap CLI that runs before the venv exists).


karmanyaahm:
@greptile-apps

sim/frontend/src/hooks/useRobotWebRTC.ts
```diff
@@ -316,7 +360,7 @@ export function useRobotWebRTC({
         }
 
         if (!hasMediaRef.current) {
-          setError("Robot signaling WebSocket closed.");
+          setError("Signaling WebSocket closed.");
```
vignesh-anand, line 363: **Silent mid-stream disconnect: no error, no auto-reconnect once media is flowing**

`onclose` only surfaces an error when no media has arrived yet (`if (!hasMediaRef.current)`). Once a track is playing, if the peer connection drops mid-stream this handler does nothing — and since this PR removes the `oniceconnectionstatechange` / `onsignalingstatechange` handlers, nothing else reacts either. The result is the `<video>` freezes on its last frame with no error shown and no recovery (retry is manual-only now that the old 5s readiness poll is gone).

On localhost this is rare and low-stakes. But this server targets remote viewing over a public IP (STUN-only UDP), where a mid-session drop is the *expected* failure mode — the viewer just sees a frozen image with no indication it's stale.

Suggest reacting to `pc.connectionState` / ICE state transitioning to `failed`/`disconnected` to surface an error and trigger `reconnect()`, or at minimum re-arm the error path in `onclose` even when media had previously been flowing.

ros2_ws/src/mars_bot/mars_sim_bringup/launch/sim_rosbridge.launch.py
```diff
@@ -16,6 +16,18 @@ def generate_launch_description():
                 "rosbridge_compatible": True,
```
vignesh-anand, line 16: **The 9090 control plane is unauthenticated *and* plaintext — there's no TLS in front of it**

`rws_server` is configured with only `port` + `rosbridge_compatible` — no token, no allow-list, no TLS. It will execute any `publish` / `call_service` / `call_action` it receives.

I went looking for the TLS layer and it isn't on this path:
- The frontend's Caddy has `auto_https off` and listens on `:80` only (sim/frontend/Caddyfile) — plain HTTP, and it doesn't proxy 9090. The browser connects straight to `ROBOT_WS_URL=ws://localhost:9090`, bypassing Caddy.
- `sim/docker-compose.dev.yml` maps `"9090:9090"` — all interfaces, not `127.0.0.1`.
- The only `wss://` in the repo is the brain→cloud backend (`wss://agent-v1.innate.bot`), a separate connection — not the browser→robot control link.

This server was always unauthenticated, but the stakes changed with the rosbridge migration: now that directives / skills / brain-lifecycle all ride 9090, anyone who can reach this port can drive the robot — `publish /brain/set_directive`, `call_service /brain/reset_brain`, fire `/execute_skill`, etc. The PR notes explicitly target a public-IP deployment ("open the negotiated UDP ports for remote viewing"), which would expose this.

Not necessarily a fix-in-this-PR item, but it should be locked down before that public-IP deploy: bind `127.0.0.1:9090` + SSH tunnel / WireGuard, or front it with an authenticated `wss://` reverse proxy. If there's already a prod proxy/firewall fronting 9090 that I didn't find, disregard.

sim/src/webrtc/webrtc_server.py
```diff
@@ -0,0 +1,172 @@
...
+    """Single-peer manager. A new `start` tears down the previous connection."""
+
+    def __init__(self, get_frame_for: Callable[[str], Optional[np.ndarray]]):
+        self._get_frame_for = get_frame_for
+        self._pc: Optional[RTCPeerConnection] = None
+        self._tracks: dict[str, CameraTrack] = {}
+
+    async def on_start(self, payload: dict) -> str:
+        """Build the peer connection + offer. Returns the offer SDP."""
+        await self.close()
```
vignesh-anand, line 52: Question (likely intentional): this is a single global session — one `WebRTCManager`, one `webrtc_signal_in/out` queue pair, no per-peer keying — and a new `start` tears down whatever connection was live. So a second browser merely *loading* the page (or the same user opening a second tab / reloading) evicts the current viewer's stream rather than getting its own session.

Is single-viewer the intended model here (matching the robot's C++ streamer), or should remote viewing support concurrent watchers? Just want it to be a conscious choice rather than a side effect of the single-manager design.

sim/src/webrtc/webrtc_server.py
```diff
@@ -0,0 +1,172 @@
...
+
+        pc = RTCPeerConnection()
```
vignesh-anand, line 54: Question (likely known): the server peer is created with no ICE servers, relying on the host candidate being directly reachable. That holds today because the sim runs natively on the (public-IP) host, so the host candidate is the real IP.

Worth confirming the assumption is written down somewhere, since it's load-bearing: if the sim is ever run in a bridge-networked container the host candidate becomes a private `172.x` and remote ICE silently fails (no STUN/TURN fallback on the server side), and viewers behind symmetric NAT / restrictive firewalls can't connect either (no TURN). Intentional for this deployment topology?

sim/src/webrtc/webrtc_server.py
```diff
@@ -0,0 +1,172 @@
...
+    manager = WebRTCManager(lambda name: shared_queues.latest_frames.get(name))
+    sig_in: queue.Queue = shared_queues.webrtc_signal_in
+    sig_out: queue.Queue = shared_queues.webrtc_signal_out
+
+    print("[WebRTC] server started (3 cameras, lazy VP8 encoding)")
+    while not shared_queues.exit_event.is_set():
+        try:
+            item = sig_in.get_nowait()
+        except queue.Empty:
+            await asyncio.sleep(0.005)
```
vignesh-anand: nit: idle busy-poll — when there's no signaling this spins at ~200 Hz (`get_nowait()` + `sleep(0.005)`). Negligible in practice, but feeding an `asyncio.Queue` from the bridge (or `run_in_executor` on a blocking `get`) would drop the poll entirely. Not worth blocking on.

sim/setup.sh
```diff
@@ -22,20 +29,20 @@ if ! command -v uv &> /dev/null; then
...
-uv pip install -r "$REQUIREMENTS_FILE" --python .venv/bin/python
+uv pip install $QUIET -r "$REQUIREMENTS_FILE" --python .venv/bin/python
 
 # Install PyTorch nightly with CUDA 12.8 for Blackwell (RTX 50xx, sm_120) support
 # Stable torch builds top out at sm_90 and will fall back to CPU on RTX 5090
 GPU_NAME=$(nvidia-smi --query-gpu=name --format=csv,noheader 2>/dev/null | head -1 || echo "")
 if echo "${GPU_NAME}" | grep -qi "RTX 50\|B200\|B100\|blackwell"; then
     echo "⚡ Blackwell GPU detected (${GPU_NAME}) - installing PyTorch nightly cu128..."
-    uv pip install --python .venv/bin/python --pre --reinstall torch \
+    uv pip install $QUIET --python .venv/bin/python --pre --reinstall torch \
```
vignesh-anand, line 67: This `--reinstall torch` now fires on **every** `innate sim up`, not just `setup` — and on Blackwell GPUs that's a multi-GB reinstall before each launch.

Two changes combine: `ensure_sim_setup` (runtime.py) lost its fast-path and now runs `setup.sh` unconditionally, and `cmd_up` calls it with no `allow_setup` gate — so `up` always reaches here. This branch is then guarded only on GPU name, with `--reinstall` forcing an uninstall+reinstall regardless of what's already installed.

So on an RTX 50xx / B100 / B200 box, every `up` uninstalls and reinstalls the PyTorch cu128 nightly (~GB from cache, or a fresh download whenever the nightly version rotates) before the sim starts. Previously `up` skipped setup entirely when the venv was already good, so this is a new per-launch cost.

Could either:
- drop `--reinstall` (let uv no-op when torch is already satisfied), or guard this branch on the installed torch not already being the cu128 nightly; or
- restore a fast-path so `up` doesn't re-run `setup.sh` when the venv is already complete.

(Non-Blackwell hosts just pay a `requirements.ubuntu.txt` re-resolve each `up`, which uv makes cheap — it's the `--reinstall` that makes this sharp.)

sim/frontend/vite.config.ts
```diff
@@ -15,6 +15,10 @@ function gitCommitHash(): string {
 export default defineConfig({
   base: "/",
   plugins: [react()],
+  build: {
+    // Keep production bundles readable for debugging on the remote server.
+    minify: false,
```
vignesh-anand, line 20: nit: `minify: false` ships unminified JS to every client from the production image — including remote viewers over the public IP — so it's both a much larger payload and fully readable source on the wire. Fine as a deliberate debugging aid for the sim, but worth confirming it's intended to stay in the prod build path rather than being gated behind a dev/debug flag (e.g. `minify: mode === "development" ? false : "esbuild"`). Not blocking.

sim/frontend/src/hooks/useRobotWebRTC.ts
karmanyaahm: not gonna worry about video in webapp cuz the other webapp already handles it

ros2_ws/src/mars_bot/mars_sim_bringup/launch/sim_rosbridge.launch.py
karmanyaahm: yeah this is only public on the temporary sim

sim/src/webrtc/webrtc_server.py
karmanyaahm: yeah the sim only runs temporarily on a public google cloud or on your own machine

sim/src/webrtc/webrtc_server.py
karmanyaahm: tbd tbh ideally i find an efficient way to use the ros webrtc server for this cuz i'm implementing multi-stream there

sim/frontend/vite.config.ts
karmanyaahm: this app will die soon anyway

sim/src/webrtc/webrtc_server.py
karmanyaahm: fxied

sim/setup.sh
karmanyaahm: fixed, no need to reinstall always

PR #457, opened 2026-06-23
ACT improvements

theo-michel:
## Summary
- Change the default **Learned base speed** scale from `0.5` to `1.0` so the learned ACT policy plays back its predicted base `cmd_vel` at full speed by default (matching the replay path).

Updated in all three places that carry the default:
- `ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py` — `declare_parameter` default (and refreshed the now-stale "half speed" comment)
- `ros2_ws/src/brain/manipulation/config/manipulation_server.yaml` — config default
- `webapp/js/settings/catalog.js` — settings UI default

A per-skill `behavior_config` still overrides this.

### DIFF
--- a/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
+++ b/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
@@ -77,10 +77,9 @@ class ManipulationServer(Node):
         self.inference_hz = inference_hz
         self.policy_speed = self.get_parameter("speed").value
         # De-rate applied to the recorded/predicted base velocity at execution time. Replay and
-        # learned inference are tuned independently: replay plays back recorded cmd_vel 1:1 by
-        # default, the learned path de-rates the policy's predicted cmd_vel to half speed.
+        # learned inference are tuned independently; both play back base cmd_vel 1:1 by default.
         self.declare_parameter("replay_base_speed_scale", 1.0)
-        self.declare_parameter("learned_base_speed_scale", 0.5)
+        self.declare_parameter("learned_base_speed_scale", 1.0)
         self.replay_base_speed_scale = self.get_parameter("replay_base_speed_scale").value
         self.learned_base_speed_scale = self.get_parameter("learned_base_speed_scale").value
         # n_action_steps=0 means "auto" (min(40, chunk_size)); a per-skill value still wins.
--- a/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
+++ b/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
@@ -381,56 +381,66 @@ class ManipulationServer(Node):
                 f"Starting policy inference for {duration} seconds at {inference_hz} Hz (progress threshold: {progress_threshold})"
             )
 
-            while rclpy.ok():
-                loop_start = time.time()
-                elapsed_time = loop_start - start_time
-                iteration_count += 1
+            # Disable automatic GC for the run. Per-step allocations (tensors, numpy,
+            # feedback msgs) are freed by refcounting, but periodic gen-2 collections
+            # cause pauses that show up as latency spikes in the 25 Hz loop. freeze()
+            # moves already-live objects out of the scan set; gc is re-enabled in the
+            # finally so every exit path (return/break/raise) restores it.
+            gc.freeze()
+            gc.disable()
+            try:
+                while rclpy.ok():
+                    loop_start = time.time()
+                    elapsed_time = loop_start - start_time
+                    iteration_count += 1
 
-                # Check for cancellation
-                if self._cancel_requested.is_set():
-                    self.get_logger().info("Behavior execution canceled")
-                    self._stop_robot()
-                    if end_pose:
-                        self.call_arm_goto_service(end_pose, end_pose_time)
-                    return "CANCELLED", "User requested cancellation"
-
-                # Check if duration completed (moved earlier to prevent infinite loop)
-                if elapsed_time >= duration:
-                    self.get_logger().info(f"Behavior timeout reached after {elapsed_time:.2f} seconds")
-                    break
-
-                progress = self._run_inference_once()
-
-                # Check if inference failed due to missing sensor data
-                if progress is None:
-                    # Check if it's due to missing sensors
-                    if not self._check_sensor_availability():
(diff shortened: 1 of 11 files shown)


ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -77,10 +78,9 @@ def __init__(self):
         self.inference_hz = inference_hz
         self.policy_speed = self.get_parameter("speed").value
         # De-rate applied to the recorded/predicted base velocity at execution time. Replay and
-        # learned inference are tuned independently: replay plays back recorded cmd_vel 1:1 by
-        # default, the learned path de-rates the policy's predicted cmd_vel to half speed.
+        # learned inference are tuned independently; both play back base cmd_vel 1:1 by default.
         self.declare_parameter("replay_base_speed_scale", 1.0)
-        self.declare_parameter("learned_base_speed_scale", 0.5)
+        self.declare_parameter("learned_base_speed_scale", 1.0)
```
vignesh-anand, line 83: **The "per-skill `behavior_config` overrides this" reassurance doesn't hold — and there's no live way to set it either.**

This is a global, unconditional change: every learned skill uses this one value, with no per-skill override.

- `LearnedExecCfg` (the schema for a learned skill's `behavior_config`, `config_validation.py:137-150`) has **no base-speed field**, and `_BaseExecCfg` sets `extra="ignore"` (`config_validation.py:112`) — so a skill that puts `learned_base_speed_scale` in its `metadata.json` has it **silently dropped**, no error.
- `self.learned_base_speed_scale` is read once here and **never reassigned** (this is the only assignment); `_publish_base` (`:925`) always uses that single cached value.

And the one knob that *does* exist is node-level and cached at init: there's no `add_on_set_parameters_callback`, and the settings UI itself says *"restart the robot to apply"* (`webapp/js/settings/main.js:217`). So it's a global, restart-only setting.

Make the description's claim true, or drop it. Either remove the "a per-skill `behavior_config` still overrides this" line, or actually wire it: add `learned_base_speed_scale` to `LearnedExecCfg` and reassign `self.learned_base_speed_scale` per behavior in `_execute_behavior` so skills can set it individually.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -140,6 +145,12 @@ def __init__(self):
         self.arm_state_pub = self.create_publisher(Float64MultiArray, "/mars/arm/commands", 10)
         # Head position command (degrees) — only published for head-enabled replay skills.
         self.head_set_position_pub = self.create_publisher(Int32, "/mars/head/set_position", 10)
+        # Per-step inference timing breakdown (JSON String), for the webapp Profiling page.
+        # Only published while a learned behavior is executing, so it's free when idle.
+        self.inference_profile_pub = self.create_publisher(String, "/brain/manipulation/inference_profile", 10)
```
vignesh-anand, line 150: Consider a typed message here instead of JSON-over-`std_msgs/String`.

Right now the schema lives in two hand-maintained places with nothing keeping them in sync:
- the dict built in `_publish_inference_profile` (producer), and
- the `Sample` typedef in `webapp/js/profiling/main.js` (consumer).

And the consumer doesn't really validate the contract — `parseSample` only checks that `total_ms` is a number and trusts every other field. So if the producer ever drops or renames a field, the page won't reject the sample; the missing value silently becomes `undefined` and propagates into `mean`/`percentile` as `NaN`. A typed `.msg` (explicit `float64`/`bool` fields) gives a single source of truth, validates the whole shape for free, and avoids the `json.dumps`/`JSON.parse` round-trip.

It also sidesteps a real failure mode of the string approach: if any field is `NaN`/`Inf`, `json.dumps` emits a bare `NaN` token (invalid JSON), and the page's `JSON.parse` then rejects the **whole** sample — so the profiler goes blank exactly when a policy diverges. A typed numeric field carries non-finite values fine.

Tradeoff: a typed msg means a new definition in the interfaces package + a rebuild, and rosbridge would marshal the fields instead of a single string. Given this matches the existing `std_msgs/String`+JSON convention used elsewhere (RobotInfo, TTS), JSON-over-String is a reasonable v1 — just flagging that typed is the cleaner long-term shape if this profiler sticks around. (If we stay on String, at minimum `json.dumps(payload, allow_nan=False)` + sanitizing non-finite values, and validating the fields the page actually reads, would stop one bad/missing field from silently corrupting the stats.)

webapp/js/profiling/main.js
```diff
@@ -0,0 +1,743 @@
...
+ *   arm_jerk?:number, base_jerk?:number }} Sample
+ */
+
+/**
+ * @param {HTMLElement} root
+ * @returns {{ destroy: () => void }}
+ */
+function buildView(root) {
+  /** @type {Sample[]} */
+  let samples = [];
```
vignesh-anand, line 41: **Unbounded sample retention while recording.** `samples` has no cap and recording has no auto-stop, so this array grows for as long as Record is on — at 25 Hz that's ~1,500/min, ~90k/hour.

Two costs that climb without limit:
- **Memory**: ~13-field objects accumulating linearly — tens of MB/hour, no ceiling. Leave it recording (which is exactly what a profiler invites) and the tab bloats until it degrades.
- **CPU**: `computeStats` runs every render tick (~7 Hz) and sorts the *full* record several times (percentiles over total/inference/engine), and `renderHistogram` re-filters/sorts the full array too — all O(n log n) with n unbounded. At hundreds of thousands of samples the page starts to stutter, which ironically skews the latency tool it's supposed to be.

Worth noting the page *looks* bounded — `MAX_PLOT_POINTS = 400` caps the plotted points — but that's drawing only; the array and the stats deliberately use the full history, so the cost is hidden behind a smooth-looking UI.

Not a leak (every sample is reachable and Clear frees it); it's just growth without a ceiling. Short sessions — the intended use — are fine. Fix is a bound: ring-buffer `samples` to the last N (same idea already used for the 400 plotted points), or auto-stop after a set duration, or at minimum document an intended recording window.

vignesh-anand (commented):
Left comments

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
theo-michel: I created a ticket tbd

PR #458, opened 2026-06-23
Reduce WebRTC stream latency

theo-michel:
## What

Reduces glass-to-glass latency on the live WebRTC teleop feed. The sender pipeline was already well-tuned (`vp8enc deadline=1`, single-buffer leaky queues, non-blocking live `appsrc`), so this PR goes after the remaining buffering sources — and adds a profiling panel that measured the win on-device.

**Headline result (measured on the webapp, 30 fps):** receiver de-jitter buffer **75 ms → ~1 ms**, with **0% packet loss, 0 dropped frames, 0 freezes** — ~74 ms of receive-side latency removed. RTT held at 4–8 ms throughout.

### 1. Sender-side playout-delay RTP header extension — the main lever
`ros2_ws/src/mars_bot/mars_cam/...` (`webrtc_streamer.cpp/.hpp`, `CMakeLists.txt`, `camera_composable.launch.py`)

The receiver's de-jitter buffer was the dominant remaining latency: libwebrtc sat on a ~75 ms buffer on a clean LAN (RTT ~6 ms, jitter ~10 ms, 0% loss). Setting `jitterBufferTarget = 0` on the browser (below) only lowers the *floor* — the adaptive estimator still picked 75 ms because nothing capped the *ceiling*.

Send the WebRTC **playout-delay** RTP header extension from `webrtcbin` so the receiver's libwebrtc clamps its buffer to `[min, max]`. GStreamer 1.20.3 ships no element for this URI, so it's implemented as a minimal `GstRTPHeaderExtension` subclass added to each VP8 payloader; `webrtcbin` then emits the `a=extmap` line and the payloader writes the 3-byte field into every RTP packet.

`min`/`max` are launch params (`playout_min_delay_ms` / `playout_max_delay_ms`, default **0 / 40**), so the latency/smoothness tradeoff is retunable with an `innate restart` — no rebuild.

**Note on `min` vs `max`:** measurement showed the browser's `jitterBufferTarget = 0` pins the buffer near the floor, so the extension's **`max` is the effective lever** (the cap is what tamed the 75 ms); `min` is largely inert on the webapp and kept at 0. Documented in the launch file.

### 2. Browser receivers — minimize the receive-side jitter buffer
`webapp/js/webrtcSession.js` and `sim/frontend/src/hooks/useRobotWebRTC.ts`

On each incoming video track, set `jitterBufferTarget = 0` (standardized `RTCRtpReceiver` knob) and `playoutDelayHint = 0` (legacy Chrome knob); both are set when present — modern Chrome honors `jitterBufferTarget` and ignores the hint. Video receivers only — the audio receiver is left untouched so robot-mic NetEq isn't starved. Feature-guarded + try/catch. This unlocks the low-latency path that the playout-delay extension then bounds.

### 3. GStreamer sender — CBR with a small rate-control buffer
`webrtc_streamer.cpp`

Switch both VP8 encoders from default VBR (~6 s rate buffer) to `end-usage=cbr` with a ~0.5 s buffer, so the encoder stops holding frames to smooth bitrate.

### 4. WebRTC profiling panel (Teleop cockpit)
`webapp/js/teleop/profilingPanel.js` (+ `main.js`, `webrtcSession.js`, `app.css`)

A discrete toggle (top-right pulse button, or `p`) opens a live read-out of the receive side: **jitter buffer delay** (headline), RTT, jitter, fps, resolution, bitrate, loss, dropped frames, freezes. Rates are deltas between 1 s polls. Hidden by default; exposes `WebRtcSession.getStats()`. This is what was used to tune and validate the numbers above.

## Tradeoffs

The buffer changes trade a little tolerance for network/bitrate jitter (possible micro-stutter under load) for lower latency — the right call for teleop, and confirmed loss-free on this LAN at 30 fps.

## Mobile app

The extension is sender-side and receiver-agnostic, so the native libwebrtc in the phone app honors it too. But the app has no `jitterBufferTarget`, so its buffer sits at its adaptive estimate **clamped to ≤ `max`** (lands near 40 ms, not ~1 ms). Lowering `max` pulls the app's buffer down directly without affecting the webapp — to be tuned/verified on-device.

## Testing

- **On-device:** built with `innate build mars_cam` and validated live via the profiling panel — jitter buffer 75 → ~1 ms, 30 fps, 0 loss / 0 drops / 0 freezes. The brief 30 → 14 fps dip seen mid-tuning was camera auto-exposure (dim scene), not this change (`framesReceived == framesDecoded`, 0 drops).
- **Frontend:** `tsc --noEmit` clean on changed files (other errors are pre-existing).
- **Not yet measured:** true motion-to-photon (glass-to-glass) — receive-side stats are now near-optimal, so remaining latency is upstream (software VP8 encode), which is the case for hardware H.264.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
+++ b/ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
@@ -522,7 +522,11 @@ std::string WebRTCStreamer::build_pipeline_description(bool& with_audio) const {
         "caps=video/x-raw,format=BGR,width=640,height=480,framerate=30/1 ! "
         "queue leaky=downstream max-size-buffers=1 max-size-time=0 max-size-bytes=0 ! "
         "videoconvert ! "
-        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 ! "
+        // CBR with a small rate-control buffer stops the encoder from holding frames to
+        // smooth out the bitrate (the VBR default buffers ~6 s). ~0.5 s of buffering trades
+        // a little bitrate stability for lower glass-to-glass latency on the live link.
+        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 "
+        "end-usage=cbr buffer-size=600 buffer-initial-size=400 buffer-optimal-size=500 ! "
         "rtpvp8pay pt=96 ! "
         "application/x-rtp,media=video,encoding-name=VP8,clock-rate=90000,payload=96 ! "
         "webrtc.sink_0 "
@@ -531,7 +535,11 @@ std::string WebRTCStreamer::build_pipeline_description(bool& with_audio) const {
         "caps=video/x-raw,format=BGR,width=640,height=480,framerate=15/1 ! "
         "queue leaky=downstream max-size-buffers=1 max-size-time=0 max-size-bytes=0 ! "
         "videoconvert ! "
-        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 ! "
+        // CBR with a small rate-control buffer stops the encoder from holding frames to
+        // smooth out the bitrate (the VBR default buffers ~6 s). ~0.5 s of buffering trades
+        // a little bitrate stability for lower glass-to-glass latency on the live link.
+        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 "
+        "end-usage=cbr buffer-size=600 buffer-initial-size=400 buffer-optimal-size=500 ! "
         "rtpvp8pay pt=97 ! "
         "application/x-rtp,media=video,encoding-name=VP8,clock-rate=90000,payload=97 ! "
         "webrtc.sink_1";
--- a/ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
+++ b/ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
@@ -522,9 +522,6 @@ std::string WebRTCStreamer::build_pipeline_description(bool& with_audio) const {
         "caps=video/x-raw,format=BGR,width=640,height=480,framerate=30/1 ! "
         "queue leaky=downstream max-size-buffers=1 max-size-time=0 max-size-bytes=0 ! "
         "videoconvert ! "
-        // CBR with a small rate-control buffer stops the encoder from holding frames to
-        // smooth out the bitrate (the VBR default buffers ~6 s). ~0.5 s of buffering trades
-        // a little bitrate stability for lower glass-to-glass latency on the live link.
         "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 "
         "end-usage=cbr buffer-size=600 buffer-initial-size=400 buffer-optimal-size=500 ! "
         "rtpvp8pay pt=96 ! "
@@ -535,9 +532,6 @@ std::string WebRTCStreamer::build_pipeline_description(bool& with_audio) const {
         "caps=video/x-raw,format=BGR,width=640,height=480,framerate=15/1 ! "
         "queue leaky=downstream max-size-buffers=1 max-size-time=0 max-size-bytes=0 ! "
         "videoconvert ! "
-        // CBR with a small rate-control buffer stops the encoder from holding frames to
-        // smooth out the bitrate (the VBR default buffers ~6 s). ~0.5 s of buffering trades
-        // a little bitrate stability for lower glass-to-glass latency on the live link.
         "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 "
         "end-usage=cbr buffer-size=600 buffer-initial-size=400 buffer-optimal-size=500 ! "
         "rtpvp8pay pt=97 ! "
--- a/ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
+++ b/ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
@@ -1,5 +1,7 @@
 #include "mars_cam/webrtc_streamer.hpp"
 
+#include <gst/rtp/rtp.h>
+
 #include <sstream>
 #include <algorithm>
 #include <cctype>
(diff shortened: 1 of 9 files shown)


webapp/js/webrtcSession.js
```diff
@@ -211,6 +219,15 @@ export class WebRtcSession {
...
     const track = event.track;
+    const receiver = event.receiver;
+    if (receiver) {
+      try {
+        if ("jitterBufferTarget" in receiver) receiver.jitterBufferTarget = 0;
+        if ("playoutDelayHint" in receiver) receiver.playoutDelayHint = 0;
+      } catch {
+        // unsupported; default buffer applies
+      }
+    }
```
vignesh-anand, line 246: This tuning block runs **above** the `if (track.kind === "audio")` branch below, so it also zeroes the jitter buffer on the **audio** receiver (robot mic). The sim path deliberately avoids this — `useRobotWebRTC.ts` returns early for non-video before tuning.

Zeroing the audio NetEq buffer starves it under any network jitter → choppy/dropping mic audio, and the measured latency win here is video-only. Suggest gating on video (move below the audio check, or `if (track.kind === "video" && receiver)`).

Minor, while here: `jitterBufferTarget` is in **ms** but `playoutDelayHint` is in **seconds** — both `0` is coincidentally correct, but if anyone later sets a nonzero floor to match the robot-side `max=40`, the values differ (`40` vs `0.04`). Worth a units comment since this is the retune surface.

webapp/js/webrtcSession.js
```diff
@@ -211,6 +219,15 @@ export class WebRtcSession {
    */
   #onTrack(event, pc) {
     const track = event.track;
+    const receiver = event.receiver;
+    if (receiver) {
+      try {
+        if ("jitterBufferTarget" in receiver) receiver.jitterBufferTarget = 0;
+        if ("playoutDelayHint" in receiver) receiver.playoutDelayHint = 0;
```
vignesh-anand, line 242: Nit (units): `jitterBufferTarget` is in **milliseconds** but `playoutDelayHint` is in **seconds**. Both `0` is coincidentally correct, but this is the retune surface — if someone later sets a nonzero floor to match the robot-side `max=40`, the correct values diverge (`jitterBufferTarget = 40` vs `playoutDelayHint = 0.04`), and a copy-pasted `40` on the hint would be 1000× off. Worth a one-line units comment here.

vignesh-anand, line 242: Nit (wording/behavior): the PR frames `playoutDelayHint` as the "older Chrome fallback", but the code sets **both** knobs whenever both exist — it's not an either/or. Harmless in practice (modern Chrome honors `jitterBufferTarget` and both request minimal delay), but the "fallback" framing doesn't match what the code does. Either only set `playoutDelayHint` when `jitterBufferTarget` is absent, or adjust the wording.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
```diff
@@ -17,6 +19,81 @@ GstWebRTCPeerConnectionState peer_connection_state(GstElement* webrtc) {
...
+// GStreamer < 1.24 ships no built-in element for this URI, so we implement it as a minimal
+// GstRTPHeaderExtension subclass and add it to each payloader; webrtcbin then emits the matching
+// a=extmap line and the payloader writes the bytes into every RTP packet.
+//
+// Wire format (WebRTC experiment): 3 bytes = MIN delay (12 bits) | MAX delay (12 bits), 10 ms units.
+#define MARS_PLAYOUT_DELAY_URI "http://www.webrtc.org/experiments/rtp-hdrext/playout-delay"
+
+// One-byte extmap id, kept at the top of the 1..14 range so it does not clash with the low ids
+// webrtcbin auto-assigns to the extensions it manages itself (mid, transport-cc, ...).
+constexpr guint kPlayoutDelayExtId = 14;
```
vignesh-anand, line 37: Fragility note (not a bug today): `kPlayoutDelayExtId = 14` works on the tested GStreamer 1.20 — the measured latency win proves the extmap was honored, so nothing collides currently. But the comment's justification rests on an unchecked assumption: that webrtcbin keeps its own managed extensions (mid, transport-cc, …) in the low id range. That's undocumented internal behavior with no API guarantee. On a base-image bump to GStreamer ≥1.24 — where webrtcbin auto-manages payloader extensions — id 14 could be double-emitted or clash; the offer still builds and the node still reports PLAYING, so the latency win would silently regress with no error. Worth either pinning the GStreamer version in a comment here, or asserting there's no duplicate `extmap:14` in the generated SDP after create-offer.

webapp/js/webrtcSession.js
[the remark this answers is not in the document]
theo-michel: fixed

webapp/js/teleop/profilingPanel.js
```diff
@@ -0,0 +1,240 @@
...
+      return;
+    }
+
+    /** @type {any} */ let vid = null;
+    /** @type {any} */ let selectedPair = null;
+    /** @type {string | undefined} */ let selectedPairId;
+
+    report.forEach((/** @type {any} */ s) => {
+      if (s.type === "inbound-rtp" && s.kind === "video") vid = s;
+      else if (s.type === "transport" && s.selectedCandidatePairId) selectedPairId = s.selectedCandidatePairId;
```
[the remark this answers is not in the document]
theo-michel: fixed

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
```diff
@@ -17,6 +19,81 @@ GstWebRTCPeerConnectionState peer_connection_state(GstElement* webrtc) {
...
+
+gssize mars_playout_delay_write(GstRTPHeaderExtension* ext, const GstBuffer* /*input_meta*/,
+                                GstRTPHeaderExtensionFlags /*write_flags*/, GstBuffer* /*output*/, guint8* data,
+                                gsize size) {
+    if (size < 3) {
+        return -1;
+    }
+    auto* self = reinterpret_cast<MarsPlayoutDelayExt*>(ext);
+    const guint min_units = (self->min_delay_ms / 10) & 0xFFF;  // 12-bit, 10 ms units
+    const guint max_units = (self->max_delay_ms / 10) & 0xFFF;
```
vignesh-anand, line 67: Low severity: `& 0xFFF` masks rather than clamps. An out-of-range `playout_max_delay_ms` (> 40950 ms, or a negative launch int cast to `guint`) wraps to a small wrong value silently instead of clamping or erroring — and the params aren't validated on read either. The defaults are sane and no one sets a 40 s+ jitter cap, so this is marginal, but a `std::min(x, 0xFFFu)` clamp (plus maybe a warn) would make a misconfig fail visibly rather than produce a silently wrong cap.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
```diff
@@ -588,6 +677,50 @@ std::string WebRTCStreamer::build_pipeline_description(bool& with_audio) const {
...
+            if (pay)
+                gst_object_unref(pay);
+            if (caps_elem)
+                gst_object_unref(caps_elem);
+            continue;
+        }
+
+        GstCaps* caps =
+            gst_caps_new_simple("application/x-rtp", "media", G_TYPE_STRING, "video", "encoding-name", G_TYPE_STRING,
+                                "VP8", "clock-rate", G_TYPE_INT, 90000, "payload", G_TYPE_INT, track.payload, nullptr);
```
vignesh-anand, line 708: Minor (sync hazard): `payload` is a second source of truth for the pt — set here as `payload=track.payload` (96/97) and also on the payloader as `pt=96/97` in the pipeline string. The two must agree or the capsfilter won't intersect rtpvp8pay's output and caps negotiation fails. Since rtpvp8pay already emits `payload` from its `pt=`, you can drop `payload` from this caps (and the `VideoTrack::payload` field) and let the pipeline string be the single source — it passes through unchanged. Removes the divergence risk.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
```diff
@@ -522,18 +607,22 @@ std::string WebRTCStreamer::build_pipeline_description(bool& with_audio) const {
         "caps=video/x-raw,format=BGR,width=640,height=480,framerate=30/1 ! "
         "queue leaky=downstream max-size-buffers=1 max-size-time=0 max-size-bytes=0 ! "
         "videoconvert ! "
-        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 ! "
-        "rtpvp8pay pt=96 ! "
-        "application/x-rtp,media=video,encoding-name=VP8,clock-rate=90000,payload=96 ! "
+        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 "
+        "end-usage=cbr buffer-size=600 buffer-initial-size=400 buffer-optimal-size=500 ! "
```
vignesh-anand, line 611: Nit (units/rationale): these six numbers arrive bare, and `600/400/500` read like a KB byte budget at a glance — especially sitting next to `target-bitrate=2000000` (bits/s) — but they're **milliseconds** (`gst-inspect vp8enc`: buffer-size/initial/optimal are all ms, defaults 6000/4000/5000). A one-liner would remove the trap, e.g. `// CBR rate-control buffer in ms: cap banked bitrate at ~0.6 s so the encoder stops holding frames to smooth bitrate`.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
```diff
@@ -522,18 +607,22 @@ std::string WebRTCStreamer::build_pipeline_description(bool& with_audio) const {
...
 
         "appsrc name=src_arm is-live=true format=time "
         "caps=video/x-raw,format=BGR,width=640,height=480,framerate=15/1 ! "
         "queue leaky=downstream max-size-buffers=1 max-size-time=0 max-size-bytes=0 ! "
         "videoconvert ! "
-        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 ! "
-        "rtpvp8pay pt=97 ! "
-        "application/x-rtp,media=video,encoding-name=VP8,clock-rate=90000,payload=97 ! "
+        "vp8enc deadline=1 target-bitrate=2000000 cpu-used=4 error-resilient=partitions keyframe-max-dist=30 "
+        "end-usage=cbr buffer-size=600 buffer-initial-size=400 buffer-optimal-size=500 ! "
```
vignesh-anand, line 623: Nit (duplication): this CBR buffer fragment is identical to the main encoder's above — change one and you have to remember the other. It's a string-concat pipeline so there's no clean constant to extract and it's only two sites, so low value to factor out; just flagging that the two must stay in sync.

webapp/js/teleop/profilingPanel.js
```diff
@@ -0,0 +1,240 @@
...
+      const dEmitted = now.jbEmitted - prev.jbEmitted;
+      const dDelay = now.jbDelay - prev.jbDelay;
+      jitterBuffer.value.textContent = dEmitted > 0 ? fmt((dDelay / dEmitted) * 1000, 0) : "—";
+
+      if (dt > 0) bitrate.value.textContent = fmt(((now.bytes - prev.bytes) * 8) / dt / 1000, 0);
+
+      const dRecv = now.packets - prev.packets;
+      const dLost = now.lost - prev.lost;
+      const denom = dRecv + dLost;
+      loss.value.textContent = denom > 0 ? fmt((dLost / denom) * 100, 1) : "0.0";
```
vignesh-anand, line 180: Display nit: when the stream stalls mid-session no new packets arrive, so `denom = dRecv + dLost === 0` and this prints `"0.0"` — a reassuring 0% loss exactly when the link is dead. Note the inconsistency: on the first poll loss honestly stays `"—"` (prev is null), but a later stall shows `"0.0"`. Using `"—"` for the `denom === 0` case would be consistent and wouldn't lie during an outage.

webapp/js/teleop/profilingPanel.js
```diff
@@ -0,0 +1,240 @@
...
+
+  toggleBtn.addEventListener("click", toggle);
+
+  /** @param {KeyboardEvent} e */
+  function onKeyDown(e) {
+    if (e.code !== TOGGLE_KEY || e.repeat || e.metaKey || e.ctrlKey || e.altKey) return;
+    const el = document.activeElement;
+    if (el instanceof HTMLElement && (el.tagName === "INPUT" || el.tagName === "TEXTAREA" || el.isContentEditable)) {
+      return;
+    }
```
vignesh-anand, line 226: Reuse: this `INPUT`/`TEXTAREA`/`isContentEditable` focus guard is a verbatim copy of `isTypingContext()` in `keyboardDrive.js` (same `teleop/` dir). It's module-private there; exporting and reusing it keeps the "is the user typing?" rule in one place so the two don't drift (e.g. if someone later adds `<select>` or `[contenteditable]` handling).

webapp/js/teleop/profilingPanel.js
```diff
@@ -0,0 +1,240 @@
...
+    value.textContent = "—";
+    const unitEl = document.createElement("span");
+    unitEl.className = "profiling-unit";
+    unitEl.textContent = unit;
+    const row = document.createElement("div");
+    row.className = "profiling-row";
+    row.append(value, unitEl);
+    el.append(label, row);
+    return { el, value };
+  }
```
vignesh-anand, line 104: Mild reuse: `metric()` is structurally the same label+value cell builder as `item()` in `telemetry.js` (`microlabel` label + `*-value mono` value, returns `{ el, value }`). `metric` adds a unit span and a headline flag, so a shared helper would need a small param — judgment call whether it's worth factoring out, just flagging the overlap.

webapp/js/teleop/profilingPanel.js
```diff
@@ -0,0 +1,240 @@
...
+
+    report.forEach((/** @type {any} */ s) => {
+      if (s.type === "inbound-rtp" && s.kind === "video") vid = s;
+      else if (s.type === "transport" && s.selectedCandidatePairId) selectedPairId = s.selectedCandidatePairId;
+    });
+    report.forEach((/** @type {any} */ s) => {
+      if (s.type !== "candidate-pair") return;
+      if (s.id === selectedPairId) selectedPair = s;
+      else if (!selectedPair && (s.nominated || s.selected)) selectedPair = s;
+    });
```
vignesh-anand, line 144: Simplification: the stats map is walked twice. It can be a single sweep that collects `vid`, `selectedPairId`, and a fallback nominated/selected pair, then resolves the pair by id afterward. 1 Hz so no perf cost — purely removing a duplicated loop.

PR #459, opened 2026-06-24
fix: restore ROS node exec bits + CI symlink-install coverage

karmanyaahm:
Split out of #452 ("make sim work") as the one cleanly-independent change — touches only `ci/*` and `ros2_ws/*`, no overlap with the rest of the sim rework.

## What
Six Python nodes declared via `install(PROGRAMS)` were tracked `0644`. A `--symlink-install` build (used by the local sim launcher) symlinks the install executable straight at the non-executable source, so `ros2 launch` reports "executable not found on the libexec directory" and the node never starts. Plain colcon (CI, prebuilt image, on-device) masks this by installing `0755` copies, so it only surfaced on local sim rebuilds.

## Commits
- `fix: restore exec bit on ROS install(PROGRAMS) node scripts` — restore +x to the six scripts, add the missing shebang to `odom_tf_broadcaster.py`
- `ci: run launch tests under both copy and --symlink-install builds` — add a second pass so this class of bug fails CI instead of only surfacing locally
- `ci: keep local colcon outputs out of the base image build context` — stop git-ignored local `build/install/log` from leaking into the docker context
- `ci: pin setuptools<80 for the symlink-install test pass` — `setup.py develop --editable` was removed in setuptools 80; the CI image ships 82


ci/run_integration_tests.sh
```diff
@@ -22,6 +22,21 @@ cleanup() {
...
 trap cleanup EXIT
 
+run_ros_integration_tests() {
+  # Launch-based tests: these actually start nodes (e.g. brain_client_node.py),
+  # so a node whose installed executable is missing or non-runnable fails here.
+  ros2 run rmw_zenoh_cpp rmw_zenohd &
+  ROUTER_PID=$!
+  sleep 2
+  colcon test --packages-select brain_client \
+    --ctest-args -R "test_pose_image|test_fake_cloud_loop" \
```
vignesh-anand, line 33: This regression guard does not actually exercise any of the five scripts this PR fixes. `test_pose_image` and `test_fake_cloud_loop` only launch `brain_client_node.py` and `skills_server.py` — both in `brain_client`, and both already `0755` on `main`. The five files repaired in this PR live in `manipulation`/`mars_arm`/`mars_bringup`/`mars_cam`/`mars_sim_bringup` and are not started by these tests, so if any of them loses its exec bit again this `--symlink-install` pass stays green.

The pass genuinely protects the `brain_client` nodes going forward, but not the sim/mars nodes that motivated the fix — which undercuts the "fails CI instead of only surfacing locally" rationale for exactly these files.

ci/run_integration_tests.sh
```diff
@@ -44,11 +59,34 @@ PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q \
...
+# image is older, which is why symlink builds work there but not here). Pin an
+# older setuptools for this build pass only — it's build-time tooling, so already
+# installed packages and Pass 1 are unaffected.
+#
+# This downgrades setuptools globally in the container, but it's the last thing
+# the script does: this pass is the final step and nothing runs in this layer
+# afterward, so the global mutation has no consumer to leak into.
+python3 -m pip install --quiet 'setuptools<80'
+rm -rf build install log
+colcon build --symlink-install \
```
vignesh-anand, line 125: Worth weighing whether a full clean `--symlink-install` rebuild + relaunch is the right depth for guarding this specific bug class. A static check — for every file listed in an `install(PROGRAMS)` block, assert it is executable and starts with `#!` — would cover all 19 such scripts (not just the handful a launch test happens to start), run in seconds, and target the exact failure mode.

The heavy pass does catch broader symlink-vs-copy divergences, so this could complement rather than replace it. But the cheap lint is what would reliably catch a future dropped exec bit on *any* node, including the four sim/mars scripts this PR fixes.

ros2_ws/src/mars_bot/mars_sim_bringup/mars_sim_bringup/odom_tf_broadcaster.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
vignesh-anand, line 1: Minor (description, not code): the PR body says "six Python nodes" / "the six scripts", but the diff fixes five — `recorder_benchmark.py`, `pid_hot_reload.py`, `rosbag.py`, `stereo_calibrator.py`, and this file. Every `install(PROGRAMS)` script in the tree is `0755` on this branch, so the fix is complete; the count in the body is just off by one.

ci/run_integration_tests.sh
```diff
@@ -22,6 +22,21 @@ cleanup() {
...
+run_ros_integration_tests() {
+  # Launch-based tests: these actually start nodes (e.g. brain_client_node.py),
+  # so a node whose installed executable is missing or non-runnable fails here.
+  ros2 run rmw_zenoh_cpp rmw_zenohd &
+  ROUTER_PID=$!
+  sleep 2
+  colcon test --packages-select brain_client \
+    --ctest-args -R "test_pose_image|test_fake_cloud_loop" \
+    --event-handlers console_direct+
+  colcon test-result --verbose
```
vignesh-anand, line 35: The whole point of this PR is "fail CI when a node cannot launch," but the red/green signal is load-bearing on this one line. `colcon test` returns `0` even when tests fail (long-standing colcon behavior), so a launch failure only fails the script via `colcon test-result --verbose` returning non-zero under `set -e`. That works today, but it is fragile: if anyone later trims or reorders this line, every launch failure silently goes green.

Passing `--return-code-on-test-failure` to the `colcon test` invocation above (or an explicit check on its result) would make the failure signal direct rather than incidental.

ci/run_integration_tests.sh
```diff
@@ -44,11 +59,34 @@ PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q \
...
+rm -rf build install log
+colcon build --symlink-install \
+  --parallel-workers "$(( $(nproc) < 4 ? $(nproc) : 4 ))" --cmake-args \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+  -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+# The copy-install dirs sourced above were just rm -rf'd; clear the overlay so
+# setup.bash rebuilds the chain from scratch instead of leaving stale, duplicated
+# AMENT_PREFIX_PATH entries that manifest-walking tools would process twice.
+unset AMENT_PREFIX_PATH CMAKE_PREFIX_PATH
```
vignesh-anand: Minor: this `unset` runs *after* the `colcon build` above, so at build time `AMENT_PREFIX_PATH` still contains the copy-install dirs that were just `rm -rf`'d (sourced earlier at the top of the script). colcon tolerates nonexistent prefixes so this is harmless in practice, but moving the `unset` to before the `colcon build` would keep the build env clean too, not just the subsequent `source`.

ci/run_integration_tests.sh
```diff
@@ -44,11 +59,34 @@ PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -m pytest -q \
...
+# colcon --symlink-install installs Python entry points via `setup.py develop
+# --editable`, which setuptools >= 80 removed (this image ships 82; the local sim
+# image is older, which is why symlink builds work there but not here). Pin an
+# older setuptools for this build pass only — it's build-time tooling, so already
+# installed packages and Pass 1 are unaffected.
+#
+# This downgrades setuptools globally in the container, but it's the last thing
+# the script does: this pass is the final step and nothing runs in this layer
+# afterward, so the global mutation has no consumer to leak into.
+python3 -m pip install --quiet 'setuptools<80'
```
vignesh-anand, line 123: This pin is justified — the 4 `ament_python` cloud packages (`innate_logger`, `innate_uninavid`, `innate_training_node`, `clients`) genuinely need `setup.py develop` to work under `--symlink-install`. But it adds a **third, divergent expression** of the same setuptools constraint, and the load-bearing one is owned by the wrong concern:

| Where | Pin |
|---|---|
| `sim/Dockerfile` (Linux sim container, runs `colcon --symlink-install`) | `setuptools==59.6.0` |
| `sim/requirements.macos.txt` (macOS host) | `setuptools==80.9.0` |
| here (this PR) | `setuptools<80` |
| `ros2_ws/pip-requirements.txt` | none |

Two things worth a tracking note:

1. **The constraint that makes local sim symlink-builds work is owned by torch, not colcon.** `sim/Dockerfile` pins `59.6.0` explicitly to "re-pin packaging/setuptools after torch resolves its deps" — it just *happens* to be `<80`. A future torch bump that needs setuptools >= 80 would raise that pin and silently break `scripts/validate_sim_ros_install.zsh` (`colcon build --symlink-install`), and this CI pass would not catch it because the two derive the constraint independently.

2. **The constraints have already diverged** (`==59.6.0` vs `==80.9.0` vs `<80`), with nothing referencing the colcon-`setup.py develop` reason behind any of them. The durable fix is upstream — colcon-ros using PEP 660 editable installs (`pip install -e`) instead of the removed `setup.py develop` — after which none of these pins would be load-bearing for symlink installs.

ci/Dockerfile.test-base.dockerignore
```diff
@@ -6,6 +6,14 @@
 
 !ros2_ws/
 !ros2_ws/**
+
+# ...but never local colcon outputs. Their CMakeCache.txt embeds absolute host
+# paths, so copying them in breaks the in-container build ("CMakeCache directory
+# is different"). CI checkouts have no build/; local dev does.
+ros2_ws/build/
```
vignesh-anand, line 13: The fix here is correct and correctly ordered (these exclusions come after `!ros2_ws/**`, so last-match-wins drops `build/install/log`). One symmetric gap worth a follow-up: the sibling `ci/Dockerfile.test.dockerignore` also has `!ros2_ws/**` but no such exclusion.

It is not a correctness bug there — `Dockerfile.test` only does `COPY ros2_ws/src` (line 53), so it never copies `build/install/log` and the CMakeCache collision cannot happen. But `!ros2_ws/**` still re-includes those local output dirs into the *build context* sent to the daemon, so any local build of `Dockerfile.test` on a dev machine with a populated `ros2_ws/build` (often multi-GB) uploads all of it for nothing.

The same three lines (`ros2_ws/build/`, `ros2_ws/install/`, `ros2_ws/log/`) in `Dockerfile.test.dockerignore` would be a free context-size win — pure efficiency, no behavior change.

ci/run_integration_tests.sh
karmanyaahm: ya but it catches any other weird bugs

ci/run_integration_tests.sh
karmanyaahm: done

ci/run_integration_tests.sh
karmanyaahm: tracked

sim/frontend/docker-entrypoint.sh
```diff
@@ -0,0 +1,69 @@
...
+      // generation and leave the app with no runtime config; keep the default.
+      console.error(`[entrypoint] ignoring invalid ${env}: ${e.message}`);
+    }
+  };
+  set("simBaseUrl",     "SIM_BASE_URL");
+  set("wsBaseUrl",      "WS_BASE_URL");
+  set("robotWsUrl",     "ROBOT_WS_URL");
+  set("directRobot",    "DIRECT_ROBOT",     v => v === "true");
+  set("pinnedSkills",   "PINNED_SKILLS",    JSON.parse);
+  fs.writeFileSync(process.argv[1], JSON.stringify(out));
```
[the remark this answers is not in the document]
karmanyaahm: marked intentional

karmanyaahm:
ok merging after checks

PR #460, opened 2026-06-24
feat(sim): containerize frontend and migrate control plane to rosbridge

karmanyaahm:
Middle of the #452 stack. Stacked on #459 (exec bits/CI). This is the **architectural rework half**: stand up the new frontend container and move the whole control plane off FastAPI HTTP routes onto rosbridge. The runtime-reliability + sensor-streaming half follows in #452.

## Containerize the web frontend (runtime config)
- `65f1d89f` move the React app into its own Caddy Docker image, reading endpoint URLs from `/config.json` at runtime instead of build-time `VITE_*`
- `fbaa20dc` docs · `033b1deb` make `public/config.json` the single source of config defaults · `cca4248e` replace top-level `innate setup` with `innate sim setup`

## Launcher tooling + deps
- `4f4c4f33` `innate sim sh` · `4f4d9410` `innate sim clean` · `673edb71` download skill assets on `sim up` · `ff313338` requirements.in + compiled lock · `bc0c6338` apt dep bump

## Web UI controls
- `f274e959` first-person/arm/chase camera switcher · `84f97a06` manual teleop drive pad · `00412e20`+`b98e71a4` request transient_local durability for the skill catalog

## HTTP → rosbridge migration
- `d3d000db` reset_position · `fe1725df` set_environment → rosbridge services
- `8511d410` drop legacy `/reset_robot` · `e4e24b1e` brain status from rosbridge · `ca312957` drop the HTTP routes · `7d000707` always use rosbridge for brain/skill control · `c3d9752c` move chat to rosbridge, delete `chat_api.py` · `c4aadab2` drop dead service_key middleware · `7d7a41bf` lint


sim/frontend/src/services/rosbridgeService.ts
```diff
@@ -573,6 +620,59 @@ function getAvailableSkillsFromTopicDirect(
   });
 }
 
+// Persistent subscription to the brain's backend status. Replaces the old
+// /stack_metrics HTTP poll. Returns an unsubscribe function.
+export function subscribeBrainBackendStatus(
+  wsUrl: string,
+  onStatus: (status: BrainBackendStatus) => void,
+): () => void {
+  const ws = new WebSocket(wsUrl);
```
vignesh-anand, line 629: **Brain-status subscription has no error/close handling and never reconnects**

This socket is created once (the consuming `useEffect` in `App.tsx` has deps `[robotWsUrl]`, which never changes) and only wires up `onopen` + `onmessage` — there's no `onerror`, no `onclose`, and no reconnect. The old `/stack_metrics` 1 Hz HTTP poll it replaces was inherently self-healing (every tick a fresh stateless request); a persistent socket is not.

This is worse than "freezes on drop" because the live subscription is the **sole** source of `brainBackendStatus` updates: `fetchAgents` only sets it `if (data.brain_backend_status)`, but `getAvailableAgentsDirect` no longer returns that field (the old HTTP `/available_agents` did). So:

1. **Initial connect fails** — on `sim up` the frontend can load before rosbridge/sim is ready; the first connection fails, is never retried, and the brain-status indicator stays dead for the whole session until a manual reload, even after the brain comes up fine.
2. **Socket drops mid-session** (rosbridge/sim restart, network blip) — the indicator freezes on its last value forever, which is most misleading exactly when the brain is reconnecting.

Controls still work (each service call opens its own throwaway socket), so this is a telemetry/health-display degradation, not a control bug — but it makes the dashboard untrustworthy.

Note `Chat.tsx` in this same PR already has the right pattern (5 attempts, backoff, reconnect on non-clean close). Suggest giving this subscription an `onclose`/`onerror` that schedules a backoff reconnect + re-subscribe (guarded by the existing `disposed` flag), and optionally re-emitting `connected: false` on close so the indicator shows disconnected rather than freezing.

sim/frontend/docker-entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+
+mkdir -p "$WEB_DIR" "$STATUS_DIR" "$CONFIG_DIR"
+
+# 1. Runtime config.json from env (browser-facing URLs).
+cat > "$CONFIG_DIR/config.json" <<EOF
+{
+  "simBaseUrl": "${SIM_BASE_URL:-http://localhost:8000}",
+  "wsBaseUrl": "${WS_BASE_URL:-ws://localhost:8000}",
+  "robotWsUrl": "${ROBOT_WS_URL:-ws://localhost:9090}",
+  "directRobot": ${DIRECT_ROBOT:-false},
```
vignesh-anand: **A single malformed env var silently reverts the entire runtime config to localhost, with no diagnostics**

Two issues compound here.

**1. `directRobot` is unquoted, so the env value must be valid JSON literally.** `"directRobot": ${DIRECT_ROBOT:-false}` only works if `DIRECT_ROBOT` is exactly `true`/`false`. Anything else a user might naturally type in `.env` — `True`, `TRUE`, `1`, `yes` — produces `"directRobot": True,`, which is invalid JSON. Because `config.json` is one document parsed atomically, that doesn't just break `directRobot`; it makes the **whole file** unparseable, taking your otherwise-correct `SIM_BASE_URL`/`WS_BASE_URL` down with it. (The string fields are quoted but unescaped, so the same string-interpolation pattern breaks on a stray `"`/backslash too.)

**2. `loadConfig` swallows the failure and silently keeps localhost defaults.** In `config.ts`, `await response.json()` throws on malformed JSON and the `catch {}` discards it — `appConfig` stays at its compiled defaults (`localhost:8000` / `localhost:9090`), nothing logged. The app then renders fine and each consumer independently fails to reach localhost: brain-status never loads, agents time out, video feed is broken, chat won't connect — a scattering of "connectivity" symptoms, none pointing back at the config file. You'd debug five separate network issues before suspecting one bad env var.

(Related sneaky variant: valid JSON but wrong *type*, e.g. `"directRobot": "false"` as a string — `Object.assign` trusts it unvalidated, and a non-empty string is truthy in JS, so the app silently flips into direct-robot mode.)

The happy path is fine (dev `public/config.json` is hand-written; `config.py` defaults `DIRECT_ROBOT` to lowercase `false`) — the exposure is entirely on user override, which is exactly when booleans get fat-fingered.

**Fix, two layers:**
- Generate guaranteed-valid JSON instead of string-splicing. Node is already in the image, so let it serialize (also normalizes `True`/`1`/`yes` → real `true` and escapes strings):
  ```sh
  node -e 'const b=v=>/^(1|true|yes)$/i.test(v||"");
    process.stdout.write(JSON.stringify({
      simBaseUrl: process.env.SIM_BASE_URL || "http://localhost:8000",
      wsBaseUrl:  process.env.WS_BASE_URL  || "ws://localhost:8000",
      robotWsUrl: process.env.ROBOT_WS_URL || "ws://localhost:9090",
      directRobot: b(process.env.DIRECT_ROBOT),
      cartesiaApiKey: process.env.CARTESIA_API_KEY || "",
    }, null, 2));' > "$CONFIG_DIR/config.json"
  ```
- Make `loadConfig`'s fallback visible — `console.warn("config.json failed to load; using localhost defaults", err)` on the `!response.ok` and `catch` paths, so silently defaulting to localhost isn't indistinguishable from success.

sim/frontend/docker-entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+<!doctype html><meta charset="utf-8"><title>Building…</title>
+<body style="font-family:monospace;padding:2rem">Building the web app… this page will refresh.
+<script>setTimeout(function(){location.reload()},2000)</script></body>
+EOF
+
+# 3. Build in the background; swap dist into the web root on success, error page on failure.
+#    Output is teed to build.log (served at /build.log) and to container stdout.
+(
+  echo "[entrypoint] starting web build"
+  if yarn build 2>&1 | tee "$BUILD_LOG"; then
```
vignesh-anand, line 50: **`docker-entrypoint.sh` rebuilds the frontend on every container start — always redundant**

The entrypoint runs `yarn build` from scratch on every container start and overwrites the image's baked `dist/`. This is redundant in every path:

- The launcher always starts the frontend with `docker compose up -d --build` (`runtime.py` `ensure_frontend_container`), so the image's `COPY . . && RUN yarn build` has already produced a current `dist/` (layer caching rebuilds it only when `src/` changed).
- Even without `--build` (manual `docker compose up` / `docker restart`), there's **no bind mount** on the `frontend` service — source enters only via `COPY` at image-build time. So `/app` is a frozen snapshot, and the baked `dist/` already matches it. Rebuilding reproduces identical output.

The runtime build takes no runtime input either (config is injected separately via `config.json`), so there's nothing it can compute that the image build didn't. Yet it sits on the blocking path: `wait_for_frontend_ready` polls `/build-status.json` (up to the 300s cap) until this redundant build finishes, so every `sim up` eats a full `yarn build` (~tens of seconds) before the UI is ready.

Proposed fix: drop `yarn build` from the entrypoint and serve the baked output — point Caddy's root at `/app/dist` (or `cp` it once) and write `{"state":"ready"}` immediately; keep the `config.json` write and the `/build-status.json` + `/build.log` endpoints so the launcher contract is unchanged. `wait_for_frontend_ready` then returns ~instantly. Optional follow-on: split into a `node` build stage + `caddy` serve stage so the runtime image sheds the Node toolchain / `node_modules`.

sim/frontend/src/components/TeleopDrive.tsx
```diff
@@ -0,0 +1,204 @@
...
+      if (!dir) return;
+      e.preventDefault();
+      addDir(dir);
+    };
+    const onKeyUp = (e: KeyboardEvent) => {
+      const dir = KEY_MAP[e.key.toLowerCase()];
+      if (dir) removeDir(dir);
+    };
+    window.addEventListener("keydown", onKeyDown);
+    window.addEventListener("keyup", onKeyUp);
```
vignesh-anand, line 164: **`TeleopDrive` keyboard drive doesn't stop when the window loses focus**

The teleop pad is meant to be hold-to-drive / release-to-stop: `keydown` adds a direction to `activeRef` and starts a 10 Hz `/cmd_vel` publish loop; `keyup` removes it and, when empty, publishes a single zero-twist stop.

The problem: **`keyup` is only delivered to the focused window.** If you hold W and then alt-tab / click another window / focus a field elsewhere, the page stops receiving key events, the `keyup` never arrives, `"forward"` stays in `activeRef`, and the 10 Hz loop keeps publishing forward velocity. Nothing listens for `blur` / `visibilitychange` to flush that state.

The sim's implicit per-tick deadman doesn't help here — it stops the robot when `/cmd_vel` *stops* arriving, but this is the opposite case: commands keep flowing. So the sim robot keeps driving. If the window is merely unfocused-but-visible the loop runs at full 10 Hz (full-speed drive); if the tab is fully hidden the loop is throttled to ~1/min (slow creep). Net effect: you alt-tab mid-drive and come back to find the robot drove itself into a wall / out of scene, corrupting whatever test/demo/episode was in progress. Not a safety issue in sim, but it silently breaks the "release = stop" contract.

(The mouse/touch path is fine — `onPointerUp`/`onPointerLeave`/`onPointerCancel` already cover it. Only the keyboard path depends on the focus-bound `keyup`.)

Suggested fix: add a `window` `blur` + `document` `visibilitychange` listener that clears `activeRef` and calls `stopLoop()` (which already emits the zero-twist), so losing focus halts the robot.

sim/frontend/docker-entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+mkdir -p "$WEB_DIR" "$STATUS_DIR" "$CONFIG_DIR"
+
+# 1. Runtime config.json from env (browser-facing URLs).
+cat > "$CONFIG_DIR/config.json" <<EOF
+{
+  "simBaseUrl": "${SIM_BASE_URL:-http://localhost:8000}",
+  "wsBaseUrl": "${WS_BASE_URL:-ws://localhost:8000}",
+  "robotWsUrl": "${ROBOT_WS_URL:-ws://localhost:9090}",
+  "directRobot": ${DIRECT_ROBOT:-false},
+  "cartesiaApiKey": "${CARTESIA_API_KEY:-}"
```
vignesh-anand: **Security (pre-existing, but worth fixing here): the Cartesia API key is shipped to the client.**

Not introduced by this PR — the billable Cartesia key has always been client-exposed (previously baked into the JS bundle via `VITE_CARTESIA_API_KEY`, and it's used client-side: `new CartesiaClient({ apiKey: appConfig.cartesiaApiKey })` in `Chat.tsx`). So this is calling out an existing vulnerability, not a regression.

That said, it should be fixed, and this PR is the natural place since it's reworking exactly this config path. Here the key gets written into `/srv/config/config.json`, which Caddy serves at a predictable, unauthenticated URL (`handle /config.json { file_server }` — no auth/allowlist), so anyone who can reach the frontend can `curl http://<host>:<FRONTEND_PORT>/config.json` and read it in clean JSON.

Mitigating: the committed default is `"cartesiaApiKey": ""` (no secret in the repo), so this only bites a deployment that sets `CARTESIA_API_KEY`, and the risk scales with how network-reachable `FRONTEND_PORT` is.

Proper fix is the standard pattern for using a paid API from the browser: don't ship the long-lived key — have a backend mint a short-lived Cartesia access token (Cartesia supports ephemeral tokens for exactly this client-side case) and hand the frontend that, or proxy Cartesia through a backend that holds the key server-side. At minimum, treat anything in `config.json` as fully public and don't expose `FRONTEND_PORT` beyond trusted networks when a real key is set.

sim/frontend/src/services/rosbridgeService.ts
```diff
@@ -224,6 +224,49 @@ export function publishRosbridgeTopic(
   });
```
vignesh-anand, line 224: Flagging two robustness gaps in the pre-existing direct-robot path, since this PR removes the HTTP fallback and makes this code the *only* control path. (Not changes in this diff — `publishRosbridgeTopic` here, and `startRosbridgeAction` below — but newly load-bearing, so worth a look.)

**(a) `publishRosbridgeTopic` has no overall timeout — it can hang forever.** Unlike `callRosbridgeService` (which arms a `setTimeout` that rejects after 10s), this function only settles via `onopen` (send + linger) or `onerror`. If the socket gets stuck in `CONNECTING` without ever firing `onerror`/`onclose` (a network black hole rather than a refused connection), the returned promise never settles and the `await`ing caller hangs indefinitely — `setDirectiveDirect`, `setActiveSkillsDirect`, `resetBrainDirect`, `setBrainBackendConfigDirect`, `publishManualSkillEventDirect`. e.g. `handleSetDirective` would hang and never run its `setBrainActive(true)` follow-up. Low likelihood against a local rosbridge (it connects or refuses), but real for direct-robot over a flaky network — which is exactly where there's no HTTP fallback anymore. Cheap fix: give it the same timeout guard `callRosbridgeService` already has.

**(b) `startRosbridgeAction` (~L432 / ~L472) — two inherited edge cases (NOT regressions; the deleted server-side `execute_rosbridge_action` behaved the same).** Just noting since this is now the sole skill-execution path:
- Cancel uses `activeGoalId`, which is only set from an incoming `action_feedback` message — even though the client generated and sent its own `goalId` in `send_action_goal`. An action that never emits feedback therefore can't be cancelled. Harmless for `ExecuteSkill` (it does feed back — the old server cancel relied on the same mechanism), but needlessly fragile.
- The 300s client timeout (`finish()` → `actionWs.close()`) closes the socket **without sending a cancel**, so a skill running >5min keeps executing on the robot while the UI reports a timeout/failure — an orphaned goal, UI and robot disagree.

Only (a) is worth actually fixing IMO; (b) are pre-existing and inherited unchanged.

sim/launcher/runtime.py
```diff
@@ -1167,6 +1188,39 @@ def ensure_sim_data(config: dict[str, object], *, allow_fetch: bool) -> None:
...
+                continue
+            log(f"Downloading skill asset {dest.relative_to(workspace)}...")
+            tmp = meta_file.parent / f"{fname}.tmp"
+            try:
+                with urlopen(Request(url), timeout=300) as resp, open(tmp, "wb") as out:
+                    shutil.copyfileobj(resp, out)
+                tmp.replace(dest)
+            except (URLError, OSError) as exc:
+                tmp.unlink(missing_ok=True)
+                raise StackError(f"Failed to download skill asset {dest}: {exc}") from exc
```
vignesh-anand, line 1217: **`ensure_skill_assets` couples `sim up` to external download success.** This runs unconditionally in `cmd_up`, and any single failed download raises `StackError`, aborting the entire `sim up`. So a transient 503 / DNS hiccup / slow mirror on one skill's asset blocks the whole sim from starting — even though the sim would otherwise run fine (that skill just wouldn't have its file). Since the step is idempotent and cached, a retry later would succeed, which makes "abort everything now" a brittle coupling: sim startup now has a hard dependency on every skill download URL being reachable at that moment.

Suggest failing soft per-asset — log a warning and continue (optionally collect failures and summarize at the end) so a flaky download degrades one skill rather than taking down the whole runtime.

sim/launcher/main.py
```diff
@@ -175,6 +188,22 @@ def cmd_down(config: dict[str, object]) -> None:
     log("Innate sim runtime is down.")
 
 
+def cmd_clean(config: dict[str, object], *, delete_data: bool = False) -> None:
+    clean_runtime(config, delete_data=delete_data)
```
vignesh-anand, line 239: **`sim clean` is destructive with no confirmation.** `clean_runtime` deletes the sim `.venv` (and with `--data`, the multi-GB re-downloadable ReplicaCAD datasets + asset pack) immediately, no `y/n`. It's documented as a full reset, but a fat-finger of `sim clean` instead of `sim down` forces a complete re-`setup` (minutes of rebuild + re-download). For a destructive command — especially the `--data` variant — a confirmation prompt (with a `--yes`/`-y` to skip for scripting) is cheap insurance against accidental data loss.

sim/frontend/src/services/rosbridgeService.ts
karmanyaahm: not gonna worry about app details but will feed this to the next app's agent

sim/frontend/docker-entrypoint.sh
karmanyaahm: this was fixed in the child pr oops

sim/frontend/docker-entrypoint.sh
karmanyaahm: WTF TBD

sim/frontend/src/services/rosbridgeService.ts
karmanyaahm: will check against new app later

sim/launcher/main.py
karmanyaahm: nah its not deleting any useful data, only redownloadable assets

karmanyaahm: although i guess if someone has terrible internet hmmm

karmanyaahm: fixed

sim/frontend/docker-entrypoint.sh
karmanyaahm: just break it idc about this web app, we should render audio on the robot side like existing and perhaps pipe it thru

sim/frontend/docker-entrypoint.sh
karmanyaahm: we are discarding the app soon so this will be rendered irrelevant

sim/launcher/runtime.py
karmanyaahm: added a --offline that is suggested after a failure on these cases

PR #461, opened 2026-06-24
Migrate ~/skills and ~/agents into workspace/custom_* (fix listed-but-broken datasets)

DavidDobas:
## Problem
The brain scans `~/skills` and `~/agents` **in place** for backwards compatibility (`get_skill_directories()` in `brain_client/common/script_paths.py`) and reports their **absolute paths** on `/brain/available_skills`. So a dataset recorded under `~/skills/<task>` (older robots, ≤ 0.5.x) appears in the web app's Datasets list and in the controller app.

But the dataset **media server** (`webapp/proxy/https_server.py`, fenced to `SKILLS_ROOT = workspace/custom_skills`) and the **`dataset_encoder`** only ever serve/scan `workspace/custom_skills`. So a `~/skills` dataset is **listed-but-broken**: no thumbnails, episodes won't play, `/episode/joints` 404s, and it can't be encoded or downloaded for training.

The reconciliation these home dirs were supposed to get **never existed** — `migrate_legacy_home_directories()` is referenced only in comments (`migrate_user_data.sh`, `post_update.sh`) and is not implemented anywhere.

(`$INNATE_OS_ROOT/skills` was already migrated by `migrate_user_data.sh`; only the true home dirs were missed.)

## Fix
Fold `~/skills` → `workspace/custom_skills` and `~/agents` → `workspace/custom_agents` in `run_user_data_migrations`, so the brain scan, media server, and encoder all agree on one location. New `_migrate_home_dir_into_workspace` reuses the existing **never-clobber / empty-only / chown-aware** contract; `_mig_home_dir` resolves the login user's home (via `MIGRATE_CHOWN_USER`) since `post_update.sh` runs as root. Symlinked `~/skills` is left untouched. Stale `migrate_legacy_home_directories` comments removed.

This fixes both the web app and the controller app (both consume the same media server) for upgraded robots.

## Testing
Smoke-tested locally:
- happy path: `~/skills/<task>` moves into `workspace/custom_skills/<task>`, empty `~/skills` removed;
- collision: a pre-existing destination is preserved (source kept in place for manual reconciliation);
- `~/agents` migrates to `custom_agents`;
- `bash -n` clean on both scripts.

## Release note
Should ship in **0.6.0** (the controller app now pins `MIN_ROBOT_VERSION = 0.6.0`), so upgraded robots have their `~/skills` datasets served correctly.

### DIFF
--- a/scripts/update/migrate_user_data.sh
+++ b/scripts/update/migrate_user_data.sh
@@ -18,10 +18,14 @@
 #   <repo>/maps/*                          -> <repo>/data/maps/
 #   <repo>/.last_mode                      -> <repo>/data/.last_mode
 #   <repo>/.last_map                       -> <repo>/data/.last_map
+#   ~/agents/*                             -> <repo>/workspace/custom_agents/
+#   ~/skills/*                             -> <repo>/workspace/custom_skills/
 #
-# Home-directory skills/agents (~/skills, ~/agents) are NOT handled here; they
-# are migrated at brain startup by
-# brain_client.script_paths.migrate_legacy_home_directories().
+# The home-directory skills/agents (~/skills, ~/agents) are folded in here too.
+# The brain still scans them in place for backwards compatibility, but the
+# dataset media server and encoder only serve workspace/custom_skills, so a
+# ~/skills dataset would otherwise list but be unplayable (no thumbnails/video/
+# joints). Moving them into the workspace keeps both pipelines consistent.
 #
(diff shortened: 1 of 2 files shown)


DavidDobas:
Scrapping this approach — we don't want to move users' files around. Replacing with a change that makes the dataset media server + encoder read from the legacy skill locations (~/skills etc.), matching what the brain's skill scan already does. New PR to follow.

PR #464, opened 2026-06-24
Warn loudly when launching a robot without a service key

theo-michel:
## Problem

Without `INNATE_SERVICE_KEY`, a robot still boots but can't reach the Innate cloud brain — AI behaviors, logging, and on-device training all fail. The only existing signal was a single throttled `warn` line in the brain_client log: easy to miss, and with no guidance on how to recover.

## Change

Surface a missing service key in three places, all pointing operators to Discord (`https://discord.gg/innate`) for a new key.

- **`scripts/print_runtime_env.py`** — add a `--has-service-key` flag that exits 0 if the key resolves to a non-empty value, 1 otherwise. Reuses the existing `/etc/innate.env` + repo `.env` merge so detection matches runtime precedence, and treats empty/quoted-empty as missing.
- **`scripts/launch_ros_in_tmux.sh`** — the robot-start chokepoint. Detects a missing key up front and prints a loud red banner as the **last** thing on screen, right after the "All systems launched" box.
- **`ros2_ws/.../transport/ws_manager.py`** — replace the single-line missing-key warnings with a multi-line banner logged **line-by-line at error level** (each line gets its own `[ERROR]` prefix, so it's greppable and hard to scroll past). Uses a **dedicated 30s throttle** so it's never suppressed by an unrelated invalid-config line. All three missing-key paths (`_trigger_connect`, `_configure_ws_client`, `_send`) route to it; `_send` only shows the key banner when the failure is specifically a missing token.

## How it looks

Launch banner (red):
```
  ════════════════════════════════════════════════════════════════════
   ⚠  NO SERVICE KEY FOUND
  ════════════════════════════════════════════════════════════════════
   This robot has no INNATE_SERVICE_KEY, so it cannot connect to the
   Innate cloud brain. AI behaviors, logging, and on-device training
   will not work until a service key is configured.

   To get a new service key, contact Innate on Discord:
       https://discord.gg/innate
  ════════════════════════════════════════════════════════════════════
```

brain_client log:
```
[brain_client]: [ERROR] ═══════════════════════════════════════════════════════════════════════
[brain_client]: [ERROR] ⚠  NO SERVICE KEY — ROBOT CANNOT REACH THE INNATE CLOUD BRAIN
[brain_client]: [ERROR] INNATE_SERVICE_KEY is missing. AI behaviors, logging, and on-device
[brain_client]: [ERROR] training will not work until a service key is configured.
[brain_client]: [ERROR] Get a new service key — contact Innate on Discord: https://discord.gg/innate
[brain_client]: [ERROR] ═══════════════════════════════════════════════════════════════════════
```

## Scope / caveats

- Detects **missing** (or empty) key only. A *present-but-rejected* key can't be known at launch without round-tripping the server; it still surfaces as the existing `backend_error` reconnect loop.
- Discord invite is the one already used in the repo READMEs.
- Changes live in a ROS node + launch script, exercised on a real robot start / `colcon build`. Verified locally: `ast.parse` + `ruff` clean on the node, `bash -n`/`zsh -n` clean on the script, `--has-service-key` tested across present/missing/empty/non-empty.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/transport/ws_manager.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/transport/ws_manager.py
@@ -125,9 +125,7 @@ class WebSocketManager:
             if self._ws_configured and not self._token_configured:
                 self._log_missing_service_key_once()
             else:
-                self._log_invalid_config_once(
-                    "Cannot send websocket message: hosted Innate agent is not configured."
-                )
+                self._log_invalid_config_once("Cannot send websocket message: hosted Innate agent is not configured.")
         elif self.ws_client.loop and self.ws_client.loop.is_running():
             try:
                 asyncio.run_coroutine_threadsafe(self.ws_client.send(message_in), self.ws_client.loop)
--- a/ros2_ws/src/brain/brain_client/brain_client/transport/ws_manager.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/transport/ws_manager.py
@@ -31,6 +31,17 @@ from brain_client.transport.messages import InternalMessage, InternalMessageType
 from brain_client.transport.ws_config import is_hosted_innate_uri, validate_token_for_uri, validate_ws_uri
 from brain_client.transport.ws_transport import WS_THREAD_STOP_JOIN_SECONDS, WSClient
 
+# Logged line-by-line at error level (one [ERROR] prefix per line) so a missing
+# service key is impossible to miss when scanning the brain_client logs.
+_MISSING_SERVICE_KEY_BANNER = (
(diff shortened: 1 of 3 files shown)


theo-michel:
Good catch on the ambiguous exit code. One note: checking strictly for `-eq 1` wouldn't fully fix it, since a Python crash also exits `1` and would still trigger the banner.

Addressed it by removing the overload entirely: `--has-service-key` now prints an explicit `present`/`missing` token and always exits `0`, and the launch script matches on that token. A crash produces no token, so it fails safe to "present" (no false banner) rather than a misleading no-key warning. Pushed in 49b6408e's follow-up.

karmanyaahm (approved):
cool

PR #470, opened 2026-06-25
webapp, sim, and WebRTC: per-client protocol, agent/skills surfaces, brain state broadcasts

karmanyaahm:
## Overview

What started as the webapp build/serve pipeline has grown into the full **webapp + sim + WebRTC protocol** release train. Base is `main`. Four areas, each usable standalone:

1. **WebRTC: per-client (`_id`) protocol** — robot (mars_cam), sim (aiortc), and webapp all speak the new lazy multi-peer protocol.
2. **Webapp: agent & skills surfaces** — Agent page, skills panel, chat, TTS playback, live agent state.
3. **Brain: broadcast state instead of client-side guessing** — `/brain/agent_status`, skill status on every trigger, shared started/stopped chat messages.
4. **Sim & launcher** — webapp served inside the sim, brain-backend auto-select, teleop fidelity (head pitch, arm status), setup/ops fixes.

Companion PR: innate-controller-app #226 (mobile app side of the same protocol + brain-state work).

## 1. WebRTC per-client protocol

**Robot (mars_cam)** — the monolithic streamer is split into `webrtc_signaling/encode/transport/stun.cpp`:
- **Lazy multi-peer backend**: one persistent VP8 encode pipeline per camera; one transport pipeline (webrtcbin) per `client_id`. No preemption — devices stream concurrently; encoders only run while some peer watches (`want` counts).
- **`_id` signaling**: `/webrtc/offer_id`, `/webrtc/answer_id`, `/webrtc/ice_in_id`, `/webrtc/ice_out_id`, all `{client_id, ...}` JSON envelopes; START carries `{client_id, source, audio, video[], renegotiate}`. A START without `renegotiate` flips the active camera/audio set on the existing peer — instant, no re-offer (this is what makes the reneg-free mic toggle work). Documented in `WEBRTC_CLIENT_ID_MIGRATION.md`. **Legacy bare topics are removed** — see “Before merge/ship”.
- **Audio negotiated up front** (opus m-line always present when a mic exists); toggling the mic is a plain START.
- **Local STUN Binding responder** on UDP 3478 (`XOR-MAPPED-ADDRESS`), so LAN clients learn a usable srflx candidate without a public round-trip; mDNS `.local` candidates are deferred briefly in favor of it.
- **Liveness**: RTCP-inactivity watchdog (10s, param-tunable), connect-timeout reaping, `/webrtc/active_streams` roster broadcast (0.5 Hz), per-peer keyframe forcing on connect.

**Sim (aiortc, `sim/src/webrtc/webrtc_server.py`)** — same protocol server-side: one `RTCPeerConnection` per `client_id`, three always-negotiated camera tracks (`first_person`/`arm_wrist`/`chase`) with lazy encoding, its own STUN responder (`SIM_STUN_PORT`), signaling-queue coalescing (a newer renegotiate START supersedes the stale session it replaces), a stale-peer reaper (connect timeout + disconnected grace), and build-time logging. Empty `video` releases the peer, mirroring the robot.

**Webapp (`webapp/js/webrtcSession.js`)** — per-page `client_id`, mid-based track routing, no-reneg camera switching for the PiP strip, robot-local STUN first with public-STUN fallback escalation, offer/media watchdogs (offer timeout sized to the sim’s non-trickle ICE gathering), and a multi-peer debug harness at `debug/webrtc.html`.

## 2. Webapp

- **Zero-build static app.** An esbuild/Lit pipeline was added early in this branch and deliberately removed again — the app ships as plain ES modules with `// @ts-check` + `tsconfig` for editor-time types only. (Node.js was dropped from the apt dependencies accordingly.)
- **Agent page** (`/agent/`): directive picker + Start/Stop, AI-thoughts stream (grouped, collapsible), active-skill display, chat composer; an “<agent> is running” indicator pill on every other page.
- **Skills panel** (teleop right dock): live roster from `/brain/available_skills`, per-skill parameter forms generated from input schemas, execution via the `/execute_skill` action with streamed feedback + cancel.
- **`rosClient` action support**: `sendActionGoal` multiplexed over the shared resilient socket — cancel targets the server-assigned `goal_id`, goals orphaned by a socket drop are cancelled on reconnect.
- **TTS playback**: plays `/tts/audio` (sim-only; see brain section). **Camera PiP strip** with live thumbnails + nav-map PiP. Dedicated Map page removed in favor of the PiP/dock.
- **Server**: `https_server.py` split into `https_server` + `media_routes` + `settings_routes`; serves the full app on both 443 (TLS) and 80/4080 (HTTP). The earlier sticky-https redirect was **removed** — HTTP is a first-class transport now (see “Open questions”).

## 3. Brain

- **`/brain/agent_status`** (latched + 3s heartbeat): `{brain_active, current_directive, active_skills}` published on every activate/deactivate/directive/skill-set change — clients no longer poll `get_available_directives` or guess from chat traffic to notice a stop/start made on another device.
- **Skill status for every trigger**: `skills_server` (the action server) broadcasts running/terminal status for all goals — agent-run, webapp, app — with a guaranteed terminal status even on internal exceptions; the runner’s duplicate local publishes are gone. One skill runs at a time (goals racing an in-flight skill are aborted promptly).
- **Shared started/stopped chat messages**: the brain emits `"<agent> started./stopped."` system messages on `/brain/chat_out` (and into history); the webapp/app local echoes were removed, so every client sees the same transcript.
- **TTS**: synthesized speech is published as base64 WAV on `/tts/audio` **in sim mode only** (no audio device — the browser is the speaker). The real robot plays through its own speaker and publishes nothing (no double voice).
- `available_skills` heartbeat so late rws subscribers get the roster.

## 4. Sim & launcher

- **Webapp inside the sim**: the OS container serves the webapp + `innate_console` (`console-webapp` tmux window); the separate `sim/frontend` React app is deleted. rosbridge is loopback-only (browser rides the same-origin `/ws` proxy).
- **Brain backend auto-select**: local cloud-agent (Gemini key) vs hosted (service key) chosen from configured keys; setup wizard prompts for missing keys (value-less `# KEY=` placeholders no longer masquerade as configured), toggles keys on backend switch, and can run the cloud-agent from source via a compose profile.
- **Teleop fidelity**: head pitch driven from `/mars/head/set_position` (±20°, republished with limits), `/mars/arm/status` published so the torque toggle works, `/cmd_vel` held between arrivals to stop motion stutter.
- **Ops**: fail-fast when `SIMULATOR_PORT` is taken (pre-flight check + the backend exits instead of “loading” into the timeout); scene-load heartbeat filters benign GL driver noise; TLS cert persisted across down/up; 80→4080 redirect; `innate` CLI on PATH in the container; log-tail flags.
- **apt dependency restructure**: split into `common` + `sim` + `hardware` overlays; third-party repo setup extracted to `scripts/update/setup_repos.sh` (shared by robot updater + sim image); NodeSource/Node.js removed entirely.

## Review status

Two review passes are addressed on the branch (path traversal, multi-peer aiortc, rejected-goal handling; then the mars_cam state-callback deadlock, skills cancel-handle loss, orphaned action goals, stuck-running status, sim peer lifecycle). Remaining known items, deliberately deferred: cleartext control-plane posture (Origin checks on `/ws` + `/settings` — decision pending, LAN rosbridge was already open), TTS overlap/multi-tab echo in sim, armPanel “Switch to HTTPS” port mapping, and a P3 polish list (docs refresh, param-type coercion, focus retention, debug-page HTML escaping, image-staleness hash inputs).

## ⚠️ Before merge/ship

- **Protocol break**: legacy `/webrtc/offer|answer|ice_in|ice_out` topics are gone. Deployed mobile apps speak the old protocol until innate-controller-app #226 ships — release ordering: app update (with its `MIN_ROBOT_VERSION` bump) must accompany the robot release that carries this.
- The robot keeps unprivileged 4443/4080 ports (pinned); 80/443 binding would need `cap_net_bind_service`.


sim/docker-compose.dev.yml
```diff
@@ -19,6 +19,12 @@ services:
       context: ..
       dockerfile: sim/Dockerfile
       target: dev-runtime
+      # Build with the host network namespace so `RUN apt-get` can reach DNS.
+      # On hosts using systemd-resolved (nameserver 127.0.0.53) the default
+      # bridge build network can't reach that host-loopback resolver, so apt
+      # fails with "Temporary failure resolving ...". Harmless where bridge
+      # DNS already works.
+      network: host
     container_name: innate-dev
```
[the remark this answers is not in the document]
karmanyaahm: Keeping `network: host` here — the "only the GPG fetch needs the network" premise isn't right. **7 of the 10 `RUN` steps** in `sim/Dockerfile` need DNS/internet: both `apt-get install` blocks, `setup_repos.sh`, the `uv` installer, `uv pip install`, oh-my-zsh, and `ssh-keyscan`. The setting exists for a documented reason: on systemd-resolved hosts the bridge build-network can't reach the `127.0.0.53` stub resolver, so apt fails with "Temporary failure resolving".

The suggested per-step `RUN --network=host` is a BuildKit instruction gated behind the `network.host` **entitlement** (`--allow network.host`), which `docker compose build` does not grant by default — those layers would fail with "network.host is not allowed". The compose `build.network: host` (legacy build-network) is a different mechanism that doesn't require the entitlement, which is exactly why it's used.

This is also a dev-only compose file, and the steps that could run untrusted code (the package installs) are precisely the ones that need the network — so per-RUN narrowing wouldn't mitigate the stated threat. Resolving as won't-fix.

webapp/proxy/https_server.py
```diff
@@ -493,25 +201,38 @@ async def process_request(connection, request):
...
+    """The browser has previously loaded over https (cookie set there), so it has
+    accepted the self-signed cert and we can safely send it back to https."""
+    return any(part.strip() == f"{HTTPS_COOKIE}=1" for part in request.headers.get("Cookie", "").split(";"))
+
+
+def _redirect_to_https(request):
+    host = request.headers.get("Host", "").rsplit(":", 1)[0] or "0.0.0.0"
+    netloc = host if HTTPS_PORT == 443 else f"{host}:{HTTPS_PORT}"
+    location = f"https://{netloc}{request.path}"
+    return Response(307, "Temporary Redirect", Headers({"Location": location, "Content-Length": "0"}), b"")
```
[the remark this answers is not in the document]
karmanyaahm: Won't-fix. Exploiting this requires the attacker to control the `Host` header on the **victim's** request — i.e. an active MITM on the LAN (or a poisoned proxy). Anyone in that position can already rewrite the response itself, including the `Location`, so validating the Host here protects against nothing the attacker can't already do directly. It only adds an allowlist to keep in sync with the cert SANs.

Reverted the host-validation pass. The IPv6 parsing bug from the sibling comment is a real correctness issue and is kept/fixed.

webapp/proxy/https_server.py
```diff
@@ -493,25 +201,38 @@ async def process_request(connection, request):
...
+
+def _has_https_cookie(request) -> bool:
+    """The browser has previously loaded over https (cookie set there), so it has
+    accepted the self-signed cert and we can safely send it back to https."""
+    return any(part.strip() == f"{HTTPS_COOKIE}=1" for part in request.headers.get("Cookie", "").split(";"))
+
+
+def _redirect_to_https(request):
+    host = request.headers.get("Host", "").rsplit(":", 1)[0] or "0.0.0.0"
+    netloc = host if HTTPS_PORT == 443 else f"{host}:{HTTPS_PORT}"
```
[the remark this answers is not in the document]
karmanyaahm: Fixed in fb809b23 — `_strip_host_port` now strips the port with bracket awareness, so an IPv6 literal like `[::1]` (with or without a port) survives intact instead of being truncated to `[::1`.

sim/docker-compose.dev.yml
```diff
@@ -34,15 +34,38 @@ services:
...
+      - "127.0.0.1:9090:9090"
+      # webapp front door (replaces the deprecated sim/frontend UI): the in-container
+      # python server binds 443 (https) + 80 (http), both serving the full app.
+      # Override the published side (e.g. SIM_HTTPS_PORT=8443 ./innate sim up) to
+      # publish on a non-privileged port; the container always binds 443/80.
+      - "${SIM_HTTPS_PORT:-443}:443"
+      - "${SIM_HTTP_PORT:-80}:80"
+      # Local STUN responder (mars_cam) for WebRTC: a remote browser queries
+      # stun:<host>:3478 to learn its srflx (LAN) candidate. UDP, not TCP.
+      - "${SIM_STUN_PORT:-3478}:3478/udp"
```
karmanyaahm:
```suggestion

```
no need now that stun is in sim monolith

DavidDobas (approved):
LGTM

scripts/update/setup_repos.sh
```diff
@@ -0,0 +1,84 @@
...
+            _repos_log "    Adding NodeSource repository..."
+            echo "deb [signed-by=/usr/share/keyrings/nodesource.gpg] https://deb.nodesource.com/node_22.x nodistro main" | \
+                tee /etc/apt/sources.list.d/nodesource.list > /dev/null
+        fi
+
+        _repos_log "  NodeSource repository configured"
+    else
+        _repos_log "  NodeSource repository already configured"
+    fi
+}
```
karmanyaahm:
```suggestion

```

ros2_ws/apt-dependencies.common.txt
```diff
@@ -10,22 +13,23 @@ python3-vcstool
...
 ccache
 
+# Node.js (webapp runtime), pinned to the 22.x line from the NodeSource repo.
+# That repo is configured in scripts/update/setup_repos.sh (robot, via
+# post_update.sh) and ci/Dockerfile.test-base (CI); the pin only resolves where
+# it is present.
+# Keep the version glob unquoted: xargs passes it to apt-get verbatim, and
+# quotes here would be consumed by xargs as quoting delimiters.
+nodejs=22.*
+
```
karmanyaahm:
```suggestion


```

webapp/proxy/https_server.py
```diff
@@ -618,27 +303,15 @@ async def main():
...
+            # Same app over cleartext — no auto-upgrade; the arm panel offers a manual HTTPS switch.
             await stack.enter_async_context(
-                serve(
-                    _reject_ws,
-                    "0.0.0.0",
-                    HTTP_PORT,
-                    process_request=process_media_request,
-                    max_size=None,
-                )
+                serve(ws_handler, "0.0.0.0", HTTP_PORT, process_request=_dispatch_request, max_size=None)
```
theo-michel, line 312: **[1/14] The plain-HTTP listener (:80/:4080) now shares the full dispatch/WebSocket handler with HTTPS, so the `/ws` rosbridge relay, the `/settings` write channel, and `/restart` are reachable over unauthenticated cleartext — the base server kept all of these TLS-only (the HTTP port previously served only read-only `/episode` media via a separate handler).**

**Failure scenario:** A page loaded in an operator's browser on the robot's LAN (or a hostile site via DNS rebinding, which cleartext HTTP cannot resist) opens `ws://<robot>/ws` — WebSockets bypass CORS, and there's no Origin/auth check — and relays rosbridge JSON to advertise+publish `/cmd_vel` or arm topics, driving the robot with no authentication. In base, `_reject_ws` + `process_media_request` blocked exactly this on the HTTP port.

webapp/proxy/settings_routes.py
```diff
@@ -0,0 +1,50 @@
...
+        async for raw in connection:
+            try:
+                req = json.loads(raw)
+                sets = req.get("sets", []) or []
+                clears = req.get("clears", []) or []
+            except (ValueError, AttributeError, TypeError):
+                await connection.send(json.dumps({"ok": False, "message": "malformed request"}))
+                continue
+            try:
+                ok, msg = await asyncio.to_thread(settings_store.apply_changes, sets, clears)
```
theo-michel, line 45: **[2/14] The `/settings` WebSocket handler calls `settings_store.apply_changes` (writing `config/settings.yaml`) with no Origin check and no auth, and is served on the cleartext `:80` listener; only the HTTP `/restart` endpoint has an `X-Requested-By` guard, which this WS path lacks.**

**Failure scenario:** Any page in the operator's browser does `new WebSocket("ws://<robot>/settings")` (WS is exempt from CORS/preflight) and sends `{"sets":[{path,value,type}]}`, silently rewriting the ROS param overlay applied on the next restart — cross-site config tampering with no same-origin defense.

webapp/js/teleop/skillsMenu.js
```diff
@@ -0,0 +1,608 @@
...
+    const guidelines = skill.guidelines || skill.guidelines_when_running;
+    if (guidelines) head.title = guidelines;
+
+    const tail = document.createElement("span");
+    tail.className = "skills-pop-tail mono";
+    if (running) tail.textContent = "…";
+    else if (expandable) tail.textContent = isExpanded ? "▾" : "›";
+    head.append(dot, name, tail);
+
+    head.addEventListener("click", () => {
```
theo-michel, line 336: **[3/14] The header-click path that launches a parameter-less skill only checks that *that* skill isn't running (not any global in-flight run), so starting a second skill overwrites the singleton `run` object and discards the first skill's cancel handle.**

**Failure scenario:** Operator starts a long parameter-less skill A (row shows Stop), reopens the popup and clicks parameter-less skill B: `startRun(B)` replaces `run` (A's cancel closure lost), skills_server rejects B ("another skill is already running"), and A's row stops rendering a Stop button (`run.skillId` is now B) — the operator can no longer stop the physically-moving robot from the menu. The form Run button and Enter key *do* have the any-run guard; this path doesn't.

sim/src/webrtc/webrtc_server.py
```diff
@@ -155,22 +302,65 @@ async def _main(shared_queues) -> None:
...
+    """Drop signaling that a later START from the same client supersedes.
+
+    A START (with renegotiate) replaces that client's peer, so any earlier
+    start/answer/ice from the same client belongs to a session that is about to
+    die — building it wastes seconds, and applying its answer poisons the new
+    peer. Other clients' items and relative order are preserved.
+    """
+    latest_start: dict[str, int] = {}
+    for index, item in enumerate(items):
+        if item.get("kind") == "start":
```
theo-michel: **[4/14] `_coalesce` keys only on `kind=='start'` and drops any earlier queued item from the same client — including a pending SDP answer — behind ANY later START, even a no-renegotiate camera-toggle START that doesn't rebuild the peer.**

**Failure scenario:** Two viewers connected; a blocking peer-build for client B queues client A's answer together with A's roster-driven no-reneg `setActiveCameras` START. On drain, `_coalesce` discards A's answer and `on_start` takes the no-reneg path (no new offer), so A's aiortc peer never gets `setRemoteDescription` — A's video stays black until the browser's ~7s media watchdog forces a full re-handshake, which can re-hit the race.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_signaling.cpp
```diff
@@ -0,0 +1,505 @@
...
+    auto* self = static_cast<WebRTCStreamer*>(user_data);
+    GstWebRTCPeerConnectionState state;
+    g_object_get(webrtc, "connection-state", &state, nullptr);
+    const char* cid = static_cast<const char*>(g_object_get_data(G_OBJECT(webrtc), "client_id"));
+
+    // On connect, force a keyframe on every encoder so this peer (which may be joining a stream that's
+    // already running for others) gets a decodable IDR immediately. Teardown is handled by the health
+    // poll on the executor thread (set-state from here would deadlock the pipeline).
+    if (cid) {
+        std::lock_guard<std::mutex> lock(self->peers_mutex_);
```
theo-michel: **[5/14] This PR newly adds a `peers_mutex_` lock inside the webrtcbin connection-state callback (base only logged there); teardown runs `destroy_peer`→`~Peer`→`gst_element_set_state(NULL)` while holding `peers_mutex_`, and `set_state(NULL)` joins the same webrtcbin thread that's blocked acquiring `peers_mutex_` — an ABBA deadlock.**

**Failure scenario:** A browser tab closes mid-negotiation: webrtcbin's PC thread dispatches `notify::connection-state` (→FAILED/CLOSED) and blocks on `peers_mutex_`, while `poll_pipeline_health` (holding `peers_mutex_`) `destroy_peer()`s that same client and `set_state(NULL)` waits for the PC thread to exit — the executor thread hangs holding the lock, and all signaling, health polling, and camera fan-out for every client freeze permanently. Narrow race window, but a total node-wide freeze.

theo-michel (commented):
Automated deep review (xhigh effort: 10 finder angles + gap sweep, each candidate independently verified against the code). Posting the 14 anchored + 1 unanchored CONFIRMED findings inline below, ranked most-severe first. Two clusters stand out:

1. **Dual-port serving regression** — the new cleartext `:80`/`:4080` listener now shares the full request dispatch with HTTPS, so `/ws` (rosbridge relay) and `/settings` (config write) are reachable unauthenticated over LAN/DNS-rebind. The old server kept these TLS-only.
2. **WebRTC lifecycle bugs in the mars_cam split and sim server** — a newly-introduced ABBA deadlock in peer teardown, an SDP-answer-dropping race in the sim's signaling coalescer, and several live-peer-destroyed / zombie-peer-leaked paths.

Also flagged: skills-menu can silently lose the ability to Stop a running skill, a skill stuck showing "running" forever after an early exception, and a `mars.urdf`-adjacent nav/teleop pose race in the sim.



**Additional finding (can't be anchored inline — the affected method `#closePc` at `webapp/js/webrtcSession.js:577-592` is unchanged code, outside this diff's hunk context):**

`#closePc()` resets `#videoTrackCount` but never clears `#videoStreams` / `#videoLive` (only the map-only branch elsewhere clears `#videoLive`), so per-camera streams and liveness flags from a torn-down `RTCPeerConnection` survive into the rebuilt session.

**Failure scenario:** The media watchdog rebuilds the connection (`#closePc` + new handshake): during the multi-second renegotiation, `state.videoLive[i]` is still true and `state.videoStreams[i]` holds dead tracks, so PiP tiles skip the 'connecting' shimmer and promoting a thumbnail (`setPrimaryCamera`) reports status 'streaming' with a dead stream — the stage shows a frozen frame from the closed connection while reporting a live link.

Happy to expand on any of these or help scope fixes.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_signaling.cpp
```diff
@@ -0,0 +1,505 @@
...
+        destroy_subscriptions();    // topics changed; drop the old-source subs...
+        reconcile_subscriptions();  // ...and re-subscribe (lazily) whatever current peers still negotiate
+        RCLCPP_INFO(this->get_logger(), "Global source switched to %s", source.c_str());
+    }
+
+    const bool audio_active = enable_audio_ && request_audio;
+    // Peers negotiate the audio m-line up front (if a mic exists) so toggling is reneg-free, like cameras.
+    const bool negotiate_audio = enable_audio_;
+
+    if (videos.empty() && !audio_active) {
```
theo-michel, line 182: **[6/14] A START whose `video` array contains only unrecognized camera names is silently filtered to empty, and the `videos.empty() && !audio_active` branch treats empty as a disconnect request and destroys the client's live peer.**

**Failure scenario:** A connected teleop client sends `START {client_id, video:["wrist"], audio:false}` where "wrist" isn't a configured camera (typo or config mismatch): `find_camera` drops every entry with no warning, `videos` becomes empty, and the node calls `destroy_peer` — tearing down the operator's live video feed mid-teleoperation instead of rejecting the bad name and leaving existing streams intact.

sim/src/webrtc/webrtc_server.py
```diff
@@ -41,84 +46,225 @@ def _parse_ice(payload: dict) -> RTCIceCandidate | None:
...
+        self._peers: dict[str, _Peer] = {}
+
+    async def on_start(self, client_id: str, payload: dict) -> str | None:
+        """(Re)build this client's connection + offer, or — for a no-reneg START —
+        just retarget which cameras it encodes. Returns the offer SDP, or None when
+        no new offer is needed (a no-reneg active-set change)."""
+        # `video` is the active (encoded) set the browser wants; fall back to the
+        # first camera so something shows before the UI learns the real roster.
+        requested = payload.get("video") or []
+        active = [c for c in requested if c in CAMERAS] or CAMERAS[:1]
```
theo-michel: **[7/14] An empty `video` list (the webapp's map-only "release the peer" message) is coerced to `CAMERAS[:1]` instead of destroying the peer, diverging from the robot's `on_start` which releases on no-streams.**

**Failure scenario:** Operator switches the teleop stage to map-only: `WebRtcSession` closes its RTCPeerConnection and publishes `START {video:[], audio:false}` expecting release. The sim instead re-activates first_person encoding on the now-dead peer (or, for a fresh client_id, builds a new pc+offer the browser ignores), leaving a zombie peer in `_peers` that never reaches failed/closed — it keeps holding CameraTracks and burning VP8 encode CPU. The documented "map-only costs zero streaming" contract is false in sim.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_signaling.cpp
```diff
@@ -0,0 +1,505 @@
...
+        RCLCPP_INFO(this->get_logger(), "START requested no streams; releasing peer");
+        destroy_peer(client_id);
+        return;
+    }
+
+    // Stream/audio switch on an already-set-up independent peer (its negotiated m-lines unchanged): flip
+    // which cameras + audio are pushed live, with no renegotiation/ICE — instant, not a reconnect.
+    if (!renegotiate) {
```
theo-michel, line 190: **[8/14] A re-sent START without `renegotiate` matches an existing (possibly dead, not-yet-reaped) peer via `peers_.find` and only calls `update_peer_active` — it never republishes an offer — contradicting `WEBRTC_CLIENT_ID_MIGRATION.md`'s instruction to "re-send START to (re)connect".**

**Failure scenario:** A client whose connection dropped re-sends START with the same client_id (as the migration doc directs, with no renegotiate flag). The FAILED/DISCONNECTED peer still sits in `peers_` (reaped only after ~15s of grace polls or the 15s connect timeout), so the node flips active streams on the dead transport and sends no offer — the client can't reconnect until the watchdog reaps the peer, giving up to ~15s of video outage the documented protocol says shouldn't happen.

ros2_ws/src/brain/brain_client/brain_client/nodes/skills_server.py
```diff
@@ -248,24 +299,62 @@ def execute_callback(self, goal_handle):
                 )
             self._skill_running = True
 
+        name = self._skill_display_name(skill_type)
+        run_id = uuid.uuid4().hex
+        self._publish_skill_status(run_id, skill_type, name, "running")
```
theo-michel, line 304: **[9/14] The 'running' `/brain/skill_status_update` is published before the `try` block, but code between it and the `try` (`skill.set_feedback_callback` / `skill.get_required_robot_states`) can raise, propagating past the terminal-status publish (the `try` has only a `finally`, no `except`), so no completed/failed status ever follows.**

**Failure scenario:** A custom code skill whose `get_required_robot_states()` raises is launched: clients receive `{status:'running'}` and then nothing terminal, so the webapp Skills button and Agent panel show the skill as the active skill forever — misleading every connected operator until a different skill run happens to reset it.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
```diff
@@ -87,862 +66,478 @@ void mars_playout_delay_ext_init(MarsPlayoutDelayExt* self) {
...
+    offer_id_pub_ = this->create_publisher<std_msgs::msg::String>("/webrtc/offer_id", 10);
+    ice_out_id_pub_ = this->create_publisher<std_msgs::msg::String>("/webrtc/ice_out_id", 10);
+    active_streams_pub_ = this->create_publisher<std_msgs::msg::String>("/webrtc/active_streams", 10);
 
     start_sub_ = this->create_subscription<std_msgs::msg::String>(
         "/webrtc/start", 10, std::bind(&WebRTCStreamer::on_start, this, std::placeholders::_1));
-
-    // Create initial subscriptions for live source
-    create_subscriptions("live");
+    answer_id_sub_ = this->create_subscription<std_msgs::msg::String>(
```
theo-michel, line 133: **[10/14] The legacy bare rosbridge topics `/webrtc/answer` and `/webrtc/ice_in` (subscriptions) and `/webrtc/offer`, `/webrtc/ice_out` (publications) were removed with no compatibility path — only client_id-tagged `*_id` topics remain, and `on_start` ignores a START with no client_id — breaking every deployed client still on the raw-SDP protocol.**

**Failure scenario:** A robot updates to this release before the mobile controller app ships the client_id migration (the migration doc itself states old clients "will not connect on this branch"): the app publishes its answer on `/webrtc/answer`, which no node subscribes to, and never receives an offer — mobile teleop video is permanently black with no error surfaced on the robot side. A real deploy-ordering hazard, not just an intended migration.

webapp/debug/webrtc.html
```diff
@@ -0,0 +1,559 @@
...
+        let j; try { j = JSON.parse(raw); } catch { el.innerHTML = '<div class="e">bad JSON</div>'; return; }
+        // The node publishes the real camera list (m-line order). Adopt it and rebuild the UI when it
+        // differs from what we have — this is what makes the page work for 1, 3, N cameras.
+        if (Array.isArray(j.cameras) && j.cameras.length &&
+            (j.cameras.length !== cameras.length || j.cameras.some((c, i) => c !== cameras[i]))) {
+          cameras = j.cameras.slice();
+          rebuildCameras();
+        }
+        if (!j.count) { el.innerHTML = '<div class="muted">no clients connected</div>'; return; }
+        el.innerHTML = (j.clients || []).map((c) => {
```
theo-michel, line 457: **[11/14] `renderActive()` interpolates unescaped client-controlled fields (`c.source`, `c.connection_state`, `s.name`, `s.topic`) from `/webrtc/active_streams` into `innerHTML`, allowing HTML/script injection into any operator's open debug page.**

**Failure scenario:** A peer on the robot's LAN publishes a START whose `source` is e.g. ``; the robot echoes it in `/webrtc/active_streams` and every browser with `debug/webrtc.html` open executes the injected markup (relayed stored XSS). Camera names are also injected unescaped into `id`/`data-l` attributes, breaking the `.tog[data-l="..."]` querySelector for names containing quotes. Reach is limited to operators who open the debug page.

sim/src/simulation/simulation_node.py
```diff
@@ -1991,51 +2069,27 @@ def run(self):
...
-                    # Store commanded velocities in world frame for odometry
-                    # MULTIPLY BY 0 TO DISABLE MESSING WITH TH EACTUAL ROBOT
-                    # ON ROS
-                    self.commanded_lin_vel = forward_dir * linear_vel * 0
-                    self.commanded_ang_vel = np.array([0, 0, angular_vel])
-
-                    print(f"Commanded vel: linear={linear_vel:.3f}, angular={angular_vel:.3f}")
+                    # Hold the command; it is integrated every step below until it
+                    # expires, so the pose advances smoothly between arrivals.
+                    self.held_velocity_cmd = latest_velocity_cmd
```
theo-michel, line 2079: **[12/14] The held-command supersede check is asymmetric: a position/nav command clears `held_velocity_cmd`, but an incoming `/cmd_vel` never clears an active `nav_target_pos`, so both the velocity integrator and nav movement mutate the base pose in the same physics step.**

**Failure scenario:** A directive issues a nav goal (`nav_target_pos` set), then the operator nudges the teleop joystick sending `/cmd_vel`: `held_velocity_cmd` is set but `nav_target_pos` stays set, so each step applies both `_apply_velocity_cmd` and `_update_navigation_movement`, and the new ~0.25s `SIM_VELOCITY_CMD_TIMEOUT` prolongs the tug-of-war → erratic/jittery robot motion for up to the timeout after each teleop input.

webapp/js/teleop/armPanel.js
```diff
@@ -77,8 +77,29 @@ export function createArmPanel(parent, rosClient, opts = {}) {
...
+      // only happens off-box, which is exactly where this redirect helps.)
+      hint.textContent = "Leader arm needs HTTPS — switch over and continue past the certificate warning.";
+      const switchBtn = document.createElement("button");
+      switchBtn.className = "arm-button";
+      switchBtn.type = "button";
+      switchBtn.textContent = "Switch to HTTPS";
+      switchBtn.addEventListener("click", () => {
+        const url = new URL(location.href);
+        url.protocol = "https:";
+        url.port = ""; // 80 -> default 443; the TLS front door listens there
```
theo-michel, line 95: **[13/14] The 'Switch to HTTPS' button blanks the URL port (forcing default 443) instead of mapping the serving http↔https port pair (80→443 AND 4080→4443), so it dead-ends whenever the app was reached on the proxy's direct HTTP port.**

**Failure scenario:** An operator opens `http://<robot>:4080` (the port `https_server.py` actually binds on the robot) where the iptables PREROUTING redirect isn't installed/active — or `http://localhost:4080` on the robot itself, where PREROUTING never applies — and clicks 'Switch to HTTPS': the button navigates to `https://<robot>:443` where nothing listens (HTTPS is on 4443) → ERR_CONNECTION_REFUSED, and the working `:4080` URL is gone from the address bar, so leader-arm WebSerial teleop is unreachable.

sim/src/webrtc/webrtc_server.py
```diff
@@ -41,84 +46,225 @@ def _parse_ice(payload: dict) -> RTCIceCandidate | None:
...
-            # Only tear down on terminal states — "disconnected" is transient and
-            # usually self-heals. Guard against a stale handler: a new `start`
-            # replaces self._pc, and the old pc's late state change must not close
-            # the live connection.
-            if pc is self._pc and pc.connectionState in ("failed", "closed"):
-                await self.close()
+            # Tear down only THIS client's pc on a terminal state. Guard against a
+            # stale handler: a rebuild replaces the peer, and the old pc's late state
+            # change must not evict the new one. "disconnected" is transient.
+            if self._peers.get(client_id) is peer and pc.connectionState in ("failed", "closed"):
```
theo-michel, line 161: **[14/14] Per-client peers are evicted only on connectionState 'failed'/'closed'; 'disconnected' is deliberately ignored, and each browser tab/reload uses a fresh `randomUUID` client_id, so abruptly-closed viewers accumulate as peers still holding CameraTracks.**

**Failure scenario:** A user repeatedly opens/reloads the teleop or agent page (new client_id each time) or loses network without a clean close: those peers sit in `_peers` in state 'disconnected', each with its CameraTracks eligible for lazy VP8 encoding, until aiortc eventually reaches 'failed' — under churn the sim spends CPU encoding frames for phantom viewers and `encode_counts` grows unbounded.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_signaling.cpp
DavidDobas: Fixed in 1dea9ad5. The callback no longer touches `peers_mutex_` at all: `media_ready` is now a shared `std::atomic<bool>` tagged onto the webrtcbin element with `g_object_set_data_full` (same lock-free pattern as `mars_gen`), so the PC thread writes it directly and `~Peer`'s `set_state(NULL)` can safely join that thread while the executor holds the mutex. All three readers (video/audio fan-out, status JSON) load it atomically; the shared atomic outlives the Peer, so a late notify after teardown is harmless.

webapp/js/teleop/skillsMenu.js
DavidDobas: Fixed in 1dea9ad5 — the header-click path now uses the same ANY-run + connected guard as the form Run button/Enter key, so launching a second parameter-less skill while one runs is a no-op instead of overwriting the run slot and discarding the running skill's cancel handle.

sim/src/webrtc/webrtc_server.py
DavidDobas: Fixed in 1dea9ad5 — `_coalesce` now only treats STARTs with `renegotiate` as superseding. A no-reneg camera-toggle START keeps the peer, so the answer queued ahead of it is still needed by the live handshake and is no longer dropped.

ros2_ws/src/brain/brain_client/brain_client/nodes/skills_server.py
DavidDobas: Fixed in 1dea9ad5 — an `except` now publishes a terminal `failed` status (with the exception message as reason) before re-raising, so a 'running' broadcast is always followed by a terminal one even when execution raises between them.

sim/src/webrtc/webrtc_server.py
DavidDobas: Fixed in 1dea9ad5 — an explicitly-empty `video` list (with no audio) now closes the peer, mirroring the robot's release path, so map-only really costs zero streaming. Deliberately kept the first-camera fallback for a START that *omits* `video` or names only unknown cameras — per your [6/14], killing a live feed over a typo is the worse failure mode.

sim/src/webrtc/webrtc_server.py
DavidDobas: Fixed in 1dea9ad5 — added a `reap_stale` pass (5s cadence): a peer that never reaches CONNECTED within 30s (abandoned handshake / reload churn) or sits 'disconnected' for 30s is released, mirroring mars_cam's connect-timeout + health-poll behavior. 'disconnected' stays transient within the grace window.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_signaling.cpp
DavidDobas: Deliberately not changed in this batch: it's a robot-side C++ behavior change and both shipped clients (webapp + controller-app #226) only request names from the robot's own roster/config, so the typo path isn't reachable from real clients today. Agreed the right behavior is reject-and-keep rather than destroy — note the sim side already does that as of 1dea9ad5 (unknown-only names fall back instead of tearing down). Queued as a follow-up alongside the next mars_cam change that gets hardware time.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_signaling.cpp
DavidDobas: Deliberately unchanged: both shipped clients always send `renegotiate: true` when they rebuild after a drop (controller-app #226 does this on every handshake for exactly this reason), so the reconnect path in practice always re-offers; a no-reneg START is only ever used on a live peer for camera/mic toggles. The real bug here is the migration doc's wording — 're-send START to (re)connect' should say 'with renegotiate: true'. Will fix the doc rather than add dead-peer detection to the no-reneg path.

webapp/proxy/https_server.py
DavidDobas: Deferred, deliberately — with the posture argument made explicit in the PR description now. On the robot the LAN control plane was already open via unauthenticated `ws://robot:9090` rosbridge, so the HTTP listener doesn't grant an attacker anything new on the LAN. The one vector cleartext does add is DNS rebinding from a hostile page, and the right counter for that is an Origin allowlist on the WS upgrade paths (`/ws`, `/settings`) — planned as the immediate follow-up rather than resurrecting the sticky redirect (which protects nothing against an on-path attacker and traps first-time visitors on the self-signed cert warning).

webapp/proxy/settings_routes.py
DavidDobas: Deferred with [1/14] — same fix planned: an Origin check on the `/settings` WS upgrade (reject cross-origin upgrades outright), which kills the cross-site tampering vector since browsers always send Origin on WS handshakes. Agreed this one is genuinely new surface (unlike /ws, which mirrors the already-open rosbridge), so it's first in line of the deferred items.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
DavidDobas: Intentional break, coordinated rather than shimmed: keeping the bare-topic path alive would mean maintaining the single-peer pipeline next to the multi-peer one (the old topics carry no client identity, so they can't map onto per-client peers cleanly). The release ordering is now called out in the PR description's 'Before merge/ship': this ships together with controller-app #226, whose MIN_ROBOT_VERSION bump keeps updated apps off old robots — and old apps get a dead camera only if a robot updates ahead of the app rollout, which the release plan avoids.

webapp/debug/webrtc.html
DavidDobas: Deferred to the P3 polish pass — agreed it should escape (or switch to textContent/dataset) before the debug page is pointed at untrusted LANs. Reach today is an operator who deliberately opens debug/webrtc.html against a robot that already has a hostile LAN peer, at which point the rosbridge port is a bigger prize than the debug DOM.

sim/src/simulation/simulation_node.py
DavidDobas: Deferred — real, but it's sim motion-quality (robot nav has its own arbitration) and the fix wants care: clearing nav_target_pos on any /cmd_vel would make a joystick brush silently kill an agent's nav goal, which is its own operator surprise. Needs a deliberate priority rule (probably teleop-wins + explicit nav-goal-cancelled feedback) rather than a symmetric clear.

webapp/js/teleop/armPanel.js
DavidDobas: Deferred to the P3 pass, but agreed on the shape: map the http↔https port *pair* (80→443, 4080→4443, and honor SIM_HTTPS_PORT) instead of blanking the port. It's also greptile's standing P1 on this PR — will land it with the polish batch.

PR #473, opened 2026-06-25
feat: robot speaker volume control on web Settings page (INN-621)

theo-michel:
## What
Adds a live **Speaker volume** control to the web app **Settings** page — a 0–100 slider that adjusts the robot's speaker output, mirroring the mobile app.

- Reads the current value from `/robot/info` (`volume_percent`, default 80).
- On slider **release**, calls the `/set_volume` service (`mars_msgs/srv/SetVolume`, `volume_percent` 0–100) and reverts + shows an error if the robot rejects it.
- Disabled with a hint until the robot socket is connected.

## Why
Feature request from Axel (hackathon chat): you couldn't change the robot's volume from the web app. INN-621.

## Notes
- Unlike the `settings.yaml` knobs (which need a robot restart), this applies **immediately** and persists on the robot, so it's a distinct live card and talks to rosbridge directly. The Settings page now auto-connects the shared `ros` socket the same way `pageMount` does.
- Does **not** touch robot/ROS code — `/set_volume` and the `/robot/info` `volume_percent` field already exist server-side.

## Files
- `webapp/js/constants.js` — `SET_VOLUME_SERVICE`
- `webapp/js/types.d.ts` — `RobotInfo.volume_percent`
- `webapp/js/settings/main.js` — the volume card + socket bootstrap

## Verification
`tsc --checkJs` (strict): no new type errors introduced (25 pre-existing, all in unrelated files). Functional test of the slider needs a live robot.

Closes INN-621.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
+++ b/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
@@ -940,7 +940,23 @@ class AppControl : public rclcpp::Node {
      * audible rather than ~-70dB (INN-467).
      */
     void apply_alsa_volume(int percent) {
-        std::string cmd = "amixer -M sset Master " + std::to_string(percent) + "% 2>/dev/null";
+        // Lift the low end: across roughly the lower half of amixer's range the
+        // speaker is inaudible on this hardware (~25% silent, ~60% barely audible)
+        // even with -M, so remap the stored 0-100 onto an audible
+        // [AUDIBLE_FLOOR, 100] band before handing it to amixer (which still
+        // applies its -M perceptual curve on top). 0 stays muted; the stored
+        // volume_percent (and /robot/info) keep the user-facing value. Retune
+        // AUDIBLE_FLOOR if the bottom of the slider still feels off.
+        constexpr int AUDIBLE_FLOOR = 55;
+        int mixer_percent;
+        if (percent <= 0) {
+            mixer_percent = 0;
+        } else if (percent >= 100) {
+            mixer_percent = 100;
+        } else {
+            mixer_percent = AUDIBLE_FLOOR + (100 - AUDIBLE_FLOOR) * percent / 100;
+        }
+        std::string cmd = "amixer -M sset Master " + std::to_string(mixer_percent) + "% 2>/dev/null";
         int ret = std::system(cmd.c_str());
         if (ret != 0) {
             RCLCPP_WARN(this->get_logger(), "amixer sset Master failed (rc=%d)", ret);
--- a/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
+++ b/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
@@ -940,13 +940,8 @@ class AppControl : public rclcpp::Node {
      * audible rather than ~-70dB (INN-467).
      */
     void apply_alsa_volume(int percent) {
-        // Lift the low end: across roughly the lower half of amixer's range the
-        // speaker is inaudible on this hardware (~25% silent, ~60% barely audible)
-        // even with -M, so remap the stored 0-100 onto an audible
-        // [AUDIBLE_FLOOR, 100] band before handing it to amixer (which still
-        // applies its -M perceptual curve on top). 0 stays muted; the stored
-        // volume_percent (and /robot/info) keep the user-facing value. Retune
-        // AUDIBLE_FLOOR if the bottom of the slider still feels off.
+        // Remap 0-100 onto the audible [AUDIBLE_FLOOR, 100] band; the speaker is
+        // silent across the lower range even with -M. 0 still mutes.
         constexpr int AUDIBLE_FLOOR = 55;
         int mixer_percent;
         if (percent <= 0) {
--- a/webapp/js/settings/main.js
+++ b/webapp/js/settings/main.js
@@ -207,83 +207,20 @@ function resetAll() {
   recompute();
 }
 
-const DEFAULT_VOLUME = 80; // robot's built-in default (raw percent) until /robot/info arrives.
+const DEFAULT_VOLUME = 80; // robot's built-in default (percent) until /robot/info arrives.
 
-// On this hardware the speaker is effectively silent below ~AUDIBLE_FLOOR% even
-// though the robot already applies amixer's perceptual -M scaling (INN-467), so
-// a raw 0–100 slider wastes its lower half and then jumps loud near the top.
-// Instead the slider is a perceptual *position* (0–100) mapped onto the audible
-// band only: position 0 = muted, any audible position lands in
-// [AUDIBLE_FLOOR, 100] and reads as a friendly level (Low/Medium/High/Max)
-// rather than a misleading number. AUDIBLE_FLOOR is the one knob to retune if
-// the quietest audible step still feels off or too loud on real hardware.
-const AUDIBLE_FLOOR = 60;
-
-/** Clamp to an integer in 0–100 (shared by slider positions and percents). */
-function clampPercent(/** @type {number} */ value) {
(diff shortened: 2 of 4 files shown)


webapp/js/settings/main.js
```diff
@@ -192,6 +207,132 @@ function resetAll() {
...
+    try {
+      /** @type {{ success: boolean, message?: string }} */
+      const res = await ros.callService(SET_VOLUME_SERVICE, { volume_percent: next });
+      if (!res.success) throw new Error(res.message || "Failed to set volume.");
+      robotVolume = next;
+      setLiveStatus("Volume set.", "ok");
+    } catch {
+      renderValue(previous);
+      setLiveStatus("Couldn't set volume. Try again.", "err");
+    } finally {
```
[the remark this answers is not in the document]
theo-michel: Done — re-seed `robotVolume = previous` in the catch block so the revert target stays consistent even if a `/robot/info` update moved it mid-save (commit e2c5a32a).

webapp/js/settings/main.js
```diff
@@ -192,6 +207,132 @@ function resetAll() {
...
+  slider.addEventListener("input", () => {
+    dragging = true;
+    cur.textContent = slider.value;
+  });
+
+  slider.addEventListener("change", async () => {
+    dragging = false;
+    const next = clampVolume(Number(slider.value));
+    renderValue(next);
+    if (next === robotVolume || saving) return;
```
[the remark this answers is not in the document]
theo-michel: Done — the slider is now disabled while a save is in flight (via a shared `refreshEnabled` helper that also respects connection state), so a mid-save release can no longer be silently discarded (commit e2c5a32a).

theo-michel:
Update — reworked the volume UX based on testing feedback (low values were inaudible until ~60%, then jumped loud):

- **Web slider reverted to a plain 0–100% control** (`webapp/js/settings/main.js`): the number shown is the real stored `volume_percent` (1:1), so it's simple and stable across page revisits — dropped the perceptual position/level-label mapping and its localStorage workaround.
- **Low-end lift moved robot-side** (`ros2_ws/.../mars_control/app.cpp`, `apply_alsa_volume`): the stored 0–100 is remapped onto an audible `[AUDIBLE_FLOOR, 100]` band before `amixer -M`, so every low value is louder while the stored/displayed percent stays user-facing. Benefits the mobile app too.

⚠️ This PR now includes a **robot C++ change**, so it needs a robot rebuild/redeploy (not just the static webapp) to take effect. `AUDIBLE_FLOOR` (currently 55) is the single tuning knob — worth confirming on hardware that the bottom of the slider is comfortably audible without the top getting too hot.

webapp/js/settings/main.js
```diff
@@ -192,6 +207,145 @@ function resetAll() {
...
+  const renderValue = (/** @type {number} */ percent) => {
+    slider.value = String(percent);
+    cur.textContent = String(percent);
+  };
+
+  // Disabled until connected, and while a save is in flight so a mid-save
+  // release can't be silently dropped (nor re-enabled by a state change).
+  const refreshEnabled = () => {
+    slider.disabled = ros.state !== "connected" || saving;
+  };
```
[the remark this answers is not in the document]
theo-michel: Done in fcfa588e — `refreshEnabled` now clears `dragging` whenever it disables the slider, so a mid-drag disconnect can no longer leave the guard stuck and freeze the slider after reconnect.

vignesh-anand:
**No in-page recovery if the first rosbridge connect fails**

The Settings page auto-connects via `ros.connect(servedHost)` but never mounts a connect panel, so recovery depends on which connection state you land in — and the two cases diverge:

**Transient drop (already connected once) — fine.** `#handleClose` sees `#everConnected === true` and goes into the reconnecting branch (exp-backoff, 1s→10s cap, forever). The slider disables, `refreshEnabled()` clears `dragging`, and on reopen rosClient replays the `/robot/info` subscription so the next tick re-enables and re-renders. Self-heals, no user action.

**First connect never opens — stuck.** If the very first connect never reaches OPEN (rosbridge down, wrong host, 8s TCP timeout), `#everConnected` is still `false`, so `#handleClose` takes the fail-fast branch: state → `disconnected`, **no reconnect scheduled**. The slider stays disabled on "Connect to the robot to set volume." with no way to retry — the page has no connect card, and the header badge only *disconnects* (disabled unless already connected). Only a full page reload recovers.

Pages that go through `mountPage` show the connect card on `disconnected`, giving a retry/address-entry path; this page doesn't.

**Severity: low.** The page is robot-served, so a first-connect failure usually means the robot is unreachable anyway (the yaml knobs over the proxy would be failing too). But the asymmetry is real — a one-time hiccup at page load leaves volume control dead with no in-page recovery, while a later drop recovers automatically.

Cheapest fix: re-`ros.connect(servedHost)` on a `disconnected` transition in the settings `onStateChange`, or surface a small "retry" affordance.

theo-michel:
> No in-page recovery if the first rosbridge connect fails

Done in 86cd8290 — the Settings page now retries the **initial**-connect-failed case itself: on a `disconnected` transition it re-`ros.connect(servedHost)` after a 5s debounce (so a refused connection can't spin), but only until the first successful open. After that, a later drop still self-heals through rosClient's own `reconnecting` path, and an explicit disconnect via the header badge is left alone. So a one-time hiccup at page load no longer strands the volume control needing a full reload.

webapp/js/settings/main.js
```diff
@@ -10,12 +10,46 @@
...
 import { CATALOG } from "./catalog.js";
 
 initShell("settings", "../");
 const stage = /** @type {HTMLElement} */ (document.getElementById("stage"));
 
+// The yaml knobs below talk to the proxy, but the speaker-volume control is a
+// live rosbridge service call, so this page needs the shared socket. Mirror
+// pageMount's bootstrap: reach the serving host, except on laptop dev.
+const servedHost = location.hostname;
+if (servedHost && servedHost !== "localhost" && servedHost !== "127.0.0.1") {
```
DavidDobas: We dont want to support localhost anymore. But I guess there are still some parts of the code which allow it?

ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
```diff
@@ -940,7 +940,23 @@ class AppControl : public rclcpp::Node {
...
      */
     void apply_alsa_volume(int percent) {
-        std::string cmd = "amixer -M sset Master " + std::to_string(percent) + "% 2>/dev/null";
+        // Lift the low end: across roughly the lower half of amixer's range the
+        // speaker is inaudible on this hardware (~25% silent, ~60% barely audible)
+        // even with -M, so remap the stored 0-100 onto an audible
+        // [AUDIBLE_FLOOR, 100] band before handing it to amixer (which still
+        // applies its -M perceptual curve on top). 0 stays muted; the stored
+        // volume_percent (and /robot/info) keep the user-facing value. Retune
+        // AUDIBLE_FLOOR if the bottom of the slider still feels off.
```
DavidDobas: Code speaks for itself

DavidDobas (approved):

webapp/js/settings/main.js
theo-michel: Done in 4d5e2e4c — dropped the `localhost`/`127.0.0.1` skip here; it now just connects to the serving host. (FWIW the "other parts" you sensed are `pageMount.js` + `connectPanel.js`, which still branch on localhost for the dev connect-card flow — happy to strip those in a follow-up if we want localhost gone repo-wide.)

ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
theo-michel: Done in e8943a1e — trimmed the comment block down to a one-line note on the floor remap.

PR #474, opened 2026-06-25
On-robot Diffusion Policy support (policy_type switch)

vignesh-anand:
Adds on-robot Diffusion Policy inference alongside ACT, selected by `execution.policy_type`. The ACT path is untouched.

## What's here
- `manipulation/DP.py` — self-contained Diffusion Policy (sibling of ACT.py): vendored U-Net + ResNet18(GroupNorm) obs encoder + mean/std normalize with the **exact** module/buffer names used in training (so checkpoints load), plus a **dependency-free DDIM** sampler (matches diffusers to ~1e-7). The denoising call is isolated in `_unet_forward` for a future TensorRT swap.
- `manipulation_server.py` — `_load_policy_for_behavior` branches on `policy_type`; "dp" builds DiffusionPolicy and shares the existing `select_action` inference loop.
- `config_validation.py` — adds `execution.policy_type` (`"act"`|`"dp"`, default `"act"`).
- training-client `activate` — tags `policy_type` from the checkpoint name (`dp_policy_*` → `"dp"`).
- `manipulation/dp_benchmark.py` — standalone (no-ROS) latency sweep over DDIM steps vs the 40 ms / 25 Hz budget.

## Validation (local)
A training checkpoint (InnateDP) loads into `DP.py` with 0 missing / 0 unexpected keys; DDIM matches diffusers to ~1e-7; benchmark runs and reports the latency/steps tradeoff.

## Notes
- v1 ships `n_obs_steps=1`; multi-frame obs history is a follow-up.
- TensorRT engine build at `DP._unet_forward` is the next step to hit the Jetson budget.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,4 +107,7 @@ data/.last_mode
 # arm_wave directory (local data, not tracked)
 arm_wave/
 
+# sock experiments (local scratch, not tracked)
+sock_experiments/
+
 .vscode
--- a/ros2_ws/src/brain/manipulation/manipulation/DP.py
+++ b/ros2_ws/src/brain/manipulation/manipulation/DP.py
@@ -1,0 +1,482 @@
+"""On-robot Diffusion Policy — self-contained sibling of ACT.py.
+
+Loads the ACT-shaped artifacts the cloud DP preset produces (a
+``dp_policy_step_*.pth`` EMA state_dict + ``dataset_stats.pt``) and exposes the
+same ``select_action(batch) -> (B, action_dim)`` surface ManipulationServer uses
+for ACT, so the inference loop is unchanged.
+
+Self-contained on purpose (like ACT.py): vendors the U-Net, the multi-image obs
+encoder, and mean/std normalization with the *exact* module/buffer names used in
+training (so ``load_state_dict`` matches), plus a dependency-free DDIM sampler
+(no ``diffusers``) matching the training scheduler (cosine betas, epsilon
+prediction). The denoising U-Net call is isolated in ``_unet_forward`` so it can
+be swapped for a TensorRT engine without touching the sampling loop.
+"""
+
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass, field
+from collections import deque
+from enum import Enum
+from typing import Dict, List, Optional, Tuple
+
+import torch
+import torch.nn as nn
+import torchvision
+
+
+# ── normalization (mirrors training dp/normalize.py exactly) ─────────
+class NormalizationMode(str, Enum):
+    IDENTITY = "identity"
+    MEAN_STD = "mean_std"
+
+
+class FeatureType(str, Enum):
+    VISUAL = "VISUAL"
+    STATE = "STATE"
+    ACTION = "ACTION"
+
+
+@dataclass
+class PolicyFeature:
+    type: FeatureType
+    shape: List[int]
+
+
+def _buffer_shape(feature: PolicyFeature) -> List[int]:
+    if feature.type == FeatureType.VISUAL:
+        return [feature.shape[0], 1, 1]
+    return list(feature.shape)
+
+
+def _sanitize(key: str) -> str:
+    return key.replace(".", "_")
+
+
+class _NormBase(nn.Module):
+    def __init__(self, features, norm_map, stats=None):
+        super().__init__()
+        self._modes = {}
+        for key, feature in features.items():
+            mode = norm_map.get(feature.type.value, NormalizationMode.IDENTITY.value)
+            self._modes[key] = mode
+            if mode != NormalizationMode.MEAN_STD.value:
+                continue
+            shape = _buffer_shape(feature)
+            sk = _sanitize(key)
+            mean = torch.full(shape, float("inf"))
+            std = torch.full(shape, float("inf"))
+            if stats is not None and key in stats:
+                if "mean" in stats[key]:
+                    mean = stats[key]["mean"].clone().to(torch.float32).view(shape)
+                if "std" in stats[key]:
+                    std = stats[key]["std"].clone().to(torch.float32).view(shape)
+            self.register_buffer(f"{sk}__mean", mean)
+            self.register_buffer(f"{sk}__std", std)
+
+    def _mean_std(self, key):
+        sk = _sanitize(key)
+        return getattr(self, f"{sk}__mean"), getattr(self, f"{sk}__std")
+
+
+class Normalize(_NormBase):
+    def forward(self, batch):
+        out = dict(batch)
+        for key, mode in self._modes.items():
+            if key not in out or mode != NormalizationMode.MEAN_STD.value:
+                continue
+            mean, std = self._mean_std(key)
+            out[key] = (out[key] - mean) / (std + 1e-8)
+        return out
+
+
+class Unnormalize(_NormBase):
+    def forward(self, batch):
+        out = dict(batch)
+        for key, mode in self._modes.items():
+            if key not in out or mode != NormalizationMode.MEAN_STD.value:
+                continue
+            mean, std = self._mean_std(key)
+            out[key] = out[key] * std + mean
+        return out
+
+
+DEFAULT_NORM_MAP = {
+    FeatureType.VISUAL.value: NormalizationMode.MEAN_STD.value,
+    FeatureType.STATE.value: NormalizationMode.MEAN_STD.value,
+    FeatureType.ACTION.value: NormalizationMode.MEAN_STD.value,
+}
+
+
+# ── conditional U-Net (mirrors training dp/conditional_unet1d.py) ────
+class SinusoidalPosEmb(nn.Module):
+    def __init__(self, dim):
+        super().__init__()
+        self.dim = dim
+
+    def forward(self, x):
+        device = x.device
+        half = self.dim // 2
+        emb = math.log(10000) / (half - 1)
+        emb = torch.exp(torch.arange(half, device=device) * -emb)
+        emb = x[:, None] * emb[None, :]
+        return torch.cat((emb.sin(), emb.cos()), dim=-1)
+
+
+class Downsample1d(nn.Module):
+    def __init__(self, dim):
+        super().__init__()
+        self.conv = nn.Conv1d(dim, dim, 3, 2, 1)
+
+    def forward(self, x):
+        return self.conv(x)
+
+
+class Upsample1d(nn.Module):
+    def __init__(self, dim):
+        super().__init__()
+        self.conv = nn.ConvTranspose1d(dim, dim, 4, 2, 1)
+
+    def forward(self, x):
+        return self.conv(x)
+
+
+class Conv1dBlock(nn.Module):
+    def __init__(self, inp, out, kernel_size, n_groups=8):
+        super().__init__()
+        self.block = nn.Sequential(
+            nn.Conv1d(inp, out, kernel_size, padding=kernel_size // 2),
+            nn.GroupNorm(n_groups, out),
+            nn.Mish(),
+        )
+
+    def forward(self, x):
+        return self.block(x)
+
+
+class ConditionalResidualBlock1D(nn.Module):
+    def __init__(self, in_ch, out_ch, cond_dim, kernel_size=3, n_groups=8, cond_predict_scale=False):
+        super().__init__()
+        self.blocks = nn.ModuleList(
+            [
+                Conv1dBlock(in_ch, out_ch, kernel_size, n_groups=n_groups),
+                Conv1dBlock(out_ch, out_ch, kernel_size, n_groups=n_groups),
+            ]
+        )
+        cond_channels = out_ch * 2 if cond_predict_scale else out_ch
+        self.cond_predict_scale = cond_predict_scale
+        self.out_channels = out_ch
+        self.cond_encoder = nn.Sequential(
+            nn.Mish(), nn.Linear(cond_dim, cond_channels), nn.Unflatten(-1, (-1, 1))
+        )
+        self.residual_conv = nn.Conv1d(in_ch, out_ch, 1) if in_ch != out_ch else nn.Identity()
+
+    def forward(self, x, cond):
+        out = self.blocks[0](x)
+        embed = self.cond_encoder(cond)
+        if self.cond_predict_scale:
+            embed = embed.reshape(embed.shape[0], 2, self.out_channels, 1)
+            out = embed[:, 0, ...] * out + embed[:, 1, ...]
+        else:
+            out = out + embed
+        out = self.blocks[1](out)
+        return out + self.residual_conv(x)
+
+
+class ConditionalUnet1D(nn.Module):
+    def __init__(self, input_dim, global_cond_dim=None, diffusion_step_embed_dim=128,
+                 down_dims=(256, 512, 1024), kernel_size=5, n_groups=8, cond_predict_scale=True):
+        super().__init__()
+        all_dims = [input_dim] + list(down_dims)
+        start_dim = down_dims[0]
+        dsed = diffusion_step_embed_dim
+        self.diffusion_step_encoder = nn.Sequential(
+            SinusoidalPosEmb(dsed), nn.Linear(dsed, dsed * 4), nn.Mish(), nn.Linear(dsed * 4, dsed)
+        )
+        cond_dim = dsed + (global_cond_dim or 0)
+        in_out = list(zip(all_dims[:-1], all_dims[1:]))
+        mid_dim = all_dims[-1]
+
+        def blk(a, b):
+            return ConditionalResidualBlock1D(a, b, cond_dim, kernel_size, n_groups, cond_predict_scale)
+
+        self.mid_modules = nn.ModuleList([blk(mid_dim, mid_dim), blk(mid_dim, mid_dim)])
+        self.down_modules = nn.ModuleList([])
+        for ind, (di, do) in enumerate(in_out):
+            is_last = ind >= (len(in_out) - 1)
+            self.down_modules.append(
+                nn.ModuleList([blk(di, do), blk(do, do), Downsample1d(do) if not is_last else nn.Identity()])
+            )
+        self.up_modules = nn.ModuleList([])
+        for ind, (di, do) in enumerate(reversed(in_out[1:])):
+            is_last = ind >= (len(in_out) - 1)
+            self.up_modules.append(
+                nn.ModuleList([blk(do * 2, di), blk(di, di), Upsample1d(di) if not is_last else nn.Identity()])
+            )
+        self.final_conv = nn.Sequential(
+            Conv1dBlock(start_dim, start_dim, kernel_size=kernel_size), nn.Conv1d(start_dim, input_dim, 1)
+        )
+
+    def forward(self, sample, timestep, global_cond=None):
+        sample = sample.transpose(1, 2)  # (B,T,D) -> (B,D,T)
+        if not torch.is_tensor(timestep):
+            timestep = torch.tensor([timestep], dtype=torch.long, device=sample.device)
+        elif timestep.ndim == 0:
+            timestep = timestep[None].to(sample.device)
+        timesteps = timestep.expand(sample.shape[0])
+        gfeat = self.diffusion_step_encoder(timesteps)
+        if global_cond is not None:
+            gfeat = torch.cat([gfeat, global_cond], dim=-1)
+        x = sample
+        h = []
+        for resnet, resnet2, downsample in self.down_modules:
+            x = resnet(x, gfeat)
+            x = resnet2(x, gfeat)
+            h.append(x)
+            x = downsample(x)
+        for m in self.mid_modules:
+            x = m(x, gfeat)
+        for resnet, resnet2, upsample in self.up_modules:
+            x = torch.cat((x, h.pop()), dim=1)
+            x = resnet(x, gfeat)
+            x = resnet2(x, gfeat)
+            x = upsample(x)
+        x = self.final_conv(x)
+        return x.transpose(1, 2)  # (B,D,T) -> (B,T,D)
+
+
+# ── obs encoder (mirrors training dp/obs_encoder.py) ─────────────────
+def _replace_bn_with_gn(root, fpg=16):
+    for name, child in root.named_children():
+        if isinstance(child, nn.BatchNorm2d):
+            setattr(root, name, nn.GroupNorm(max(1, child.num_features // fpg), child.num_features))
+        else:
+            _replace_bn_with_gn(child, fpg)
+    return root
+
+
+class MultiImageObsEncoder(nn.Module):
+    def __init__(self, rgb_keys, low_dim_keys, obs_shapes, resnet_name="resnet18", use_group_norm=True):
+        super().__init__()
+        self.rgb_keys = list(rgb_keys)
+        self.low_dim_keys = list(low_dim_keys)
+        self.obs_shapes = obs_shapes
+        self.key_model_map = nn.ModuleDict()
+        for key in self.rgb_keys:
+            model = getattr(torchvision.models, resnet_name)(weights=None)
+            model.fc = nn.Identity()
+            if use_group_norm:
+                model = _replace_bn_with_gn(model)
+            self.key_model_map[_sanitize(key)] = model
+        self._output_dim = self._compute_output_dim()
+
+    @torch.no_grad()
+    def _compute_output_dim(self):
+        ex = {}
+        for key in self.rgb_keys:
+            c, h, w = self.obs_shapes[key]
+            ex[key] = torch.zeros(1, c, h, w)
+        for key in self.low_dim_keys:
+            ex[key] = torch.zeros(1, *self.obs_shapes[key])
+        return self.forward(ex).shape[-1]
+
+    @property
+    def output_dim(self):
+        return self._output_dim
+
+    def forward(self, obs_dict):
+        feats = []
+        for key in self.rgb_keys:
+            feats.append(self.key_model_map[_sanitize(key)](obs_dict[key]))
+        for key in self.low_dim_keys:
+            feats.append(obs_dict[key])
+        return torch.cat(feats, dim=-1)
+
+
+# ── DDIM sampler (no diffusers; matches DDPM cosine betas) ───────────
+def _cosine_alphas_cumprod(num_train_timesteps: int) -> torch.Tensor:
+    """``squaredcos_cap_v2`` schedule (diffusers betas_for_alpha_bar)."""
+    def alpha_bar(t):
+        return math.cos((t + 0.008) / 1.008 * math.pi / 2) ** 2
+
+    betas = []
+    for i in range(num_train_timesteps):
+        t1 = i / num_train_timesteps
+        t2 = (i + 1) / num_train_timesteps
+        betas.append(min(1 - alpha_bar(t2) / alpha_bar(t1), 0.999))
+    alphas = 1.0 - torch.tensor(betas, dtype=torch.float64)
+    return torch.cumprod(alphas, dim=0)
+
+
+class DDIMSampler:
+    """Deterministic DDIM (eta=0), epsilon prediction, clip_sample=True."""
+
+    def __init__(self, num_train_timesteps=100, num_inference_steps=16, clip_sample=True):
+        self.acp = _cosine_alphas_cumprod(num_train_timesteps)
+        self.final_alpha = torch.tensor(1.0, dtype=torch.float64)  # set_alpha_to_one
+        self.step_ratio = num_train_timesteps // num_inference_steps
+        self.timesteps = [int(t) for t in (
+            torch.arange(0, num_inference_steps) * self.step_ratio
+        ).round().flip(0).tolist()]
+        self.clip_sample = clip_sample
+
+    def step(self, eps, t, sample):
+        prev_t = t - self.step_ratio
+        a_t = self.acp[t]
+        a_prev = self.acp[prev_t] if prev_t >= 0 else self.final_alpha
+        a_t = a_t.to(sample.device, sample.dtype)
+        a_prev = a_prev.to(sample.device, sample.dtype)
+        beta_t = 1 - a_t
+        x0 = (sample - beta_t.sqrt() * eps) / a_t.sqrt()
+        if self.clip_sample:
+            x0 = x0.clamp(-1, 1)
+        direction = (1 - a_prev).sqrt() * eps
+        return a_prev.sqrt() * x0 + direction
+
+
+# ── policy ───────────────────────────────────────────────────────────
+@dataclass
+class DPRuntimeConfig:
+    input_shapes: Dict[str, List[int]] = field(default_factory=dict)
+    output_shapes: Dict[str, List[int]] = field(default_factory=dict)
+    n_obs_steps: int = 1
+    horizon: int = 30
+    n_action_steps: int = 8
+    num_train_timesteps: int = 100
+    num_inference_steps: int = 16
+    down_dims: Tuple[int, ...] = (256, 512, 1024)
+    kernel_size: int = 5
+    n_groups: int = 8
+    diffusion_step_embed_dim: int = 128
+    cond_predict_scale: bool = True
+    vision_backbone: str = "resnet18"
+    use_group_norm: bool = True
+
+    def _features(self, shapes):
+        out = {}
+        for k, s in shapes.items():
+            t = FeatureType.VISUAL if "image" in k else (FeatureType.ACTION if k == "action" else FeatureType.STATE)
+            out[k] = PolicyFeature(type=t, shape=list(s))
+        return out
+
+    @property
+    def input_features(self):
+        return self._features(self.input_shapes)
+
+    @property
+    def output_features(self):
+        return self._features(self.output_shapes)
+
+    @property
+    def image_keys(self):
+        return [k for k in self.input_shapes if "image" in k]
+
+    @property
+    def state_keys(self):
+        return [k for k in self.input_shapes if "image" not in k]
+
+    @property
+    def action_dim(self):
+        return self.output_shapes["action"][0]
+
+
+class DiffusionPolicy(nn.Module):
+    def __init__(self, config: DPRuntimeConfig, dataset_stats=None):
+        super().__init__()
+        self.config = config
+        self.normalize_inputs = Normalize(config.input_features, DEFAULT_NORM_MAP, dataset_stats)
+        self.normalize_targets = Normalize(config.output_features, DEFAULT_NORM_MAP, dataset_stats)
+        self.unnormalize_outputs = Unnormalize(config.output_features, DEFAULT_NORM_MAP, dataset_stats)
+        self.obs_encoder = MultiImageObsEncoder(
+            config.image_keys, config.state_keys, config.input_shapes,
+            config.vision_backbone, config.use_group_norm,
+        )
+        self.unet = ConditionalUnet1D(
+            input_dim=config.action_dim,
+            global_cond_dim=self.obs_encoder.output_dim * config.n_obs_steps,
+            diffusion_step_embed_dim=config.diffusion_step_embed_dim,
+            down_dims=config.down_dims, kernel_size=config.kernel_size,
+            n_groups=config.n_groups, cond_predict_scale=config.cond_predict_scale,
+        )
+        self.sampler = DDIMSampler(config.num_train_timesteps, config.num_inference_steps)
+        self._action_queue: deque = deque([], maxlen=config.n_action_steps)
+
+    def reset(self):
+        self._action_queue = deque([], maxlen=self.config.n_action_steps)
+
+    def _unet_forward(self, sample, t, global_cond):
+        """Isolated denoising call — swap for a TensorRT engine here."""
+        return self.unet(sample, t, global_cond=global_cond)
+
+    def _encode_obs(self, nb):
+        cfg = self.config
+        B = nb[cfg.image_keys[0]].shape[0]
+        per_step = {}
+        for key in cfg.image_keys + cfg.state_keys:
+            x = nb[key]
+            is_image = key in cfg.image_keys
+            if (is_image and x.ndim == 4) or (not is_image and x.ndim == 2):
+                x = x.unsqueeze(1)
+            To = x.shape[1]
+            per_step[key] = x[:, :To].reshape(B * To, *x.shape[2:])
+        return self.obs_encoder(per_step).reshape(B, -1)
+
+    @torch.no_grad()
+    def predict_action(self, batch):
+        cfg = self.config
+        nb = self.normalize_inputs(batch)
+        gcond = self._encode_obs(nb)
+        B = gcond.shape[0]
+        traj = torch.randn(B, cfg.horizon, cfg.action_dim, device=gcond.device)
+        for t in self.sampler.timesteps:
+            eps = self._unet_forward(traj, t, gcond)
+            traj = self.sampler.step(eps, t, traj)
+        actions = self.unnormalize_outputs({"action": traj})["action"]
+        start = cfg.n_obs_steps - 1
+        return actions[:, start:start + cfg.n_action_steps]
+
+    @torch.no_grad()
+    def select_action(self, batch):
+        self.eval()
+        if len(self._action_queue) == 0:
+            actions = self.predict_action(batch)
+            self._action_queue.extend(actions.transpose(0, 1))
+        return self._action_queue.popleft()
+
+
+def load_dp_policy(checkpoint_path, dataset_stats, action_dim, device,
+                   n_action_steps=None, **overrides):
+    """Build a DiffusionPolicy, load EMA weights + stats, return it ready to run.
+
+    Mirrors ACTPolicy loading in manipulation_server: tolerant to compile/DDP
+    prefixes and ``strict=False`` key matching.
+    """
+    cfg = DPRuntimeConfig(
+        input_shapes={
+            "observation.image_camera_1": [3, 224, 224],
+            "observation.image_camera_2": [3, 224, 224],
+            "observation.state": [6],
+        },
+        output_shapes={"action": [action_dim]},
+        **overrides,
+    )
+    if n_action_steps:
+        cfg.n_action_steps = n_action_steps
+
+    state_dict = torch.load(checkpoint_path, map_location=device)
+    if isinstance(state_dict, dict) and "state_dict" in state_dict:
+        state_dict = state_dict["state_dict"]
+    cleaned = {}
+    for k, v in state_dict.items():
+        if k.startswith("module."):
+            k = k[len("module."):]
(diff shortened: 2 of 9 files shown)


ros2_ws/src/cloud/clients/training-client/training_client/src/skill_manager.py
```diff
@@ -499,18 +499,28 @@ def activate_run(
             if "execution" not in meta:
                 meta["execution"] = {}
 
+            # Infer policy family from the checkpoint name so the robot loads the
+            # right policy: DP emits dp_policy_step_*.pth, ACT emits act_policy_step_*.pth.
+            policy_type = "dp" if Path(checkpoint).name.startswith("dp_policy") else "act"
```
vignesh-anand, line 504: Mixed checkpoints in one run dir decide policy family by step number, not intent. `_find_latest_checkpoint` does a recursive `rglob("*_step_*.pth")` and returns whichever has the highest step across *all* matches. If a run directory ever contains both an ACT and a DP checkpoint (re-used dir, resumed run, stray file), `policy_type` here is derived from whichever happens to have the larger step count — decoupled from what the operator intended.

Not introduced by this PR, but the new prefix logic now inherits it: a stray higher-step checkpoint of the other family would silently flip `policy_type`. Edge case, and the recursive glob widens the blast radius.

ros2_ws/src/cloud/clients/training-client/training_client/src/skill_manager.py
```diff
@@ -499,18 +499,28 @@ def activate_run(
...
             checkpoint,
             stats_file,
+            policy_type,
         )
 
-        return {"checkpoint": checkpoint, "stats_file": stats_file}
+        return {
+            "checkpoint": checkpoint,
+            "stats_file": stats_file,
+            "policy_type": policy_type,
```
vignesh-anand, line 522: Stale docstring: the return shape now includes `policy_type`, but the docstring above (L463-464) still says *"Returns a dict with `checkpoint` and `stats_file` paths"*. Since this is the function's contract doc and the PR changed the return shape, worth updating it to mention `policy_type`.

ros2_ws/src/cloud/clients/training-client/training_client/src/skill_manager.py
vignesh-anand, line 504: The architecture decision rests on an unverifiable, untested, cross-system string convention with no fail-safe. The "truth" of whether a checkpoint is DP lives in a filename produced by a *different system not in this repo* — there's no shared constant, no contract, and no test pinning it. If the cloud preset ever emits `diffusion_policy_step_*.pth`, `dp_ema_step_*.pth`, or capitalizes it, this silently returns `"act"`. The failure then surfaces downstream and confusingly: the robot loads DP weights through the ACT loader and dies with a key/shape mismatch, not a clear "policy type misdetected." Given DP support is the whole point of the PR, this binary guess is the weakest link.

Compounding it: there's no "unrecognized" branch — the `else` swallows *every* non-matching name (including a malformed/renamed checkpoint) into `"act"` with no signal that classification was uncertain.

Cheap fix — check both known prefixes and surface the miss:

```python
name = Path(checkpoint).name
if name.startswith("dp_policy"):
    policy_type = "dp"
elif name.startswith("act_policy"):
    policy_type = "act"
else:
    logger.warning("Checkpoint %s matches neither dp_policy/act_policy; defaulting to act", name)
    policy_type = "act"
```

Better long-term (probably a follow-up): have training embed `policy_type` in the checkpoint or a sidecar `config.json` so activation reads a *declared* value instead of guessing from a name.

ros2_ws/src/brain/manipulation/manipulation/config_validation.py
```diff
@@ -150,6 +150,14 @@ class LearnedExecCfg(_BaseExecCfg):
...
     # checkpoint is loaded; the schema only enforces ``>= 1``.
     n_action_steps: int | None = Field(None, ge=1)
+    # DP-only: number of DDIM denoising steps at inference. None = use the model
+    # default baked into DP.py (16). Lower (8/4) is faster but rougher; ignored
+    # for ACT.
+    num_inference_steps: int | None = Field(None, ge=1)
+    # Which policy family the checkpoint belongs to. Set by `training-client
+    # activate` from the training preset. "act" (default) keeps every existing
+    # skill unchanged; "dp" selects the Diffusion Policy loader/inference path.
+    policy_type: Literal["act", "dp"] = "act"
```
vignesh-anand, line 160: No test coverage for either new field, despite an established pattern right there. `test/test_config_validation.py` already tests `n_action_steps` thoroughly — `test_full_happy_path` (sets `n_action_steps: 40`), `test_n_action_steps_rejected`, `test_n_action_steps_null_accepted` — but this PR adds `policy_type` and `num_inference_steps` with no corresponding tests and doesn't touch the test file.

Cheap to add and would pin the contract: `"dp"`/`"act"` accepted, `"foo"` rejected (the `Literal` guard), default is `"act"`, null → default; `num_inference_steps` rejects `0`/negative/`NaN` and null → `None`. Since this schema is the contract layer, untested new contract fields are the main gap in this section.

vignesh-anand, line 160: Note (no action needed here, defense-in-depth): the schema trusts `policy_type` with no way to verify it against the actual checkpoint. A misclassified or hand-edited `"dp"` on a real ACT checkpoint validates cleanly and routes to the wrong loader, failing downstream with a key/shape mismatch. Not fixable at this layer (the schema can't see checkpoint contents) — it's the classifier concern from `skill_manager.activate_run` resurfacing. Flagging only so it's clear the schema adds no safety net for it.

ros2_ws/src/brain/manipulation/manipulation/config_validation.py
```diff
@@ -150,6 +150,14 @@ class LearnedExecCfg(_BaseExecCfg):
     # chunk_size-aware clamping happens inside create_act_config once the
     # checkpoint is loaded; the schema only enforces ``>= 1``.
     n_action_steps: int | None = Field(None, ge=1)
+    # DP-only: number of DDIM denoising steps at inference. None = use the model
+    # default baked into DP.py (16). Lower (8/4) is faster but rougher; ignored
```
vignesh-anand, line 154: Minor doc coupling: this comment hardcodes DP.py's constant ("baked into DP.py (16)"). If DP.py's `num_inference_steps` default ever changes, this comment silently drifts. Cosmetic, but worth a note like "the DP.py default" without the literal, or a shared constant.

ros2_ws/src/brain/manipulation/manipulation/config_validation.py
```diff
@@ -150,6 +150,14 @@ class LearnedExecCfg(_BaseExecCfg):
...
+    # for ACT.
+    num_inference_steps: int | None = Field(None, ge=1)
```
vignesh-anand, line 156: `num_inference_steps` is DP-only but lives in the shared `LearnedExecCfg`, so it's silently ignored for ACT. An operator can set `num_inference_steps: 8` on an ACT skill; it validates cleanly and does nothing (the comment even says "ignored for ACT") — no signal that the knob had no effect.

Unlike the horizon / num_train_timesteps relationships (whose bounding value isn't in the schema), this one *is* enforceable here since both fields are in the model:

```python
@model_validator(mode="after")
def _warn_dp_only_knob(self):
    if self.num_inference_steps is not None and self.policy_type == "act":
        logger.warning("num_inference_steps is ignored for policy_type=act")
    return self
```

Low severity, but it's the one cross-field check that genuinely belongs at this layer.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+
+    def step(self, eps, t, sample):
+        prev_t = t - self.step_ratio
+        a_t = self.acp[t]
+        a_prev = self.acp[prev_t] if prev_t >= 0 else self.final_alpha
+        a_t = a_t.to(sample.device, sample.dtype)
+        a_prev = a_prev.to(sample.device, sample.dtype)
+        beta_t = 1 - a_t
+        x0 = (sample - beta_t.sqrt() * eps) / a_t.sqrt()
+        if self.clip_sample:
```
vignesh-anand, line 335: **clip_sample=True clamps z-scored actions to [-1,1] — saturates any action beyond ±1σ.** Actions are `MEAN_STD` normalized (`DEFAULT_NORM_MAP`), so normalized targets routinely exceed [-1,1]; clamping `x0` here truncates every large motion toward the mean. In stock LeRobot/diffusers, `clip_sample=True` is only safe with min-max-to-[-1,1] action normalization — and this file has no MIN_MAX mode at all.

Failure: an action dim whose true target is 2.5σ above mean has normalized value ~2.5; the final denoise steps clamp its `x0` to 1.0, unnormalized to `mean + 1.0*std` instead of `mean + 2.5*std`. Please confirm the training scheduler's `clip_sample` setting + action normalization mode; if training used mean_std actions, this should be `clip_sample=False`.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+        super().__init__()
+        self._modes = {}
+        for key, feature in features.items():
+            mode = norm_map.get(feature.type.value, NormalizationMode.IDENTITY.value)
+            self._modes[key] = mode
+            if mode != NormalizationMode.MEAN_STD.value:
+                continue
+            shape = _buffer_shape(feature)
+            sk = _sanitize(key)
+            mean = torch.full(shape, float("inf"))
```
vignesh-anand, line 68: **Missing/failed `dataset_stats` degrades to NaN actions with only a warning — no hard fail.** mean/std init to `float("inf")` as a poison sentinel, but nothing ever checks it. If `dataset_stats.pt` fails to load (the server catches, warns, and passes `None`) and the checkpoint doesn't carry these buffers, `strict=False` leaves them `inf` → `Normalize` produces NaN → the robot commands NaN joint targets. The warm-up runs `select_action` but discards the result without a finiteness check (only the standalone `dp_test_inference.py` checks `isfinite`).

The `inf` sentinel intends fail-loud but never completes the contract. Suggest asserting `torch.isfinite` over the normalize buffers after load (or a finiteness gate on the warm-up output) so a stats problem aborts the load instead of silently reaching the arm.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+        else:
+            gcond = self._encode_obs(self.normalize_inputs(batch))
+        B = gcond.shape[0]
+        traj = torch.randn(B, cfg.horizon, cfg.action_dim, device=gcond.device)
+        for t in self.sampler.timesteps:
+            eps = self._unet_forward(traj, t, gcond)
+            traj = self.sampler.step(eps, t, traj)
+        actions = self.unnormalize_outputs({"action": traj})["action"]
+        start = cfg.n_obs_steps - 1
+        return actions[:, start:start + cfg.n_action_steps]
```
vignesh-anand, line 447: **`n_action_steps > horizon` silently under-fills the action queue.** This slice on a `horizon`-length trajectory yields at most `horizon` actions, while the deque `maxlen` is the oversized value. The ACT-oriented global `default_n_action_steps` can leak into DP via `_execute_learned_behavior` and halve replan reactivity with a misleading `n_action_steps=40` log.

Fix: clamp in the loader, mirroring ACT's `create_act_config` — `n_action_steps = min(n_action_steps, horizon - (n_obs_steps - 1))`, with a warning when it clamps.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+    return torch.cumprod(alphas, dim=0)
+
+
+class DDIMSampler:
+    """Deterministic DDIM (eta=0), epsilon prediction, clip_sample=True."""
+
+    def __init__(self, num_train_timesteps=100, num_inference_steps=16, clip_sample=True):
+        self.acp = _cosine_alphas_cumprod(num_train_timesteps)
+        self.final_alpha = torch.tensor(1.0, dtype=torch.float64)  # set_alpha_to_one
+        self.step_ratio = num_train_timesteps // num_inference_steps
```
vignesh-anand, line 321: **`num_inference_steps > num_train_timesteps` → `step_ratio = 0` → garbage output.** Integer division gives 0, so every entry of `self.timesteps` is 0 and `step()` never advances toward `t=0`/`final_alpha` — the denoise loop produces noise-level actions with no error. `num_inference_steps` is schema-validated only as `ge=1` (no upper bound), so this is reachable via config.

This is the right enforcement point for the Section-2 finding: guard here (`step_ratio = max(1, ...)` or an explicit raise) and it covers the config, benchmark, and dp_trt paths at once.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+def load_dp_policy(checkpoint_path, dataset_stats, action_dim, device,
+                   n_action_steps=None, **overrides):
+    """Build a DiffusionPolicy, load EMA weights + stats, return it ready to run.
+
+    Mirrors ACTPolicy loading in manipulation_server: tolerant to compile/DDP
+    prefixes and ``strict=False`` key matching.
+    """
+    cfg = DPRuntimeConfig(
```
vignesh-anand, line 465: **`load_dp_policy` hardcodes the entire obs schema** — exactly 2 cameras, `3×224×224`, state dim `6`, fixed key names; only `action_dim` is parameterized. A checkpoint trained with a different state dim or camera set has its mismatched normalize buffers *silently skipped* by `strict=False`, then crashes (or broadcasts wrongly) at first real inference — and the warm-up won't catch it because it feeds the same hardcoded dim-6/224 dummy batch.

Acknowledged v1 scope, but the failure mode (silent skip → runtime crash on the robot, invisible to warm-up) is worth hardening — at least assert the checkpoint-inferred shapes against these constants.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+            "observation.image_camera_2": [3, 224, 224],
+            "observation.state": [6],
+        },
+        output_shapes={"action": [action_dim]},
+        **overrides,
+    )
+    if n_action_steps:
+        cfg.n_action_steps = n_action_steps
+
+    state_dict = torch.load(checkpoint_path, map_location=device)
```
vignesh-anand, line 477: **`torch.load` without `weights_only=True`.** The ACT path deliberately uses `weights_only` ("faster/safer for a pure tensor state_dict") and loads to CPU with mmap; DP loads the full checkpoint straight to `device` via pickle. Minor perf/consistency/safety gap — align with the ACT loader (`weights_only=True`, CPU + mmap, then `.to(device)`).

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+        direction = (1 - a_prev).sqrt() * eps
+        return a_prev.sqrt() * x0 + direction
+
+
+# ── policy ───────────────────────────────────────────────────────────
+@dataclass
+class DPRuntimeConfig:
+    input_shapes: Dict[str, List[int]] = field(default_factory=dict)
+    output_shapes: Dict[str, List[int]] = field(default_factory=dict)
+    n_obs_steps: int = 1
```
vignesh-anand, line 346: **`n_obs_steps` is configurable but only `1` is actually supported.** `_encode_obs` and the server obs pipeline assume a single frame; setting `n_obs_steps>1` (via `**overrides`) makes `global_cond` the wrong width vs the U-Net's `global_cond_dim = output_dim * n_obs_steps`, i.e. a silent shape mismatch/crash. Acknowledged follow-up, but the live config knob invites the misconfiguration — consider asserting `n_obs_steps == 1` until multi-frame obs lands.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+be swapped for a TensorRT engine without touching the sampling loop.
+"""
+
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass, field
+from collections import deque
+from enum import Enum
+from typing import Dict, List, Optional, Tuple
```
vignesh-anand, line 22: `Optional` is imported but never used as a type (the only "Optional" in the file is a comment). Drop it from the import.

ros2_ws/src/brain/manipulation/manipulation/DP.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+class Normalize(_NormBase):
+    def forward(self, batch):
+        out = dict(batch)
+        for key, mode in self._modes.items():
+            if key not in out or mode != NormalizationMode.MEAN_STD.value:
+                continue
+            mean, std = self._mean_std(key)
+            out[key] = (out[key] - mean) / (std + 1e-8)
```
vignesh-anand, line 90: Minor: `Normalize` divides by `std + 1e-8` but `Unnormalize` multiplies by bare `std` (line 101) — not exact inverses. Negligible for real std values, noting for completeness / symmetry.

ros2_ws/src/brain/manipulation/manipulation/dp_trt.py
```diff
@@ -0,0 +1,343 @@
...
+    image_shape = cfg.input_shapes[cfg.image_keys[0]]
+    state_dim = cfg.input_shapes[cfg.state_keys[0]][0]
+
+    enc_path = build_encoder_engine(policy, checkpoint_path, device, precision, log)
+    unet_path = build_unet_engine(policy, checkpoint_path, device, precision, log)
+    policy._trt_encoder = TRTEncoder(enc_path, cond_dim, image_shape, state_dim, device)
+    policy._trt_unet = TRTUNet(unet_path, cfg.action_dim, cfg.horizon, cond_dim, device)
+
+
+def _main():
```
vignesh-anand, line 324: **DP engines are never pre-built — this is an operational regression vs ACT.** ACT's engines are built at download time by `innate_training_node`: `workers.py` spawns `python -m manipulation.act_trt <checkpoint> <action_dim> fp32` per checkpoint ([workers.py:270](https://github.com/innate-inc/innate-os/blob/vig/dp-policy/ros2_ws/src/cloud/innate_training_node/innate_training_node/workers.py#L270)) plus a `--batch` startup sweep ([workers.py:463](https://github.com/innate-inc/innate-os/blob/vig/dp-policy/ros2_ws/src/cloud/innate_training_node/innate_training_node/workers.py#L463)), so the 1-2 min build is paid off the critical path.

Nothing pre-builds DP. The only reference to `dp_trt` outside this file is the lazy `from manipulation.dp_trt import attach_engines` inside `_load_dp_policy_for_behavior` — i.e. **the first activation of a DP skill builds both engines inline on the robot, blocking ~1-2 min during skill load.** And this `_main` has no `--batch` mode to plug into the sweep even if someone wanted to wire it up, and (unlike ACT's `_main`) no `validate_action_dim` guard.

Fix should mirror ACT: have `innate_training_node` detect `dp_policy_*` checkpoints and spawn `python -m manipulation.dp_trt`, and add a `--batch` mode + action_dim validation here to match `act_trt._main`.

ros2_ws/src/brain/manipulation/manipulation/dp_trt.py
```diff
@@ -0,0 +1,343 @@
...
+    def __call__(self, batch):
+        self._execute({
+            "image_camera_1": batch["observation.image_camera_1"],
+            "image_camera_2": batch["observation.image_camera_2"],
+            "state": batch["observation.state"],
+        })
+        return self._buffers["global_cond"].clone()
+
+
+def attach_engines(policy, checkpoint_path: str, device, precision: str = "fp16", log=print):
```
vignesh-anand, line 306: On precision: rather than the current split (ACT defaults fp32, DP defaults fp16), I think **both ACT and DP should run in fp16 (or bf16)**. The engine budget on Orin basically forces reduced precision for DP anyway (16 U-Net calls/replan), and keeping ACT at fp32 just means two different accuracy/perf regimes to reason about. Unify on one.

If accuracy under fp16 is a worry — especially for DP, where rounding compounds across the iterative denoise — prefer **bf16**: same throughput on Ampere/Orin, but its wider exponent range is much more forgiving of the intermediate magnitudes in the sampler than fp16's. So: make precision consistent across both paths, default to fp16/bf16, and characterize the action RMSE for DP the way `act_trt` already did for ACT before locking it in.

ros2_ws/src/brain/manipulation/manipulation/dp_trt.py
```diff
@@ -0,0 +1,343 @@
...
+        batch = {
+            "observation.image_camera_1": image_camera_1,
+            "observation.image_camera_2": image_camera_2,
+            "observation.state": state,
+        }
+        return self.policy._encode_obs(self.policy.normalize_inputs(batch))
+
+
+# ── engine cache path ────────────────────────────────────────────────
+def engine_path_for(checkpoint_path: str, kind: str, dims_tag: str, precision: str = "fp16") -> str:
```
vignesh-anand, line 71: **D2 — no `dataset_stats` signature in the cache key; ACT has exactly this.** `act_trt` hashes `dataset_stats.pt` into the engine filename ([act_trt.py:58-90](https://github.com/innate-inc/innate-os/blob/vig/dp-policy/ros2_ws/src/brain/manipulation/manipulation/act_trt.py#L58)) *because the graph constant-folds mean/std into the engine* — regenerate stats in place with the same checkpoint path and it would otherwise silently reuse an engine baking the old normalization ("systematically wrong actions").

DP's **encoder engine bakes obs mean/std the same way** (`_EncoderExportWrapper`), yet `dims_tag` omits any stats hash — the docstring just says "delete the engine manually if you regenerate stats." That's the exact assumption ACT's author judged unsafe. Failure: regenerate `dataset_stats.pt` in place, keep the checkpoint, and DP serves stale normalization with no error. Fix is nearly free — reuse `act_trt._dataset_stats_signature` and add `.s{sig}` to the tag.

ros2_ws/src/brain/manipulation/manipulation/dp_trt.py
```diff
@@ -0,0 +1,343 @@
...
+    eager torch.
+    """
+    cfg = policy.config
+    cond_dim = _cond_dim(policy)
+    image_shape = cfg.input_shapes[cfg.image_keys[0]]
+    state_dim = cfg.input_shapes[cfg.state_keys[0]][0]
+
+    enc_path = build_encoder_engine(policy, checkpoint_path, device, precision, log)
+    unet_path = build_unet_engine(policy, checkpoint_path, device, precision, log)
+    policy._trt_encoder = TRTEncoder(enc_path, cond_dim, image_shape, state_dim, device)
```
vignesh-anand, line 320: **D3 — eager weights stay resident after attach; ACT frees them.** ACT replaces the policy object with `TRTACTPolicy`, which holds no torch weights so the eager model is released after the engine builds ([act_trt.py:221](https://github.com/innate-inc/innate-os/blob/vig/dp-policy/ros2_ws/src/brain/manipulation/manipulation/act_trt.py#L221)). Here we *attach* engines to the live `DiffusionPolicy`, so the eager `self.unet` + 2× ResNet18 encoder stay in memory **alongside** both TRT engines. Once TRT is attached, `_unet_forward`/`predict_action` never touch the eager modules — they're dead weight, ~2× the model footprint ACT would use on a RAM-constrained Orin.

If the eager path only matters before a successful attach, drop the eager modules post-attach (or offload them). At minimum this should be a conscious choice, not an accident of attaching-vs-replacing.

ros2_ws/src/brain/manipulation/manipulation/dp_trt.py
```diff
@@ -0,0 +1,343 @@
...
+            expected = tuple(buffers[name].shape)
+            actual = tuple(self._engine.get_tensor_shape(name))
+            if actual != expected:
+                raise RuntimeError(
+                    f"Engine tensor {name!r} shape {actual} != expected {expected}; "
+                    "cached engine was built for a different architecture"
+                )
+            self._context.set_tensor_address(name, buffers[name].data_ptr())
+
+    def _execute(self, inputs: dict):
```
vignesh-anand, line 242: **D4 — no profiling instrumentation; ACT's is consumed by the server.** ACT's runner records `last_engine_ms` / `engine_ran` (CUDA events) and `last_disagreement`, which `_run_inference_once` reads for the inference-profile topic. This runner records none, and `DiffusionPolicy` exposes none — so the server's `getattr(self.current_policy, "engine_ran", True)` / `"last_engine_ms", 0.0` fall to defaults. Net: **DP telemetry always reports `engine_ran=True`, `engine_ms=0`** even when the engine ran. Telemetry-only, but silently wrong for DP — consider recording engine time here like `act_trt._run_engine` does.

ros2_ws/src/brain/manipulation/manipulation/dp_trt.py
vignesh-anand, line 320: **S1 — partial attach leaves a mixed TRT-encoder + eager-U-Net path with a misleading log.** `_trt_encoder` is assigned (line 320) before `TRTUNet(...)` is constructed (line 321). If the U-Net runner raises (stale/corrupt engine, shape-validation failure), `_trt_encoder` stays set while `_trt_unet` is None — the server catches and logs "using eager torch," but the encoder is still TRT. Output is correct (the TRT encoder ≡ normalize+encode), so this is low severity, but the fallback isn't clean and the log lies.

Fix: construct both runners into locals first, then assign together so attach is atomic —
```python
enc = TRTEncoder(enc_path, ...)
unet = TRTUNet(unet_path, ...)
policy._trt_encoder, policy._trt_unet = enc, unet
```

ros2_ws/src/brain/manipulation/manipulation/dp_trt.py
```diff
@@ -0,0 +1,343 @@
...
+        # Wait for the caller's stream (where inputs were produced) before copying and
+        # launching, so the engine can't read the buffers before the copies land.
+        self._stream.wait_stream(torch.cuda.current_stream())
+        with torch.cuda.stream(self._stream):
+            for name, val in inputs.items():
+                self._buffers[name].copy_(val)
```
vignesh-anand, line 248: **S3 (micro) — `global_cond` is re-copied into the bound buffer every denoise step.** `global_cond` is constant across all 16 U-Net calls in a replan, but `_execute` copies every entry of `inputs` (including `global_cond`) into its buffer each call. Negligible — a `(1, cond_dim)` copy ×16 — but if you ever want to shave it, bind `global_cond` once per replan and only copy `sample`/`timestep` per step. Noting for completeness, not blocking.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -795,6 +819,86 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim, n_action_steps_
...
+        otherwise DP runs eager torch.
+        """
+        try:
+            from manipulation.DP import load_dp_policy
+
+            self.get_logger().info(f"Loading Diffusion Policy checkpoint: {checkpoint_path}")
+            dp_overrides = {}
+            if num_inference_steps_override:
+                dp_overrides["num_inference_steps"] = num_inference_steps_override
+            policy, load_result = load_dp_policy(
```
vignesh-anand, line 839: **A — no `validate_action_dim` for DP; ACT fails loud, DP fails cryptic.** The ACT path calls `validate_action_dim(state_dict, action_dim, checkpoint_path)` (line 690) specifically to fail loudly when the metadata `action_dim` disagrees with the checkpoint's action head, before it can be baked into a cached engine. DP has no equivalent — if metadata `action_dim` is wrong, `load_dp_policy` builds a U-Net head of the wrong width and hits an opaque `load_state_dict` size-mismatch `RuntimeError` (size mismatches raise even under `strict=False`), surfacing as a generic "Failed to load Diffusion Policy." Not silently-wrong, but the operator loses ACT's clear diagnostic. Cheap to replicate against the checkpoint's inferred action head.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -795,6 +819,86 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim, n_action_steps_
...
+                f"Diffusion Policy loaded in {time.time() - load_start:.2f}s "
+                f"({param_count/1e6:.1f}M params), n_action_steps={policy.config.n_action_steps}, "
+                f"inference_steps={policy.config.num_inference_steps}"
+            )
+
+            # Build/attach TRT engines (encoder once-per-replan + U-Net per denoise step),
+            # the accelerated DP inference path. First load builds+caches (1-2 min); later
+            # loads reuse the cache. Falls back to eager torch on any failure (missing
+            # tensorrt/onnx, build error) so DP still runs, just slower.
+            try:
```
vignesh-anand, line 871: **B — no engine-cache fast path; DP builds the full eager policy on every load, even a cache hit.** ACT's key optimization (lines 729-736): on a cache hit it deserializes the engine directly and *skips* the eager model, `load_state_dict`, device transfer, and stats load — "TRTACTPolicy needs neither the eager model nor the loaded dataset_stats." Here `_load_dp_policy_for_behavior` unconditionally builds the eager `DiffusionPolicy` (random ResNet init + `load_state_dict` + stats) *before* `attach_engines`, so even when both engines are already cached DP pays full eager construction every load. Combined with eager weights never being freed after attach (see the D3 comment on `dp_trt.attach_engines`), DP is heavier than ACT in both load time and resident memory on the common path. Consider a cache-hit fast path that deserializes engines without the eager build.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -795,6 +819,86 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim, n_action_steps_
...
+                from manipulation.dp_trt import attach_engines
+                trt_start = time.time()
+                attach_engines(policy, checkpoint_path, self.device, precision="fp16",
+                               log=self.get_logger().info)
+                self.get_logger().info(f"DP TensorRT engines attached in {time.time() - trt_start:.2f}s (fp16)")
+            except Exception as e:
```
vignesh-anand, line 877: **C — eager fallback silently loads a non-real-time DP; ACT fails loud when TRT is unavailable.** ACT deliberately does *not* catch a TRT import/build failure — it propagates and fails the load, because "TRT stays required." Here we catch it and continue on eager torch. But per dp_trt's own numbers, eager U-Net is ~35-70 ms *each* ×16 steps ≈ 0.5-1.1 s per replan — hopelessly over the 40 ms/tick budget. So a TRT failure makes DP "load successfully" and then run at an unusable control rate, silently, whereas ACT refuses to load. For a robot that's arguably worse than failing fast — at minimum this warning should say the policy will not meet real-time, or the DP load should also treat TRT as required.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -795,6 +819,86 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim, n_action_steps_
...
+
+            # Warm-up (denoising is heavier than ACT — a couple of iterations).
+            dummy_batch = {
+                "observation.image_camera_1": torch.zeros(1, 3, 224, 224, device=self.device),
+                "observation.image_camera_2": torch.zeros(1, 3, 224, 224, device=self.device),
+                "observation.state": torch.zeros(1, 6, device=self.device),
+            }
+            warmup_start = time.time()
+            self.current_policy.reset()
+            for _ in range(3):
```
vignesh-anand, line 888: **E — the 3× warm-up only denoises once.** ACT warms with a single `select_action` (line 805), which is correct for it. Here the intent is "a couple of iterations" because denoising is heavier — but `select_action` fills the action queue on call 1 and only `popleft`s on calls 2-3, so **exactly one** `predict_action`/denoise runs. The extra two iterations do nothing, and the first real execution still eats any uncached autotune/lazy-init. To actually warm N denoises, call `predict_action` directly in the loop or `reset()` between iterations.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -641,6 +648,23 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim, n_action_steps_
...
             checkpoint_path = os.path.expanduser(checkpoint_path)
 
+            # Diffusion Policy: self-contained loader, separate from the ACT/TRT path
+            # below. Load its normalization stats here — the ACT path loads stats lazily
+            # (only on an engine-cache miss), but DP always needs them.
+            if policy_type == "dp":
+                stats_path = os.path.join(os.path.dirname(checkpoint_path), 'dataset_stats.pt')
+                dataset_stats = None
+                try:
+                    dataset_stats = torch.load(stats_path, map_location='cpu')
```
vignesh-anand, line 658: **F — raw `torch.load` for stats, vs ACT's shared `load_torch_file` helper.** The ACT path loads both the checkpoint and stats via `load_torch_file` (line 741) — the shared helper that handles mmap, `weights_only`, and consistent logging. DP uses raw `torch.load(stats_path, map_location='cpu')` here, a minor consistency/safety gap (ties to the `weights_only` note on `DP.load_dp_policy`) and a second stats-load code path to maintain. Use `load_torch_file` for parity.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -641,6 +648,23 @@ def _load_policy_for_behavior(self, checkpoint_path, action_dim, n_action_steps_
...
+                    self.get_logger().info("Dataset stats loaded")
+                except Exception as e:
```
vignesh-anand, line 660: **G — a stats load failure should abort the DP load here, not proceed to NaN.** This `except` warns and passes `dataset_stats=None` onward; `DP._NormBase` then leaves its mean/std buffers at the `float("inf")` sentinel → `Normalize` produces NaN → the robot commands NaN joint targets, and the load still returns `True` (see the sentinel comment on `DP.py`). Unlike ACT — where a cache-hit engine has stats baked in and doesn't need the file at all — DP *always* needs stats, so a failure here is unrecoverable. Fail the load on this exception (or assert finite normalize buffers after `load_dp_policy`) instead of warning and continuing.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -347,12 +347,16 @@ def _execute_learned_behavior(self, goal_handle, skill_dir, params: LearnedExecC
             n_action_steps_override = params.n_action_steps
             if n_action_steps_override is None and self.default_n_action_steps > 0:
                 n_action_steps_override = self.default_n_action_steps
```
vignesh-anand, line 349: Now that this fallback also feeds DP, the global `default_n_action_steps` default should be a DP-friendly value. Its current semantics are ACT-shaped ("0 = auto min(40, chunk_size)"), so any non-zero operator value tuned for ACT's chunk_size (40+) silently clamps to DP's `horizon` (16) — halving DP reactivity with a misleading log.

Two ways to make it DP-safe: either pick a shared default that's valid for both (≤ DP horizon, e.g. 8, matching DP's own default), or scope the fallback so it only applies on the ACT path and DP uses its own default. Pair this with the `n_action_steps` clamp-with-warning suggested on `DP.load_dp_policy` so an out-of-range value is at least visible.

.gitignore
```diff
@@ -107,4 +107,7 @@ data/.last_mode
 # arm_wave directory (local data, not tracked)
 arm_wave/
 
+# sock experiments (local scratch, not tracked)
+sock_experiments/
```
vignesh-anand, line 111: This `.gitignore` change is unrelated to Diffusion Policy — `sock_experiments/` is a local scratch dir with nothing to do with DP support. It's scope creep in this PR; please drop it and land it separately (or just keep it local). Ideally this PR stays limited to the DP feature so the diff is easy to reason about and revert.

ros2_ws/src/brain/manipulation/manipulation/dp_test_inference.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
```
vignesh-anand, line 1: These two standalone scripts (`dp_test_inference.py` + `dp_benchmark.py`) should be merged into one and moved under the package's test/debug location rather than shipped in `manipulation/manipulation/`.

They already share most of their scaffolding — `load_dp_policy` build, the dummy-obs dict, device selection, and a `predict_action` call — and differ only in what they do after: one times the denoise across a DDIM-step sweep, the other checks load key-match + output finiteness/shape/range. That's two subcommands on one shared setup, e.g.:

```
python -m ... dp_tools bench  [--steps 4,8,16,32] [--checkpoint ...]
python -m ... dp_tools check  --checkpoint ... [--stats ...]
```

with a single `build_policy` / `dummy_obs` — which also gives one place to keep the hardcoded obs schema in sync with `DP.load_dp_policy`.

Placement: put it where the package keeps tests/debug utilities (`manipulation/test/`, next to `test_config_validation.py`) rather than in the shipped package dir. The `check` path is a natural pytest smoke test (assert 0 missing/unexpected keys, assert `torch.isfinite` on the output — the finiteness gate the runtime load path is missing). And add the `SPDX-License-Identifier` + copyright header the other tools (`recorder_benchmark.py`, `act_trt.py`, `dp_trt.py`) carry.

vignesh-anand:
## Review summary — On-robot Diffusion Policy

Reviewed section by section (training-client tagging → config schema → `DP.py` → `dp_trt.py` → server integration → tooling). Left ~30 inline comments; this is the triage. Overall the port is careful and the isolation design is good — the ACT path is genuinely untouched, `select_action` reuse keeps the inference loop unchanged, the DDIM math checks out against diffusers, and the benchmark is honest about amortized-vs-sync cost. The concerns cluster around **failure modes that silently reach the arm** and **divergences from the ACT path** that aren't clearly intentional.

### 🔴 Merge-blockers (silent wrong/unsafe behavior on the robot)
- **`clip_sample=True` saturates z-scored actions** — actions are `MEAN_STD` normalized but `x0` is clamped to `[-1,1]` (no MIN_MAX mode exists), truncating every motion beyond ±1σ toward the mean. Confirm the training scheduler's `clip_sample` + action-norm before merge. `DP.py`
- **Missing/failed `dataset_stats` → NaN actions, load still returns success** — the `float("inf")` sentinel is never checked; a stats hiccup silently commands NaN joint targets. Should abort the load (or assert finite normalize buffers). `DP.py` + `manipulation_server` DP branch.
- **TRT-failure eager fallback loads a non-real-time DP silently** — eager U-Net ≈0.5–1.1 s/replan, hopelessly over the 40 ms budget, yet the load "succeeds." ACT treats TRT as required and fails fast; DP should at least warn it won't meet real-time. `_load_dp_policy_for_behavior`.

### 🟠 Should-fix before merge (correctness / operational / diverges from ACT)
- **Policy-type classifier has no fail-safe** — decided by a `dp_policy` filename prefix owned by an out-of-repo system, no `elif act_policy`/warn, defaults silently to ACT on any mismatch. `skill_manager.activate_run`.
- **No `validate_action_dim` for DP** — ACT fails loud on metadata/checkpoint action_dim disagreement; DP gives an opaque size-mismatch crash. `_load_dp_policy_for_behavior`.
- **`n_action_steps > horizon` silently under-fills the queue** + the ACT-shaped global `default_n_action_steps` feeds DP — clamp to `horizon - (n_obs_steps-1)` with a warning, and pick a DP-friendly default. `DP.predict_action` / server.
- **`num_inference_steps > num_train_timesteps` → `step_ratio=0` → garbage** — guard in `DDIMSampler` (covers config/benchmark/trt at once); schema has no upper bound.
- **No `dataset_stats` hash in the TRT engine cache key** — ACT hashes it because the encoder bakes stats; DP bakes obs stats too but omits it, so regenerated-in-place stats silently reuse a stale-normalization engine. Reuse `act_trt._dataset_stats_signature`. `dp_trt.engine_path_for`.
- **DP engines are never pre-built** — ACT builds at download time via `innate_training_node` (incl. a `--batch` sweep); DP builds inline at first skill activation, blocking ~1–2 min. `dp_trt._main`.
- **Eager weights never freed after attach + no cache-hit fast path** — ACT deserializes the engine and drops the eager model; DP always builds the full eager policy and keeps it resident alongside both engines → heavier in load time and memory every load. `dp_trt.attach_engines` / server.
- **Precision divergence** — ACT defaults fp32 (characterized RMSE); DP hardcodes fp16 with no characterization, over 16 compounding denoise steps. Recommend unifying both on fp16/bf16 (bf16 for the wider range) and measuring DP action RMSE.
- **No tests for the new schema fields** despite an established pattern in `test_config_validation.py`.

### 🟡 Nice-to-have / cleanup
- 3× warm-up only denoises once (queue fills on call 1). `_load_dp_policy_for_behavior`.
- No profiling telemetry on the DP runner → inference-profile reports `engine_ms=0`/`engine_ran=True` always. `dp_trt._TRTRunner`.
- Partial attach leaves a mixed TRT-encoder + eager-U-Net with a misleading "eager torch" log — construct both runners then assign together.
- DP uses raw `torch.load` (no `weights_only`, not the shared `load_torch_file`) in a few places.
- `num_inference_steps` is DP-only but silently ignored for ACT (a `model_validator` warn is the one cross-field check that belongs in the schema).
- `.gitignore` `sock_experiments/` is unrelated scope creep — drop it.
- Merge `dp_test_inference.py` + `dp_benchmark.py` into one `bench`/`check` tool under `manipulation/test/`; wire `check` as a pytest smoke test (which also becomes the missing finiteness gate). Add SPDX headers.
- Minor: mixed-checkpoint "highest step wins" family inference; stale docstrings (`activate_run` return shape, config comment hardcoding "16"); unused `Optional` import; `Normalize`/`Unnormalize` epsilon asymmetry.

**Bottom line:** the architecture and math are sound and the ACT path is safely isolated. The work before merge is mostly about **failing loud instead of silent** (clip/NaN/eager-fallback/action_dim) and **closing the gaps vs the ACT TRT path** (prebuild, cache-key stats hash, eager-weight release, precision). Happy to pair on the guards for the three blockers.

vignesh-anand:
G1 — BLOCKER: robot cannot submit a DP run. [node.py:56](vscode-webview://0trisvikefrj505prllkrgebg2jhi7kteuo3928ruqrcpk26034e/ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py#L56) hardcodes KNOWN_PRESETS = {"act-default"}. Any dp-default submission is rejected at _build_training_params with "unknown preset." The orchestrator, activation, and inference all understand DP — only this one line gates it. There's already a # TODO: fetch from discovery URL once available. Fix: add "dp-default" to the set (one line).

axelpey:
Closing the stale implementation against the older manipulation runtime; it also retains the unresolved missing-stats/NaN-action safety failure. Re-cut from current main if Diffusion Policy support is still wanted.

PR #475, opened 2026-06-25
feat(webapp): prompt mobile visitors to get the native app

theo-michel:
## What

When the robot's web cockpit (`webapp/`) is opened on a **phone or tablet**, show a dismissible popup nudging the visitor to download the native app, with a platform-correct link:

- **iOS** → TestFlight (`testflight.apple.com/join/YeChe4A7`)
- **Android** → APK (`cdn.innate.bot/innate-app-latest.apk`)

The links match what the docs hand out (`get-started/mars-quick-start.mdx`, `hackathon/robohacks.mdx`). The web app is built for a laptop; on the go the native app is the better surface (leader-arm teleop in the field), so this points mobile users there.

## How

- **`js/appPromo.js`** (new) — `maybeShowAppPromo(root)`: detects the platform from the user agent (including the iPadOS-as-Mac case via touch points), builds the popup, and gates on a `localStorage` flag so it shows once and never nags across page switches or repeat visits. No-op on desktop.
- **`js/shell.js`** — invoked from `initShell()`, the one hook every page already runs, so it covers all pages with the correct asset root.
- **`css/app.css`** — `.app-promo*` styles reusing the existing design tokens (panel / hairline / amber accent), fixed above the icon rail.
- **`tests/appPromo.test.js`** (new) — zero-dep node test for the detection logic (iPhone / Android / desktop-Mac / iPad), following the repo's `tests/` convention.

## Notes

- Dismissal is **permanent** (localStorage), not per-session — easy to switch to `sessionStorage` if we'd rather re-prompt each visit.
- The two URLs are hardcoded constants at the top of `appPromo.js`; update them there if the TestFlight/APK links change.

## Verification

- `node --check` clean; `tsc` reports no errors in the new/changed files.
- `node tests/appPromo.test.js` → 4 passed; existing `dynamixel.test.js` still green.
- Rendered both variants in a browser at phone size and confirmed the iOS/Android labels + correct hrefs, on-brand styling, and that "Continue in browser" removes the overlay.

### DIFF
--- a/webapp/js/appPromo.js
+++ b/webapp/js/appPromo.js
@@ -20,6 +20,9 @@ export function maybeShowAppPromo(root) {
   const platform = detectMobilePlatform();
   if (!platform || storageGet(DISMISS_KEY)) return;
 
+  // Remember what had focus so we can hand it back when the dialog closes.
+  const prevFocus = document.activeElement;
+
   const backdrop = document.createElement("div");
   backdrop.className = "app-promo-backdrop";
   backdrop.addEventListener("click", dismiss); // tap outside the card to dismiss
@@ -29,6 +32,7 @@ export function maybeShowAppPromo(root) {
   card.setAttribute("role", "dialog");
   card.setAttribute("aria-modal", "true");
   card.setAttribute("aria-label", "Get the Innate app");
+  card.tabIndex = -1; // focusable target so the dialog can be announced on open
   card.addEventListener("click", (e) => e.stopPropagation());
 
   const close = document.createElement("button");
@@ -70,6 +74,8 @@ export function maybeShowAppPromo(root) {
   backdrop.appendChild(card);
   document.body.appendChild(backdrop);
   document.addEventListener("keydown", onKey);
+  // Move focus into the dialog so screen-reader / keyboard users land on it.
+  card.focus();
 
   /** @param {KeyboardEvent} e */
   function onKey(e) {
@@ -80,6 +86,8 @@ export function maybeShowAppPromo(root) {
     storageSet(DISMISS_KEY, "1");
     document.removeEventListener("keydown", onKey);
     backdrop.remove();
+    // Hand focus back to wherever it was before the dialog opened.
+    if (prevFocus instanceof HTMLElement) prevFocus.focus();
   }
 }
 
@@ -90,11 +98,12 @@ export function maybeShowAppPromo(root) {
 export function detectMobilePlatform() {
   const ua = navigator.userAgent || "";
   if (/android/i.test(ua)) return "android";
-  // iPhone/iPod/iPad, plus iPadOS which reports itself as desktop Safari on a
-  // Mac — distinguished from an actual Mac by the touch screen.
+  // iPhone/iPod/iPad, plus iPadOS 13+ which reports itself as desktop Safari
+  // ("Macintosh" UA) — distinguished from an actual Mac by the touch screen.
+  // (navigator.platform would work too but is deprecated.)
   const isIOS =
     /iphone|ipad|ipod/i.test(ua) ||
-    (navigator.platform === "MacIntel" && navigator.maxTouchPoints > 1);
+    (/macintosh/i.test(ua) && navigator.maxTouchPoints > 1);
   return isIOS ? "ios" : null;
 }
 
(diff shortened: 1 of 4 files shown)


webapp/js/appPromo.js
```diff
@@ -0,0 +1,118 @@
...
+ * Exported for unit testing the gating logic. @returns {MobilePlatform | null} null on desktop / unknown.
+ */
+export function detectMobilePlatform() {
+  const ua = navigator.userAgent || "";
+  if (/android/i.test(ua)) return "android";
+  // iPhone/iPod/iPad, plus iPadOS which reports itself as desktop Safari on a
+  // Mac — distinguished from an actual Mac by the touch screen.
+  const isIOS =
+    /iphone|ipad|ipod/i.test(ua) ||
+    (navigator.platform === "MacIntel" && navigator.maxTouchPoints > 1);
```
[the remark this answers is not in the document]
theo-michel: Fixed in 078877cb — dropped `navigator.platform` in favor of the `Macintosh` UA string + `maxTouchPoints` check (exactly the suggested form). Test stub updated to match; still covers iPhone/Android/desktop-Mac/iPadOS.

webapp/js/appPromo.js
```diff
@@ -0,0 +1,118 @@
...
+  const stay = document.createElement("button");
+  stay.type = "button";
+  stay.className = "app-promo-stay";
+  stay.textContent = "Continue in browser";
+  stay.addEventListener("click", dismiss);
+
+  card.append(close, logo, title, blurb, cta, stay);
+  backdrop.appendChild(card);
+  document.body.appendChild(backdrop);
+  document.addEventListener("keydown", onKey);
```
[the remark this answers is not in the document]
theo-michel: Fixed in 078877cb — on open the dialog now moves focus to the card (`tabindex=-1`, so VoiceOver announces its `aria-label`/content rather than just the ✕), and on close it restores focus to the previously-focused element. Verified end-to-end with an iPadOS-spoofed UA: `document.activeElement` is the `role="dialog"` card after open, and `BODY` after dismiss. I skipped a full focus trap since `Esc` and tap-outside already dismiss this two-button promo.

webapp/js/appPromo.js
```diff
@@ -0,0 +1,127 @@
...
+  const prevFocus = document.activeElement;
+
+  const backdrop = document.createElement("div");
+  backdrop.className = "app-promo-backdrop";
+  backdrop.addEventListener("click", dismiss); // tap outside the card to dismiss
+
+  const card = document.createElement("div");
+  card.className = "app-promo";
+  card.setAttribute("role", "dialog");
+  card.setAttribute("aria-modal", "true");
```
vignesh-anand, line 33: **a11y: `aria-modal="true"` with no focus trap or inert background.** The dialog tells assistive tech the background is inert, but the rail and `#stage` behind the backdrop stay focusable — the only key handler (line 82) catches Escape, not Tab. A keyboard/screen-reader user can Tab straight out of the card into the still-live UI behind it. Fix: trap focus within the card, or set `inert`/`aria-hidden` on the rail + `#stage` while the dialog is open (and restore on dismiss). Minor for a dismissible promo, but the modal claim is currently unbacked.

webapp/js/appPromo.js
```diff
@@ -0,0 +1,127 @@
...
+
+  const blurb = document.createElement("p");
+  blurb.className = "app-promo-blurb";
+  blurb.textContent =
+    "The web app is built for a laptop. On your phone, the Innate app is the better way to drive — including leader-arm teleoperation in the field.";
+
+  const cta = document.createElement("a");
+  cta.className = "app-promo-cta";
+  cta.href = platform === "ios" ? IOS_APP_URL : ANDROID_APP_URL;
+  cta.target = "_blank";
```
vignesh-anand: **Android `.apk` CTA opens via `target="_blank"`.** On Chrome for Android, a `_blank` link to a downloadable file commonly spawns an empty tab that lingers after the download starts. Consider dropping `target="_blank"` for the Android branch (let the APK download in place); iOS TestFlight still wants `_blank`. e.g. set `cta.target` only when `platform === "ios"`.

webapp/js/appPromo.js
theo-michel: It's for assistive tech doesn't matter here

webapp/js/appPromo.js
theo-michel: Done in 79c128b9 — `target="_blank"` (and `rel="noopener noreferrer"`) are now set only on the iOS branch, so the Android `.apk` downloads in place without spawning a lingering empty tab on Chrome for Android.

theo-michel: fixed

PR #479, opened 2026-06-26
feat: make TTS voice a single global setting with a voice picker

theo-michel:
## Problem
The TTS voice lived as **two** separate ROS params — `brain_client_node.cartesia_voice_id` (the audible chat-TTS path) and `input_manager_node.cartesia_voice_id` (the realtime-voice path). Operators had to set both; setting only one left the two speech paths on different voices, and setting only `input_manager_node` had no audible effect at all. This bit us in practice.

## Change
**1. One real setting, not two.** The voice is now a single `/**` global ROS param. Both nodes declare `cartesia_voice_id`, and `settings.yaml` layers last in each launch's `parameters` list, so the `/**` value overrides the per-node launch defaults on both. The two per-node stanzas are dropped from `settings.yaml.template`.

Verified before relying on it:
- **ROS 2 Humble precedence**: a `/**` param loaded last (as `settings.yaml` always is) overrides a node-specific launch param — last-loaded wins regardless of node-name-vs-wildcard specificity.
- **Regenerator**: setting the voice through the proxy settings store produces a clean `/**: ros__parameters: cartesia_voice_id: …`, parses correctly, with no per-node blocks.

**2. Voice picker.** Added an optional `options` list to a catalog knob, rendering a `<select>`. The picker offers the default (Katie) plus five Cartesia stock voices. An off-list id set over SSH still round-trips as a "Custom" option.

## Files
- `config/settings.yaml.template` — voice → `/**`; per-node stanzas removed
- `webapp/js/settings/catalog.js` — `options` schema field; single `/**` voice knob with 6 choices
- `webapp/js/settings/main.js` — `<select>` rendering; friendly default label

## Notes
- Voice is read at node startup — **restart required** to apply (no live reload).
- The OpenAI realtime/transcribe model fields still have the same per-node duplication; not touched here.

### DIFF
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -37,6 +37,8 @@
 #     inflation_layer:              # costmap obstacle inflation (every costmap)
 #       inflation_radius: 0.3       # metres of inflation around obstacles
 #       cost_scaling_factor: 10.0   # higher = cost decays faster with distance
+#     # TTS voice for BOTH speech paths (brain chat-TTS + realtime voice loop)
+#     cartesia_voice_id: "9fdaae0b-f885-4813-b589-3c07cf9d5fea"
 
 # ── Teleop: joystick ──────────────────────────────────────────────────
 # joystick_controller:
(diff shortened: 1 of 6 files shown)


karmanyaahm (dismissed):
@theo-michel can we add a little tooltip or ? icon next to the setting to tell people there are hundreds of options they can pick on cartesia, and link to https://play.cartesia.ai/voices . So they know they can pick their own?

PR #483, opened 2026-06-30
manipulation: replace sensor-availability gate with clean-room reimplementation

theo-michel:
## Summary
The pre-execution sensor-readiness gate in `_execute_learned_behavior` (`manipulation_server.py`) is removed and re-implemented clean-room. The gate blocks before any motion until all required sensors (both cameras + arm joint state) are publishing, so policy inference never runs on empty/stale buffers, failing fast with a USB-C hint if they don't arrive within 2s.

This is a round-trip, in two commits:
1. **Remove** the original wait loop (introduced in #374) — it came from an external contributor whose license we do not hold.
2. **Re-implement** the same behavior clean-room (written from the spec, not the original source).

## What changed in the reimplementation
- The wait lives in a new `_wait_for_sensors()` helper returning a bool; the call site is a single guard clause.
- Uses a **monotonic** deadline (`time.monotonic()`) instead of wall-clock `time.time()`, so an NTP/clock jump can't fire the timeout early or hang the wait.
- Reuses the existing in-house `_check_sensor_availability()` predicate (authored internally, untouched by the removal).

## Contract (unchanged from original behavior)
- 2.0s timeout, 0.1s poll interval (10 Hz).
- Already-ready → falls through immediately, no sleep.
- AND across all required inputs; any one missing keeps waiting.
- Single info log on success, single error log on timeout; failure returns `("FAILURE", <user message>)`.

### DIFF
--- a/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
+++ b/ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
@@ -354,19 +354,6 @@ class ManipulationServer(Node):
             ):
                 return "FAILURE", f"Failed to load policy from {checkpoint_path}"
 
-            # Wait for sensor data to arrive after subscription creation
-            sensor_wait_deadline = time.time() + 2.0
-            while not self._check_sensor_availability():
-                if time.time() > sensor_wait_deadline:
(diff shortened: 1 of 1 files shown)


karmanyaahm:
tested, lgtm

vignesh-anand (approved):

PR #484, opened 2026-06-30
On-robot policy rollout evaluation: auto-stop, eval datasets, profiling UI

theo-michel:
## Overview

Tooling to **run and evaluate learned ACT policy rollouts on the robot**, end to end. A learned skill can now terminate on its own, a rollout can be captured as a judged eval-dataset episode with its full profile trace, and the good ones can be promoted into a training dataset — all driven from a reworked Profiling page. Supporting changes: a client-side webapp router, debugging-page improvements, and a longer episode cap.

## Auto-stop for learned skills

Learned ACT policies never terminated on their own — `_execute_learned_behavior` just kept requesting action chunks, so the only reliable exit was the `duration` cap (default 120 s). The pre-existing single-frame `progress > progress_threshold` check was effectively dead (default threshold `2.0` is unreachable for a ~0→1 progress head).

- **`manipulation/auto_stop.py`** *(new)* — `LearnedStopDetector` + `StepSignals`, pure stdlib (ROS-free). Ends a skill early when EMA-smoothed **progress** crosses `progress_threshold`, or via an **engage-then-settle** stability rule (progress must first dip below `engage_below`, then hold `>= stable_min` for `stable_seconds`) for heads that saturate high both before and after the task. `min_duration` floor and the `duration` hard cap always apply.
- **`manipulation_server.py`** — `_run_inference_once` returns `StepSignals(progress)` for the detector and tracks `_prev_action_np` unconditionally; the learned loop drives the detector and reports the real stop reason. Motion signals (`arm_delta`/`base_speed`) are still computed for the profiler's jerk charts.
- **`config_validation.py`** — opt-in `LearnedExecCfg` auto-stop knobs behind a single `auto_stop` switch; unset knobs fall back to a tuned engage-then-settle default.
- **`skills/catalog.py`** / **`skills/loader.py`** — new-skill template surfaces the knobs; loader plumbs them through.

## Rollout capture & eval datasets

- **Episode provenance** — recorder stamps how an episode was produced (`source`: teleop / rollout / replay) and, for rollouts, the driving `policy`. New `NewEpisode.srv`; `SetEpisodeOutcome.srv` gains failure-mode `tags`. (`recorder_node`, `task_manager`.)
- **Eval datasets** — rollout captures land in dedicated `type == "eval"` datasets so evaluation runs never mix into a skill's training data.
- **Copy-to-training** — new `CopyEpisode.srv` + `task_manager` copy an eval episode (h5 + per-camera mp4s + profile trace + raw data) into a training dataset under its next id, carrying provenance/outcome/tags; the source episode is untouched. Path comparison is canonicalized to tolerate spelling differences across service requests.
- **Profile trace persistence** — **`dataset_encoder/profile_recorder_node.py`** *(new)* tees the per-step inference-profile stream to `episode_<id>_profile.jsonl` next to the HDF5 while an episode records, so every saved rollout carries its profiling trace for offline evaluation.

## Profiling page

- Skill picker + **Run/Stop** to drive rollouts, a unified rollout loop, and a **reboot-arm** control.
- Headless robot-side **episode capture** for rollouts (`episodeCapture.js`) and an in-memory **eval-session tally** (`evalSession.js`) that reads as a running score.
- Auto-stop tuning stats surfaced for reading thresholds off real runs.

## Datasets page

- Episode list reworked into training vs. evaluation flavors (provenance badges, per-policy filter chips, success-rate summary, outcome dropdown with an explicit Unlabeled state).
- Shared **episode context menu** (`episodeMenu.js`) for the promote-to-training flow, and an inline **profile-trace chart** (`profileTrace.js`) in the player for rollout episodes.

## Webapp router refactor

- Client-side **router** (`router.js`) replacing per-page `index.html` files; single app shell with instant page switches, SPA fallback + content-hash ETags in the proxy (`https_server.py`), and a `/episode/profile` media route (`media_routes.py`).

## Debugging page

- Click a log source to filter and select it; expand the tree to the selected record; copy-logs buttons.

## Recorder

- Episode cap raised to **10 minutes** (18000 timesteps).

## Misc

- `mars_control/app.cpp` — OS version reporting only considers release-version tags (`x.y.z` / `-rcN`), ignoring unrelated tags like `sim-assets-*`.

## Backwards compatibility

Auto-stop is off by default (skills run to the `duration` cap unchanged until they opt in). Provenance defaults to `teleop` for legacy episodes; episodes without a profile trace simply hide the chart.

## Testing

- `ruff check` + `ruff format --check`: clean.
- No unit tests included.
- Auto-stop thresholds are not yet tuned on hardware — the on-robot tuning pass (log `progress` and motion on the Profiling page to set thresholds from real data) is the natural follow-up.


theo-michel:
Second review pass (ddf55e6) — disposition of the findings:

- **Idle dwell through the `min_duration` floor** (also flagged by greptile): fixed. Demos often start with a still beat, and with the default `progress_gate=0` the detector could fire "idle" the instant the floor expired. The dwell now only starts counting after the floor.
- **NaN robustness** (new finding): a diverging policy emits NaN actions; `NaN >= eps` is False, so NaN read as "still" and a diverged run could end as a *successful* idle stop. Non-finite motion now counts as moving, and NaN progress is skipped by the EMA instead of poisoning it permanently.
- **`idle_base_eps` tuning guidance** (greptile): the profiler now publishes `base_speed` (the exact quantity the detector checks — `base_jerk` is the command delta, a different metric), and the comment points at it.
- **Stale module reference in config comment** (greptile): fixed.
- **`_idle_since` across transient inference failures** (greptile P1): intentionally left as-is. During a failure window no new commands are published, so the robot genuinely holds its last command; letting the dwell span the gap is consistent. If the policy jumps on recovery, the first post-recovery `arm_delta` (diffed across the gap) resets the timer.

12/12 unit tests pass; ruff clean.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -404,24 +419,22 @@ def _execute_learned_behavior(self, goal_handle, skill_dir, params: LearnedExecC
...
-                    if progress > progress_threshold:
+                    stop, reason = stop_detector.update(signals, elapsed_time, loop_start)
+                    if stop:
                         early_termination = True
-                        self.get_logger().info(
-                            f"Early termination triggered! Progress: {progress:.4f} > {progress_threshold}"
-                        )
+                        stop_reason = reason
+                        self.get_logger().info(f"Early termination triggered! {reason}")
                         break
```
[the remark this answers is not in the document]
theo-michel: Fixed in eccceda5 — the loop now calls `stop_detector.note_gap()` when `_run_inference_once()` returns None, so the idle dwell only counts observed stillness and can't silently span a failure window.

ros2_ws/src/brain/manipulation/manipulation/config_validation.py
```diff
@@ -149,6 +149,29 @@ class LearnedExecCfg(_BaseExecCfg):
     # checkpoint is loaded; the schema only enforces ``>= 1``.
     n_action_steps: int | None = Field(None, ge=1)
 
+    # --- Auto-stop tuning ---------------------------------------------------
+    # The learned loop ends early when one of these fires, otherwise at ``duration``
+    # (the always-on hard cap). Every field below defaults to a no-op, so a skill
+    # behaves exactly as before until it opts in. See
+    # ``manipulation_server._LearnedStopDetector`` for how they combine.
```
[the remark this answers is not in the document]
theo-michel: Fixed in ddf55e67 — comment now points at `manipulation.auto_stop.LearnedStopDetector`.

ros2_ws/src/brain/manipulation/manipulation/auto_stop.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if progress is not None and progress > self.progress_threshold:
+            return True, f"progress {progress:.4f} > {self.progress_threshold}"
+
+        if self.idle_seconds > 0 and self._idle_since is not None:
+            held = now - self._idle_since
+            gate_ok = self.progress_gate <= 0 or (progress is not None and progress >= self.progress_gate)
+            if held >= self.idle_seconds and gate_ok:
+                note = "" if self.progress_gate <= 0 else f", progress {progress:.4f} >= {self.progress_gate}"
+                return True, f"idle for {held:.1f}s (>= {self.idle_seconds}s){note}"
```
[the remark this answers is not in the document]
theo-michel: Fixed in ddf55e67 — the idle dwell no longer accumulates inside the `min_duration` floor; the robot must hold still for a full `idle_seconds` after the floor. Pinned by `test_idle_dwell_starts_after_min_duration_floor`.

ros2_ws/src/brain/manipulation/manipulation/config_validation.py
```diff
@@ -149,6 +149,29 @@ class LearnedExecCfg(_BaseExecCfg):
...
+    # smaller = smoother but slower to react. Applied to both the progress-threshold
+    # and progress-gate checks.
+    progress_ema_alpha: float = Field(1.0, gt=0, le=1.0)
+    # Stop once the robot has held still this long (s); 0 = idle stop disabled.
+    idle_seconds: float = Field(0.0, ge=0)
+    # Per-step "not moving" thresholds: arm = L2 norm of the commanded joint-target
+    # change (rad), base = |linear.x| + |angular.z| (m/s + rad/s). Read them off the
+    # Profiling page's ``arm_jerk`` and base velocity while the skill holds its end pose.
+    idle_arm_eps: float = Field(0.01, gt=0)
+    idle_base_eps: float = Field(0.01, gt=0)
```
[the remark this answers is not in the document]
theo-michel: Fixed in ddf55e67 — the profiler now publishes `base_speed` (`|linear.x| + |angular.z|`, the exact quantity the detector checks) alongside `base_jerk`, and the tuning comment points at it.

ros2_ws/src/brain/manipulation/manipulation/auto_stop.py
```diff
@@ -0,0 +1,141 @@
...
+            self._idle_since = now
+
+        if progress is not None and progress > self.progress_threshold:
+            return True, f"progress {progress:.4f} > {self.progress_threshold}"
+
+        if self.idle_seconds > 0 and self._idle_since is not None:
+            held = now - self._idle_since
+            gate_ok = self.progress_gate <= 0 or (progress is not None and progress >= self.progress_gate)
+            if held >= self.idle_seconds and gate_ok:
+                note = "" if self.progress_gate <= 0 else f", progress {progress:.4f} >= {self.progress_gate}"
```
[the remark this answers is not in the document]
theo-michel: Fixed in 162b94d3 — a cross-field validator on `LearnedExecCfg` now rejects `progress_gate > 0` with `action_dim < 10` at accept time (clear message instead of a silently-dead idle stop). Also added the missing config-schema tests for all six knobs and registered `test_auto_stop.py` in CI's pytest list.

theo-michel:
## Phase-0 tuning result → new stop mechanism (pushed 8ae0512f)

Captured a real on-hardware run off the profiling tab (1692 samples, ~69.5s). Key finding: **the progress head saturates near its max at both ends of the run.** The opening ~20s (before the arm engages) look identical to the converged final tail — high, stable, low-jerk — with the actual work (progress dipping to 0.4–0.7) in the middle.

Consequences, verified by replaying the samples through the detector:
- Progress-threshold and *bare* progress-stability both false-positive early — "smoothed progress ≥0.93 for 3s" fires at **~8s**, not the tail.
- Idle stop can't fire at all: `still_longest_s=0.7` (per-chunk replan jerk spikes break any dwell with eps 0.01).

So I added a third opt-in stop: **engage-then-settle**. Require smoothed progress to first dip below `engage_below`, *then* stop once it holds ≥ `stable_min` for `stable_seconds`. On this run the recommended config fires at **65.3s — 4.2s before the end**, matching where the skill actually completed.

**Recommended per-skill config for this checkpoint** (`metadata.json → execution`):
```json
{
  "min_duration": 5.0,
  "progress_ema_alpha": 0.3,
  "engage_below": 0.75,
  "stable_min": 0.93,
  "stable_seconds": 3.0
}
```
`stable_min` sits just under the settled peak (`progress_max=1.0157`); `engage_below` sits above the active-phase dips. Idle knobs left at 0 (not usable here). All three new knobs default to no-ops; `stable_seconds>0` requires a progress head (`action_dim>=10`).

theo-michel:
## Added: one-button rollout capture + saved-episode profiling (bee54ddc)

Builds the profiling/eval capture layer on top of this PR's profiling tab:

**Profiling page → "Capture episode"**: saves the current rollout as a full dataset episode — video + trajectory via the existing recorder/encoder pipeline, **plus** the complete per-step inference profile persisted as `data/episode_<id>_profile.jsonl`. Then Save/Discard + ✓/✗ outcome label. Episodes land in a dedicated **"Policy Rollouts"** skill dataset (auto-created) so eval runs never mix into a real skill's training data.

**How** (all reuse-shaped):
- profile records now carry the **full raw action vector** (pre-scale, all dims incl. progress/termination heads) — one line in `manipulation_server`
- new `profile_recorder` node in the `dataset_encoder` package, mirroring the encoder's RecorderStatus-following pattern; subscribes to the profile stream **only while an episode is recording**, so the subscriber-gated profiler stays free when idle
- `GET /episode/profile` route beside `/episode/joints`
- episode player gains an **inference profile** section (progress / arm motion / base speed / step ms / disagreement overlay + summary) next to the video — saved rollouts are compared on the Datasets page like any episodes

**Tested**: profile_recorder lifecycle (subscribe→pause→save→cancel, metadata-based id resolution, skill-context line) against real rclpy+brain_messages in the CI image; ruff/pre-commit/JS-syntax clean. Not yet exercised on hardware — needs one live rollout + button press.

Note: `RecorderStatus.episode_number` is a running count, not the h5 slot id — the node resolves the real id from `dataset_metadata.json` after save.

webapp/js/datasets/profileTrace.js
```diff
@@ -0,0 +1,132 @@
...
+
+  const plot = document.createElement("div");
+  plot.className = "telemetry-plot";
+  plot.style.height = "120px";
+  const svg = document.createElementNS(SVG_NS, "svg");
+  svg.setAttribute("viewBox", "0 0 1000 200");
+  svg.setAttribute("preserveAspectRatio", "none");
+  svg.setAttribute("class", "telemetry-svg");
+
+  for (const [key, [name, color]] of SERIES) {
```
DavidDobas, line 98: Each series here is independently normalized to its own min/max ([min = Math.min(...vals)](webapp/js/datasets/profileTrace.js) / `span = max - min`), plotted into a shared 0–200 viewBox with no axis, gridlines, or tick labels. That means the vertical position of a line carries no absolute meaning across series — "progress" and "step ms" can look like they move in lockstep purely because both get stretched to fill the same band, even if their real ranges are wildly different.

There's also no hover/tooltip on the path — it's a static SVG with no per-point markers or mouseover handler, so there's no way to read an actual (step, value) pair off the chart, only eyeball the shape.

This matters beyond just readability: `config_validation.py`'s own guidance for tuning `auto_stop` is "read the progress trace off the profiling tab to pick `stable_min`/`engage_below`" — but as built, this chart can't give you an actual progress value at a point, only a normalized squiggle. Worth adding an axis (at least for the `progress` series, since it has a meaningful absolute 0–1 range) and/or a hover tooltip showing raw values, otherwise the recommended tuning workflow isn't really practical off this UI.

ros2_ws/src/brain/manipulation/manipulation/config_validation.py
```diff
@@ -63,6 +63,19 @@ class BehaviorConfigError(ValueError):
 
 KNOWN_BEHAVIOR_TYPES = ("learned", "poses", "replay")
 
+# Recommended engage-then-settle auto-stop config, applied to any knob a skill
+# leaves unset once ``auto_stop`` is enabled (so flipping the one switch yields a
+# working stop). Tuned on phase-0 pick-sock profiler data: smoothed progress must
+# first dip below ``engage_below`` (the policy engaged), then hold >= ``stable_min``
+# for ``stable_seconds``. See ``manipulation.auto_stop.LearnedStopDetector``.
+_AUTO_STOP_DEFAULTS = {
```
DavidDobas, line 71: The `auto_stop` tuning knobs (`engage_below`, `stable_min`, `stable_seconds`, `min_duration`, `progress_ema_alpha`) aren't documented anywhere outside comments like this one — `manipulation` has no README, and the skills README doesn't cover the `execution`/`behavior_config` fields at all.

Worth a short doc (or README section) covering:
- What each knob does and how they combine (engage-then-settle vs. the single-frame `progress_threshold`).
- That `engage_below` is a one-way latch for the run (see `auto_stop.py`'s `_engaged` flag) — once progress dips below it, "engaged" stays true for the rest of the rollout, even if the dip was noise rather than real task start.
- The recommended tuning workflow: read the profiling-page progress trace to pick `stable_min` (just under the settled peak) and `engage_below` (above the active-phase dips) per skill, rather than trusting these defaults blindly.
- That these defaults are tuned on one skill (phase-0 pick-sock) and not yet validated on hardware for others.

Right now this knowledge only lives in this dict and its neighboring comments, so anyone authoring a new skill's metadata.json has to go read the implementation to use `auto_stop` correctly.

ros2_ws/src/brain/manipulation/manipulation/auto_stop.py
```diff
@@ -0,0 +1,106 @@
...
+        # fires once smoothed progress holds >= ``stable_min`` for ``stable_seconds``.
+        # ``engage_below <= 0`` arms immediately (bare stability, no dip required).
+        if self.stable_seconds > 0 and self.stable_min > 0 and ema is not None:
+            if self.engage_below <= 0 or ema < self.engage_below:
+                self._engaged = True
+            if self._engaged and ema >= self.stable_min:
+                if self._stable_since is None:
+                    self._stable_since = now
+                held = now - self._stable_since
+                if held >= self.stable_seconds:
```
DavidDobas, line 101: On false positives (stopping before the task actually succeeds): this firing condition is guarded only by threshold tuning (`stable_min` + `stable_seconds`), which doesn't fully cover it. The progress head can be confidently wrong, not just noisy — a sub-goal state that visually/proprioceptively resembles the final state could read as high, stable progress without the task having succeeded. Raising `stable_min` just moves the bar higher on the same self-reported signal; it doesn't add any independent corroboration that the task actually finished.

Which is why I wanted to ask: why was the idle-stop mechanism (`a7b56b0b`) removed? That gave exactly this kind of corroboration — requiring the arm/base to actually be physically still (`arm_delta`/`base_speed` near zero) in addition to progress. If progress reads "done" but the arm is still mid-motion, that's a strong signal the progress reading is wrong. Was it purely "not on by default so not worth the surface area," or was there a correctness/reliability issue with the idle-stop path itself? If it's the former, it might be worth keeping as an optional corroborating check even if off by default, given it's the only thing in this design that catches a confidently-wrong progress head rather than just a noisy one.

For now this is probably an acceptable tradeoff given the usage context (supervised rollout evaluation on the Profiling page, human watches each run, an early stop just means rerunning) — but worth being explicit that it's a real gap and not a solved problem if `auto_stop` ever runs unsupervised or on a skill with a poorly-calibrated progress head.

webapp/js/profiling/main.js
```diff
@@ -341,7 +374,7 @@ function scaffold(host, key, build) {
  * @param {HTMLElement} host
  * @param {(s: Sample) => number|undefined} accessor
  * @param {string} color
- * @param {{lo?: number, hi?: number}} fixed  optional fixed y-range (else auto)
+ * @param {{lo?: number, hi?: number}} fixed  fixed y-range (falls back to the data range if unset)
  */
 function renderLine(samples, host, accessor, color, fixed) {
   const pts = samples.slice(-MAX_PLOT_POINTS).map(accessor).filter((v) => typeof v === "number");
```
DavidDobas: Same class of issue as the episode-player chart, but worse here since this is the live view: `pts` is a hard 400-point sliding window (`samples.slice(-MAX_PLOT_POINTS)`) — once a rollout exceeds that, earlier samples aren't just scrolled off-screen, they're sliced out of `pts` entirely, so there's no way to scroll/zoom back to see the start of a longer recording after the fact.

The x-axis also carries no time information: each point is placed by array index (`x = (i / (pts.length - 1)) * W`) rather than elapsed time, and the only x-axis label is `now ${last.toFixed(2)}` at line 416 — which is actually the last *value*, not a timestamp, despite being captioned "now". So there's no way to tell how much time a chart window spans, or when along the run a spike happened, and (same as the other chart) no hover to read an exact (time, value) pair.

Given `config_validation.py`'s guidance is to tune `auto_stop`'s `stable_min`/`engage_below`/`stable_seconds` off exactly this kind of chart, not being able to correlate a value with a real timestamp makes that tuning workflow pretty imprecise in practice.

webapp/js/profiling/rolloutControl.js
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  function toReady(flash) {
+    state = "ready";
+    sync(flash);
+    if (flash) setTimeout(() => sync(), 1600);
+  }
+
+  /** One failure funnel: whatever step died, don't leave a half-open episode. */
+  function onError(err) {
```
DavidDobas, line 166: Field report from on-robot testing: when the rollout start chain fails, the operator gets only the generic "Failed — see console" flash, and `console.error` here goes to the *browser devtools* console — which you typically don't have open while standing at a robot. In my case the actual failure was `skills_server.py` crash-looping on the robot (rclpy `InvalidHandle` → zenoh Rust panic → SIGABRT, respawned by launch), so every `capture.start()` call was timing out for ~2 minutes — and the only observable trace anywhere was the recorder logging "No active or stopped episode to cancel" from this funnel's `capture.discard()`. That made it look like the webapp was broken when it was actually a backend blip.

Two suggestions:

1. Surface `err.message` in the UI instead of (or alongside) the generic flash — e.g. "Service call /brain/create_physical_skill timed out". A timeout naming the service immediately points at the right backend node; "Failed — see console" points nowhere.
2. Consider making the start chain a bit more resilient: `runRollout("eval")` makes three sequential service calls (create_physical_skill → activate_physical_primitive → new_episode) before the skill goal is even sent, and any one of them timing out during a backend restart fails the whole run. This matters extra for **auto-continue**, which will happily keep slamming into a restarting skills server every 5 seconds — each attempt failing with the same opaque flash. A single retry with backoff (or at least stopping the auto-continue loop after a failed start) would make eval sessions survive a server respawn.

Relatedly, the crash itself is worth knowing about for anyone testing this PR: both skills_server crashes I hit came immediately after a skill *cancellation* — i.e., the exact run → Stop → run-again cycle this page's rollout loop (and especially auto-continue) drives repeatedly. The crash is in rclpy/rmw_zenoh teardown, not this PR's code, but this PR's workflow is what makes it easy to hit.

DavidDobas (approved):
Up to you if you find comments relevant

webapp/js/profiling/rolloutControl.js
```diff
@@ -0,0 +1,317 @@
...
+  /** @param {(() => Promise<void>) | (() => void)} fn */
+  async function guarded(fn) {
+    if (busy) return;
+    busy = true;
+    try {
+      await fn();
+    } catch (err) {
+      onError(err);
+    } finally {
+      busy = false;
```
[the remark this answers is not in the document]
theo-michel: Fixed in 5f53d4cb — `runRollout` re-checks `destroyed` after `await capture.start(...)`: it discards the episode again (destroy()'s own discard can race the episode-open service call, so this second discard catches an episode that landed after it) and returns without ever sending the skill goal. The "never leave a skill running headless" invariant now holds through the starting window too.

webapp/js/profiling/rolloutControl.js
theo-michel: Fixed in 5f53d4cb — failures now land in a persistent error line on the page showing the real message, so your case reads `Service call /brain/create_physical_skill timed out after 10000ms` and points straight at the skills server instead of at the webapp. When auto-continue is checked the line appends "— auto-continue paused". Also named the previously-bare `create failed`/`activate failed` fallbacks.

On the loop slamming a restarting server: auto-continue already can't do that — the countdown is only ever armed from a successful save, so the first failed start ends the session (`onError` now documents that invariant). Given that, I left retry-with-backoff out: a respawn window like the one you hit (~2 min) outlasts any single retry, and a retry doubles time-to-message on every genuine failure — with the error line naming the sick node, rerunning is one click. Happy to add it if you still want it.

And thanks for the field report on the rclpy/rmw_zenoh teardown crash — it's real, pre-existing, and this page's run→Stop→run cycle is indeed the easiest way to hit it; I'm spinning it off to be investigated separately rather than burying it here.

webapp/js/datasets/profileTrace.js
theo-michel: Fixed in 6eeb477d — progress is now pinned to an absolute 0–1 scale (extended if the head overshoots 1.0) with labeled gridlines at 0 / 0.5 / 1, so its height is a real value you can set `stable_min`/`engage_below` against. The other series have no shared unit, so they stay shape-normalized — but the chart now has the same hover readout as the joints chart above it: guide line + exact time/step + the raw value of every series under the cursor. The summary line also gains the run duration.

webapp/js/profiling/main.js
theo-michel: Fixed in 6eeb477d — the quality charts now draw the whole retained run instead of the last-400 slice: the record is decimated to the point budget with min/max buckets (each bucket keeps its extremes, so spikes survive), x is elapsed time labeled `0s → Ns`, and hover reads the exact (time, value) pair off the frozen run — the missing piece of the tuning workflow. The mislabeled caption is now `now <value>` flanked by real time labels, so it reads as the right-edge value rather than a timestamp.

The latency timeseries intentionally stays a sliding window (its per-step engine-ran markers don't survive decimation, and recent cadence is what it's for), but it now labels the window explicitly (`last 400 steps`) instead of implying it's the whole run.

ros2_ws/src/brain/manipulation/manipulation/config_validation.py
theo-michel: Added in cf87487a — `ros2_ws/src/brain/manipulation/README.md` now has an auto-stop section covering exactly this list: every knob and how they combine (engage-then-settle vs. the legacy single-frame threshold, `duration` always the backstop), the `engage_below` one-way-latch caveat, the read-thresholds-off-the-profiling-charts workflow (which the chart fixes in 6eeb477d make practical — hover now gives raw values), the phase-0-pick-sock-only provenance of the shipped defaults, and a known-limitations list. The `auto_stop` comment in `config_validation.py` points at it, so a skill author landing in the schema finds the doc.

ros2_ws/src/brain/manipulation/manipulation/auto_stop.py
theo-michel: It was purely "the data said it can never fire", not a correctness issue with the path itself. Before dropping it (a7b56b0b) I replayed the phase-0 hardware capture through the detector: the commanded stream never goes still — ACT's per-chunk replanning puts a jerk spike at every chunk boundary, and the longest still window at eps=0.01 was **0.7 s** — so no `(idle_seconds, eps)` pair both fires at the settled tail and stays quiet mid-run. Command deltas measure the planner's chatter, not the robot's stillness.

Your corroboration argument is right, though — it's the only check in this design that would catch a *confidently wrong* progress head rather than a noisy one. But as built it corroborated nothing on real data: it silently never fired, which is arguably worse than absent because it reads as a safety net. If we get a stillness signal that's actually still (measured joint velocities rather than command deltas, or a dedicated termination head), reintroducing it is the natural move.

Per your last point, the gap is now explicit in the new README's known-limitations section (cf87487a): `stable_min`/`stable_seconds` filter noise, not systematic error; there's no independent corroboration of "done"; and auto-stop should be treated as untested outside supervised eval rollouts.

theo-michel:
Third review pass — disposition of the open findings:

- **Headless skill start on teardown** (greptile P1): fixed in 5f53d4cb. `runRollout` re-checks `destroyed` after `capture.start()` resolves; it re-discards the episode (covering the race with destroy()'s own discard) and never sends the goal.
- **Opaque failures at the robot** (@DavidDobas field report): fixed in 5f53d4cb. Errors now show in a persistent on-page line with the real message (timeouts name the failing service), plus an "auto-continue paused" note. Auto-continue already couldn't slam a restarting server — it only re-arms from a successful save — so retry/backoff was deliberately left out (rationale in-thread).
- **Charts couldn't support the tuning workflow** (@DavidDobas, both charts): fixed in 6eeb477d. Live quality charts now draw the *whole* run (min/max-bucket decimation, spikes survive) on a real time axis with a hover (time, value) readout; the episode-player trace pins progress to an absolute 0–1 scale with labeled gridlines and gets the same hover readout.
- **Auto-stop knobs undocumented** (@DavidDobas): cf87487a adds `manipulation/README.md` — knobs, how they combine, the `engage_below` one-way latch, per-skill tuning workflow, defaults provenance, and known limitations.
- **Why idle-stop was removed** (@DavidDobas): answered in-thread — hardware data showed the command stream is never still (longest window 0.7 s at usable eps), so it silently never fired; the confidently-wrong-progress-head gap it would have covered is now documented explicitly as an open limitation.

Verification: `node --check` clean on all touched JS, `ruff check`/`format` clean, and the new decimation/nearest-point helpers property-tested (30k-sample run → 400 points, single-sample spike preserved, temporal order kept, hover clamps at both edges).

theo-michel:
Field-report fix (4830ffab): promoting several rollout episodes in rapid succession made later ones fail with `copy_episode timed out after 10000ms`. A copy is a real file copy on the robot (h5 + MP4s + profile trace) that the recorder serializes, and the webapp fired every call at click time — so a queued call's 10s timer expired while it waited behind earlier copies. And since rosbridge doesn't cancel a timed-out call, the copy usually still landed, making the error toast a lie and a retry a duplicate.

Copies are now chained client-side (each call's timeout window covers only its own copy) with a 60s window sized for long episodes; a failed copy doesn't block the queued ones, and a genuine timeout now says the copy may still finish so you refresh instead of double-adding.

Left alone deliberately: all recorder services share one mutually-exclusive callback group, so a copy in flight still delays (not fails) a concurrent `new_episode`/`save_episode` by a few seconds. Splitting copies into their own callback group would fix that but requires making `TaskManager`'s in-memory state thread-safe first — worth its own change if it bites in practice.

theo-michel:
Field-report fix #2 (1c29140c): training crashed at collate with `stack expects each tensor to be equal size, but got [30, 10] ... and [30, 14]` after promoting rollout episodes into a training dataset. Root cause: the dataset ended up with mixed `/action` widths — every *current* writer produces 10 columns (6 arm + 2 base + 2 head/termination), so the 14-wide episodes were recorded by a different robot software version than the 10-wide ones, and `copy_episode` merged them without checking.

`copy_episode` now probes the `/action` column count of the incoming episode against one readable destination episode and refuses a mismatch with a message naming both widths — same shape of guard as the existing mixed-frequency refusal. The refusal surfaces in the webapp toast via the existing error path.

Sibling gap, not addressed here: *resuming recording* into an old dataset with a newer recorder can create the same mix without any copy involved — if that bites, the same probe belongs at `new_episode`/resume time.

theo-michel:
Greptile's latest-pass summary finding (no inline thread) fixed in 3e3118e0: `runRollout` read `select.value` twice around the `capture.start()` await, so the episode's policy provenance and the executed `skill_type` could diverge if the skills topic rebuilt the select mid-start (worst case: an empty `skill_type` goal). The picked skill is now read once up front and feeds both.

All inline threads from this review round were already addressed and resolved in earlier passes — no other changes.

PR #498, opened 2026-07-07
Make camera resolution controllable via settings.yaml (INN-508)

theo-michel:
## Summary

Participants asked how to change camera resolution via software ([INN-508](https://linear.app/innate/issue/INN-508)). The main/arm camera drivers were already ROS-param driven, but the knobs weren't surfaced in the centralized `config/settings.yaml`, the WebRTC teleop encoder hardcoded 640x480, and the depth estimator dropped every frame that didn't match its own `image_width`/`image_height` params — so changing resolution silently broke depth.

- **settings.yaml template**: camera stanza now exposes the full resolution chain for the main camera (capture `width`/`height` → `publish_stereo_*` → `publish_left_*`), plus new `arm_camera_driver` (capture size/fps) and `webrtc_streamer` (per-camera stream encode size) stanzas. Fixed the stanza's `fps` to match the shipped default (15.0) so uncommenting unedited stays a no-op.
- **WebRTC streamer**: encode resolution is now a per-camera param (`main_width/main_height`, `arm_width/arm_height`, default 640x480); the node now layers `settings_params()` in the composable launch like the other camera nodes.
- **Stereo depth estimator**: drops the redundant `image_width`/`image_height` params and adopts the incoming image size (SGM always runs at calibration resolution; input size only sets output size/scale), so depth follows the camera automatically instead of needing a second config edit. `depth_scale_` state removed — computed at its single use site.
- **docs/PARAMETERS.md**: table updated with the new stanzas.

## Testing (live robot R7-27)

- `innate build mars_cam` clean; all camera nodes healthy at defaults (main left + arm topics at 640x480, WebRTC encoders 640x480, encode pipeline PLAYING).
- Overrode via settings.yaml: `publish_left_* = 320x240`, arm capture `320x240`, `main_width/height = 320x240` → after `innate restart`: main left topic 320x240, arm sensor negotiated + published 320x240, WebRTC main encoder 320x240, compressed (brain) topics flowing, encode pipeline PLAYING, no errors.
- Restored defaults, re-verified 640x480 on all topics.
- Note: this robot has no stereo calibration (`data/` has no `calibration_config`), so depth is gated off here exactly as before the change; the adoption path runs post-calibration and only affects input validation + output sizing.

### DIFF
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -66,12 +66,22 @@
 #     max_jerk: 150.0               # rad/s^3 trajectory jerk limit (0 disables)
 #     stress_enabled: false         # motor-protection cooldown
 
-# ── Camera (physical main camera; hardware only) ──────────────────────
+# ── Camera: main (stereo; hardware only) ──────────────────────────────
+# Resolution is staged: the sensor captures the full side-by-side stereo frame at
+# width x height (must be a mode the sensor supports), the GPU downscales it to
+# publish_stereo_*, and each eye (half the stereo width) is resized to
+# publish_left_* — the size every ROS topic carries. Downstream consumers (depth,
+# brain, manipulation) follow the published size automatically; the teleop stream
+# size is separate (see webrtc_streamer below).
 # main_camera_driver:
 #   ros__parameters:
-#     publish_left_width: 640       # streamed main-camera image width (px)
-#     publish_left_height: 480      # streamed main-camera image height (px)
-#     fps: 30.0                     # camera frame rate
+#     width: 2560                   # capture width, both eyes combined (px)
+#     height: 720                   # capture height (px)
+#     publish_stereo_width: 1280    # stereo-frame downscale before the left/right split
+#     publish_stereo_height: 480
+#     publish_left_width: 640       # published per-eye image width (px)
+#     publish_left_height: 480      # published per-eye image height (px)
+#     fps: 15.0                     # camera frame rate
 #     jpeg_quality: 80              # JPEG compression quality (1-100)
 #     auto_exposure_mode: 0         # 0=hardware AE, 1=custom PID, 2=manual
 #     exposure: -1                  # manual exposure time (-1 = keep current; 1-10000)
@@ -79,6 +89,23 @@
 #     default_gain: 110             # gain used in auto-exposure mode (0-255)
 #     target_brightness: 128.0      # auto-exposure target brightness (0-255)
 #     ae_kp: 0.8                    # auto-exposure proportional gain
+
+# ── Camera: arm/wrist (hardware only) ─────────────────────────────────
+# arm_camera_driver:
+#   ros__parameters:
+#     width: 640                    # capture width (must be a mode the sensor supports)
+#     height: 480                   # capture height
+#     fps: 30.0                     # camera frame rate
+
+# ── Camera: teleop stream (WebRTC) ────────────────────────────────────
+# The size each camera is encoded at for the app/teleop video stream — incoming
+# frames are resized to it, independent of the published topic size above.
+# webrtc_streamer:
+#   ros__parameters:
+#     main_width: 640               # main-camera stream size (px)
+#     main_height: 480
+#     arm_width: 640                # arm-camera stream size (px)
+#     arm_height: 480
 
 # ── Manipulation (learned-skill execution) ────────────────────────────
 # chunk_size is fixed by the trained model. These are node-level (restart to apply);
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -70,7 +70,9 @@
 # Resolution is staged: the sensor captures the full side-by-side stereo frame at
 # width x height (must be a mode the sensor supports), the GPU downscales it to
 # publish_stereo_*, and each eye (half the stereo width) is resized to
-# publish_left_* — the size every ROS topic carries. Downstream consumers (depth,
+# publish_left_* — the size every ROS topic carries. publish_left_* above half the
+# publish_stereo size only upscales (no extra detail): for a sharper image, raise
+# width x height and publish_stereo_* together. Downstream consumers (depth,
 # brain, manipulation) follow the published size automatically; the teleop stream
 # size is separate (see webrtc_streamer below).
 # main_camera_driver:
--- a/docs/PARAMETERS.md
+++ b/docs/PARAMETERS.md
@@ -47,7 +47,9 @@ To tune something, **uncomment a whole stanza** (the `node:`, `ros__parameters:`
 | `bringup` `battery` | `warning_percentage` / `critical_percentage` | `20` / `10` |
 | `bringup` `safety` (hard `/cmd_vel` clamp) | `max_speed` / `max_angular_speed` | `0.4` / `2.5` |
 | `mars_arm` | `max_jerk` / `stress_enabled` | `150.0` / `false` |
-| `main_camera_driver` (hardware) | `publish_left_width/height`, `fps`, `jpeg_quality`, `auto_exposure_mode`, `exposure`, `gain`, `default_gain`, `target_brightness`, `ae_kp` | see template |
+| `main_camera_driver` (hardware) | `width/height` (capture), `publish_stereo_width/height`, `publish_left_width/height` (published size — depth/brain/manipulation follow it automatically), `fps`, `jpeg_quality`, `auto_exposure_mode`, `exposure`, `gain`, `default_gain`, `target_brightness`, `ae_kp` | see template |
+| `arm_camera_driver` (hardware) | `width/height` (capture), `fps` | `640/480`, `30.0` |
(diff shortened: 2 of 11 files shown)


ros2_ws/src/mars_bot/mars_cam/include/mars_cam/stereo_depth_estimator.hpp
```diff
@@ -177,9 +177,8 @@ class StereoDepthEstimator : public rclcpp::Node {
     int disparity_border_margin_;
 
     // Image dimensions
-    int image_width_, image_height_;  // input (from camera)
-    int calib_width_, calib_height_;  // calibration (processing)
-    double depth_scale_;
+    int image_width_ = 0, image_height_ = 0;  // input/output — adopted from the incoming images
+    int calib_width_, calib_height_;          // calibration (processing)
```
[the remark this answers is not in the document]
theo-michel: Done — `calib_width_`/`calib_height_` now default to 0 in 89346773.

docs/PARAMETERS.md
```diff
@@ -47,7 +47,9 @@ To tune something, **uncomment a whole stanza** (the `node:`, `ros__parameters:`
 | `bringup` `battery` | `warning_percentage` / `critical_percentage` | `20` / `10` |
 | `bringup` `safety` (hard `/cmd_vel` clamp) | `max_speed` / `max_angular_speed` | `0.4` / `2.5` |
 | `mars_arm` | `max_jerk` / `stress_enabled` | `150.0` / `false` |
-| `main_camera_driver` (hardware) | `publish_left_width/height`, `fps`, `jpeg_quality`, `auto_exposure_mode`, `exposure`, `gain`, `default_gain`, `target_brightness`, `ae_kp` | see template |
+| `main_camera_driver` (hardware) | `width/height` (capture), `publish_stereo_width/height`, `publish_left_width/height` (published size — depth/brain/manipulation follow it automatically), `fps`, `jpeg_quality`, `auto_exposure_mode`, `exposure`, `gain`, `default_gain`, `target_brightness`, `ae_kp` | see template |
```
vignesh-anand: Nit — inconsistent default-column convention within this PR. This `main_camera_driver` row's default column says "see template," while the two new rows just below (`arm_camera_driver`, `webrtc_streamer`) give explicit defaults (`640/480`, `30.0`; `640/480 each`). Same table, same PR, two conventions — and this row is now a dense one-liner. Either give this row explicit defaults too, or keep all three as "see template."

theo-michel: Done — all three camera rows now say "see template" (the main row is too key-heavy for explicit defaults), and trimmed the main row's parenthetical. Pushed in 98543efb.

config/settings.yaml.template
```diff
@@ -66,12 +66,22 @@
...
+# brain, manipulation) follow the published size automatically; the teleop stream
+# size is separate (see webrtc_streamer below).
 # main_camera_driver:
 #   ros__parameters:
-#     publish_left_width: 640       # streamed main-camera image width (px)
-#     publish_left_height: 480      # streamed main-camera image height (px)
-#     fps: 30.0                     # camera frame rate
+#     width: 2560                   # capture width, both eyes combined (px)
+#     height: 720                   # capture height (px)
+#     publish_stereo_width: 1280    # stereo-frame downscale before the left/right split
```
vignesh-anand, line 82: Not introduced by this PR (these params/behavior pre-date it — this PR just documents them), but flagging while it's fresh since this stanza is where a user meets them:

**1. Three resolution knobs is one too many, and the extra one is a trap.** The per-eye source is `publish_stereo_width/2 × publish_stereo_height` (main_camera_driver.cpp:125), so `publish_left_*` set *above* half-stereo is pure upscaling — no added detail, just a blurrier bigger image. A user chasing "higher res" (literally the ticket) will bump `publish_left_width` and get nothing. `publish_left` only does something useful when it's *smaller* than the natural crop (bandwidth) or a different aspect — a power-user escape hatch promoted to a first-class setting. Two knobs (sensor mode + published per-eye size, intermediate *derived*) would cover ~95% of cases without the footgun.

**2. Capture-high-then-hardware-shrink (2560×720 → 1280×480) is probably fine but worth a sanity check.** There are legit reasons (full FOV vs a cropped low-res sensor mode, cleaner supersampled downscale, VIC shrink is ~free), so I don't think it's wrong — but capturing high does cost real USB bandwidth + `nvv4l2decoder` work every frame, and it's not obvious from the config that raising resolution is cheap at capture (hardware) yet proportionally expensive for every downstream *compressed-topic* decode (those are CPU `cv::imdecode`, not NVDEC). If the high capture is deliberately for FOV, a one-liner saying so would kill the "why aren't we just capturing at 1280×480" question for the next reader.

Neither blocks this PR — it's a documentation PR and both are pre-existing. Raising as a possible follow-up: collapse to 2 knobs + note the capture-res rationale. @karmanyaahm curious for your take since you know this driver's history.

ros2_ws/src/mars_bot/mars_cam/mars_cam/webrtc_streamer.cpp
```diff
@@ -184,15 +184,17 @@ void WebRTCStreamer::configure_cameras() {
         cam->live_topic = this->declare_parameter<std::string>("live_" + name + "_camera_topic", def_live);
         cam->replay_topic = this->declare_parameter<std::string>("replay_" + name + "_camera_topic", def_replay);
         cam->fps = static_cast<int>(this->declare_parameter<int>(name + "_fps", def_fps));
+        cam->width = static_cast<int>(this->declare_parameter<int>(name + "_width", 640));
+        cam->height = static_cast<int>(this->declare_parameter<int>(name + "_height", 480));
```
vignesh-anand, line 188: Not for this PR — just flagging for later. Some of these resolution params could benefit from some kind of validation. Right now nothing catches a bad value: e.g. `main_width: 0` in settings.yaml sails through to the appsrc caps and fails the pipeline build with a cryptic gst error instead of a clear message, and there's no guard against the coupling traps we discussed (per-eye/stream size set above what the upstream stage actually provides — silent upscaling).

Not sure what the best approach is — options I can think of are ROS `ParameterDescriptor` ranges (native but unused anywhere in this repo, and can't express cross-param rules like `publish_left ≤ publish_stereo/2`), or extending the existing launch-time `config_loader` validator (already does the int-vs-double type check, so it'd be consistent and could handle cross-param rules too). No strong opinion — just something worth thinking about in a follow-up.

ros2_ws/src/mars_bot/mars_cam/mars_cam/depth_estimator/publishing.cpp
```diff
@@ -113,7 +113,7 @@ void StereoDepthEstimator::publishDisparityMsg(const cv::Mat& disparity_float, c
     msg->image.data.resize(msg->image.height * msg->image.step);
     memcpy(msg->image.data.data(), disp_full.data, msg->image.data.size());
 
-    const float scale_up = 1.0f / static_cast<float>(depth_scale_);
+    const float scale_up = static_cast<float>(image_width_) / static_cast<float>(calib_width_);
```
vignesh-anand, line 116: Observation — pre-existing, dormant, and **not introduced by this PR** (just noting it while tracing the scale math here, low confidence).

This scales `f` by `s = image_width_/calib_width_`, but the disparity *values* stay at calib scale — the `cv::resize` above (line 102) is spatial-only (INTER_NEAREST, no value multiply). So a consumer computing metric depth as `Z = f·T/d` from the DisparityImage topic would see `f` and `d` inconsistent by factor `s` whenever output ≠ calib resolution.

Why this is a heads-up, not a finding:
- **Not a regression** — the old code applied the same `scale_up` to `f` and never scaled the disparity either. This PR's math is algebraically identical (`1/depth_scale_` == `image_width_/calib_width_`); it just moved the computation to the use site. Out of scope.
- **Dormant in-repo** — grepped the whole tree: nothing subscribes to the disparity topic, only mars_cam publishes it. No in-repo consumer is affected.
- **The real depth path is correct** — the dedicated depth topic (16SC1 mm) carries physical distances, resolution-independent, so metric-depth consumers are unaffected regardless of `s`.
- I may also be missing a normalization — `delta_d = 1/32` (VPI Q10.5 fixed-point) suggests a disparity scale convention I haven't fully traced.

So: worth an eye only if an *external* tool is ever expected to derive metric depth from the disparity topic at a non-default resolution. Genuinely irrelevant to merging this PR.

config/settings.yaml.template
theo-michel: Added the upscaling warning to the stanza in 54595fbd ("publish_left_* above half the publish_stereo size only upscales — raise width x height and publish_stereo_* together"). Leaving the knob-collapse + capture-res rationale to the follow-up you proposed (and karmanyaahm's take), so not resolving this thread.

karmanyaahm: Yeah from what I remember it was something to do with weird FOV behavior - though for the camera driver I mostly followed @vignesh-anand's original implementation, kept the logic the same and only cleaned up some of the high level code.

Would be useful to re-test

PR #503, opened 2026-07-07
fix(scripts): show webapp URL as mDNS hostname instead of Wi-Fi IP

theo-michel:
## Summary
- The `Web app:` link shown in the tmux launch banner, SSH welcome, networking restart script, and the `innate` CLI printed `https://<wifi-ip>` (e.g. `https://192.168.0.71`).
- All of these go through the shared helper `scripts/webapp_uri.zsh`, so this changes just that helper to print `https://<hostname>.local` instead (e.g. `https://mars-the-27th.local`).
- The Wi-Fi connectivity check is unchanged, so the URL still only appears when the robot is online; if `hostname` fails we fall back to the IP URL.

## Testing
- `zsh -n` syntax check passes.
- Sourced the helper with a stubbed Wi-Fi IP: prints `https://<host>.local`; with `hostname` stubbed to fail, falls back to `https://192.168.0.71`.

### DIFF
--- a/scripts/webapp_uri.zsh
+++ b/scripts/webapp_uri.zsh
@@ -24,6 +24,27 @@ innate_current_wifi_ip() {
     done
 }
 
+# Resolve the mDNS name actually advertised for the given IP (avahi can
+# publish e.g. mars-the-27th-2.local on a name conflict, diverging from the
+# system hostname). Mirrors mars_control's get_hostname() in app.cpp.
+innate_mdns_name() {
+    local ip="$1" name
+
+    name="$(avahi-resolve -a "$ip" 2>/dev/null | awk '{ print $2; exit }')"
+    if [[ -n "$name" ]]; then
+        print -r -- "$name"
+        return 0
+    fi
+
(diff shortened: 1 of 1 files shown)


theo-michel:
Addressed in e3fd236: the helper now reverse-resolves the *actually advertised* mDNS name from the Wi-Fi IP via `avahi-resolve -a` (same approach as `mars_control`'s `get_hostname()` in app.cpp), so a divergence between the system hostname and the advertised name (e.g. `-2` conflict suffix) yields the correct, resolvable link. `hostname -s` is now only a fallback when avahi-resolve is unavailable, and the raw IP remains the final fallback.

theo-michel:
Fixed in $(git rev-parse --short HEAD): dropped the `hostname -s` fallback entirely. `innate_mdns_name` now returns failure unless `avahi-resolve -a` reverse-resolves the advertised name from the Wi-Fi IP, so `innate_webapp_uri` falls back to the known-reachable `https://<ip>` in every other case.

PR #508, opened 2026-07-08
fix(nav): lethal-start fixes (re-landed) + controller retune, cancel race, switch serialization

theo-michel:
## Summary

Two commits of nav-stack fixes, all live-verified on mars-the-44th.

**⚠️ Commit 1 is a re-land.** #504 was merged before its second commit was picked up, so the nav fixes described in #504's PR body never reached main — only the mars_cam changes did. Commit `0baa0b32` here is that nav commit, cherry-picked unchanged. Commit 2 (`3b2d18d5`) is the new tier.

## Commit 1 (re-land) — lethal-start chain & dead config

Map-frame navigation failed deterministically with `GridBased: Starting point in lethal space!` — 0/2 before, 2/2 after on the same speckled map:

- **`amcl.yaml` was never applied** (file keyed `amcl:`, node renamed `navigation_amcl`) — re-keyed with `/**:` and retuned the values that would be harmful once actually applied.
- **SmacPlanner2D → NavFn + `allow_unknown`** (NavFn clears the robot's start cell; Smac2D has no start tolerance).
- **Localization handoff fixed**: grid_localizer seeds AMCL via `/set_initial_pose` with retry (the latched `/initialpose` never reached AMCL's VOLATILE subscription); mode_manager triggers `/localize` after map switches.
- **Spin recovery** in both BTs; **real `nav_through_poses.xml`** (was `<AlwaysSuccess/>` — NavigateThroughPoses goals fake-succeeded instantly).
- Dead/ignored config purged (unplugged voxel block, Jazzy-era params); `simulate_ahead_time` 0→1.5 (BackUps no longer reverse blind); boot warning when the camera obstacle layer is starved (robot silently lidar-only).

## Commit 2 — controller realism, cancel race, switch serialization

- **MPPI 20→10 Hz** (same 2.5 s horizon via `time_steps` 25 × `model_dt` 0.1, batch 1500), `failure_tolerance: 1.0`, smoother `velocity_timeout` 0.2→0.5. At 20 Hz the loop missed its rate constantly and each overrun past 0.2 s zeroed cmd_vel → stutter. After: missed-rate warnings down from continuous to ~14 across two full runs, zero aborts.
- **Velocity caps 0.45→0.4** to match bringup's hardware clamp (`safety.max_speed`), so MPPI's model stops assuming speed the motors never deliver.
- **Router cancel race**: a cancel arriving between server-accept and internal-accept was accepted then silently dropped — the robot kept driving to a "canceled" goal (matches the long-standing `navigate_to_position.py` comment about cancelTask not producing CANCELED). The execute callback now re-checks `is_cancel_requested` after the handoff; cancel forwarding logs rejections. Verified: mid-drive cancel → CANCELED, 5 cm coast.
- **Map/mode switch serialization**: `change_map_callback` now shares the mode service's reentrant group and `_mode_change_lock` (previously the two could interleave lifecycle transitions on the same nodes); `current_map` is restored and not persisted when a switch fails. Verified: two concurrent map changes serialize, both succeed, relocalization fires for each.
- **Scan throttle 6→4 Hz** honesty fix: the lidar delivers ~8 Hz and topic_tools' 6.0 beat down to every-other-scan = 4 Hz anyway; same effective rate, intentional now.
- **Global costmaps 5→2 Hz** update (BT replans at 1 Hz, scans at 4 Hz); the controller's 5 Hz local costmap untouched. Deliberately NOT deactivating the mapfree stack in navigation mode — `local_frame` goals route through it (verified live).
- **`mobility.rotate()`** busy-loop (`while not isTaskComplete(): pass`) pinned a core for entire rotations; now polls at 10 Hz.

## Live verification (post-restart, params queried from running nodes)

controller 10 Hz / tolerance 1.0 / vx_max 0.4 / steps 25 / dt 0.1 / batch 1500; smoother 0.5 s / 20 Hz / [0.4, 0, 0.6]; costmaps 2/2/5 Hz — all confirmed. Local- and map-frame `navigate_to_position` both succeed; cancel and concurrency tests above.

## Known remaining (not in this PR)

Re-record the map (still speckle); restore stereo calibration (camera obstacle layer inert, warned at boot); router awaits internal results without timeout; grid_localizer blocks its executor during GPU search; mode_manager executor idles ~16% CPU.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
@@ -1,7 +1,9 @@
 # NOTE: the AMCL node is renamed to `navigation_amcl` in navigation.launch.py.
 # Param files match by NODE NAME, so this file MUST be keyed `navigation_amcl`
-# (a previous top-level `amcl:` key matched nothing and every value below was
-# silently ignored — AMCL ran on Nav2 defaults).
+# (commit cfdf21d4 renamed the node without re-keying this file, which
+# silently un-applied every value below for months — AMCL ran on Nav2
+# defaults). The values themselves are the long-serving hand-tuned set and
+# are intentionally unchanged by the re-key.
 /**:
   navigation_amcl:
     ros__parameters:
@@ -19,30 +21,25 @@
       global_frame_id: "map"
       lambda_short: 0.1
       laser_likelihood_max_dist: 2.0
-      laser_max_range: 12.0   # RPLidar max usable range
+      laser_max_range: 10.0
       laser_min_range: 0.2
       laser_model_type: "likelihood_field"
       max_beams: 60
-      # 3000 particles is plenty for rooms this size and roughly halves the
-      # filter CPU vs the old 6000 on an already saturated box.
-      max_particles: 3000
-      min_particles: 500
+      max_particles: 6000
+      min_particles: 1000
       odom_frame_id: "odom"
       pf_err: 0.05
       pf_z: 0.99
       recovery_alpha_fast: 0.2
       recovery_alpha_slow: 0.005
-      # Only update/resample after real motion. 0.0/0.0 with resample_interval 1
-      # would update AND resample on every scan -> particle depletion +
-      # overconvergence (and needless CPU) while standing still.
-      resample_interval: 2
-      update_min_a: 0.2
-      update_min_d: 0.15
+      resample_interval: 1
       robot_model_type: "nav2_amcl::DifferentialMotionModel"
       save_pose_rate: 0.5
       sigma_hit: 0.2
       tf_broadcast: true
       transform_tolerance: 1.0
+      update_min_a: 0.0
+      update_min_d: 0.0
       z_hit: 0.5
       z_max: 0.05
       z_rand: 0.5
--- a/ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
@@ -1,44 +1,54 @@
-amcl:
-  ros__parameters:
-    use_sim_time: false
-    alpha1: 0.4
-    alpha2: 0.3
-    alpha3: 0.4
-    alpha4: 0.3
-    alpha5: 0.3
-    base_frame_id: "base_footprint"
-    beam_skip_distance: 0.5
-    beam_skip_error_threshold: 0.9
-    beam_skip_threshold: 0.3
-    do_beamskip: false
-    global_frame_id: "map"
-    lambda_short: 0.1
-    laser_likelihood_max_dist: 2.0
-    laser_max_range: 10.0     
-    laser_min_range: 0.2     
-    laser_model_type: "likelihood_field"
-    max_beams: 60
-    max_particles: 6000
-    min_particles: 1000
-    odom_frame_id: "odom"
-    pf_err: 0.05
-    pf_z: 0.99
-    recovery_alpha_fast: 0.2
-    recovery_alpha_slow: 0.005
-    resample_interval: 1
-    robot_model_type: "nav2_amcl::DifferentialMotionModel"
-    save_pose_rate: 0.5
-    sigma_hit: 0.2
-    tf_broadcast: true
-    transform_tolerance: 1.0
-    update_min_a: 0.0
-    update_min_d: 0.0
-    z_hit: 0.5
-    z_max: 0.05
-    z_rand: 0.5
-    z_short: 0.05
-    scan_topic: /scan
-    map_topic: /map
-    set_initial_pose: false
-    always_reset_initial_pose: true # if lifecycle resets, assume any previous info is invalid
-    first_map_only: false
+# NOTE: the AMCL node is renamed to `navigation_amcl` in navigation.launch.py.
+# Param files match by NODE NAME, so this file MUST be keyed `navigation_amcl`
+# (a previous top-level `amcl:` key matched nothing and every value below was
+# silently ignored — AMCL ran on Nav2 defaults).
+/**:
+  navigation_amcl:
+    ros__parameters:
+      use_sim_time: false
+      alpha1: 0.4
+      alpha2: 0.3
+      alpha3: 0.4
+      alpha4: 0.3
+      alpha5: 0.3
+      base_frame_id: "base_footprint"
+      beam_skip_distance: 0.5
+      beam_skip_error_threshold: 0.9
+      beam_skip_threshold: 0.3
+      do_beamskip: false
+      global_frame_id: "map"
+      lambda_short: 0.1
+      laser_likelihood_max_dist: 2.0
+      laser_max_range: 12.0   # RPLidar max usable range
+      laser_min_range: 0.2
+      laser_model_type: "likelihood_field"
+      max_beams: 60
+      # 3000 particles is plenty for rooms this size and roughly halves the
+      # filter CPU vs the old 6000 on an already saturated box.
+      max_particles: 3000
+      min_particles: 500
+      odom_frame_id: "odom"
+      pf_err: 0.05
+      pf_z: 0.99
+      recovery_alpha_fast: 0.2
+      recovery_alpha_slow: 0.005
+      # Only update/resample after real motion. 0.0/0.0 with resample_interval 1
+      # would update AND resample on every scan -> particle depletion +
+      # overconvergence (and needless CPU) while standing still.
+      resample_interval: 2
+      update_min_a: 0.2
+      update_min_d: 0.15
+      robot_model_type: "nav2_amcl::DifferentialMotionModel"
+      save_pose_rate: 0.5
+      sigma_hit: 0.2
+      tf_broadcast: true
+      transform_tolerance: 1.0
+      z_hit: 0.5
+      z_max: 0.05
+      z_rand: 0.5
+      z_short: 0.05
+      scan_topic: /scan
+      map_topic: /map
+      set_initial_pose: false
+      always_reset_initial_pose: true # if lifecycle resets, assume any previous info is invalid
+      first_map_only: false
--- a/ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
@@ -1,7 +1,13 @@
 /**:
   ros__parameters:
     use_sim_time: false
-    controller_frequency: 20.0
+    # 10 Hz: 20 Hz MPPI was not achievable on this CPU (constant "Control loop
+    # missed its desired rate" spam), and every overrun past the velocity
+    # smoother's timeout zeroed cmd_vel -> stop-start stutter.
+    controller_frequency: 10.0
+    # Tolerate brief controller hiccups instead of aborting the whole goal on
+    # the first missed cycle (seconds; 0.0 = abort immediately).
+    failure_tolerance: 1.0
     min_x_velocity_threshold: 0.001
     min_y_velocity_threshold: 0.5         # Differential drive: no lateral motion
     min_theta_velocity_threshold: 0.001
@@ -32,14 +38,15 @@
       
       # Motion model and basic parameters
       motion_model: "DiffDrive"
-      time_steps: 50                    # 50 steps * 0.05s = 2.5s prediction horizon (matches old sim_time)
-      model_dt: 0.05                    # 1/controller_frequency = 1/20 = 0.05s
-      batch_size: 2000                  # Number of trajectory samples
+      time_steps: 25                    # 25 steps * 0.1s = same 2.5s prediction horizon at half the compute
+      model_dt: 0.1                     # 1/controller_frequency = 1/10 = 0.1s
+      batch_size: 1500                  # Number of trajectory samples
       iteration_count: 1
-      
+
       # Velocity limits (from existing DWB config)
       vx_min: -0.2                      # matches min_vel_x
-      vx_max: 0.45                       # matches max_vel_x
+      vx_max: 0.4                       # bringup hard-clamps at safety.max_speed 0.4 — commanding above it
+                                        # makes MPPI's model assume speed the motors never deliver
       vy_max: 0.0                       # differential drive - no lateral motion
       wz_max: 0.6                       # matches velocity_smoother max
       
--- a/ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
@@ -1,13 +1,7 @@
 /**:
   ros__parameters:
     use_sim_time: false
-    # 10 Hz: 20 Hz MPPI was not achievable on this CPU (constant "Control loop
-    # missed its desired rate" spam), and every overrun past the velocity
-    # smoother's timeout zeroed cmd_vel -> stop-start stutter.
-    controller_frequency: 10.0
-    # Tolerate brief controller hiccups instead of aborting the whole goal on
-    # the first missed cycle (seconds; 0.0 = abort immediately).
-    failure_tolerance: 1.0
+    controller_frequency: 20.0
     min_x_velocity_threshold: 0.001
     min_y_velocity_threshold: 0.5         # Differential drive: no lateral motion
     min_theta_velocity_threshold: 0.001
@@ -38,15 +32,14 @@
       
       # Motion model and basic parameters
       motion_model: "DiffDrive"
-      time_steps: 25                    # 25 steps * 0.1s = same 2.5s prediction horizon at half the compute
-      model_dt: 0.1                     # 1/controller_frequency = 1/10 = 0.1s
-      batch_size: 1500                  # Number of trajectory samples
+      time_steps: 50                    # 50 steps * 0.05s = 2.5s prediction horizon (matches old sim_time)
+      model_dt: 0.05                    # 1/controller_frequency = 1/20 = 0.05s
+      batch_size: 2000                  # Number of trajectory samples
       iteration_count: 1
-
+      
       # Velocity limits (from existing DWB config)
       vx_min: -0.2                      # matches min_vel_x
-      vx_max: 0.4                       # bringup hard-clamps at safety.max_speed 0.4 — commanding above it
-                                        # makes MPPI's model assume speed the motors never deliver
+      vx_max: 0.45                       # matches max_vel_x
       vy_max: 0.0                       # differential drive - no lateral motion
       wz_max: 0.6                       # matches velocity_smoother max
       
--- a/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
@@ -2,10 +2,8 @@ navigation:
   global_costmap:
     global_costmap:
       ros__parameters:
-        # 2 Hz is plenty for a static-map costmap: the BT replans at 1 Hz and
-        # the only dynamic input (/scan) arrives at 4 Hz. Was 5 Hz of pure CPU.
-        update_frequency: 2.0
-        publish_frequency: 2.0
+        update_frequency: 5.0
+        publish_frequency: 5.0
         global_frame: map
         robot_base_frame: base_link
         use_sim_time: false
@@ -137,12 +135,8 @@ mapfree:
   global_costmap:
     global_costmap:
       ros__parameters:
-        # 2 Hz: the planner replans at 1 Hz and /scan arrives at 4 Hz; this
-        # rolling costmap only feeds path planning (the controller avoids
-        # obstacles via the 5 Hz local costmap). This stack runs even while in
-        # navigation mode (local_frame goals use it), so its idle cost matters.
-        update_frequency: 2.0
-        publish_frequency: 2.0
+        update_frequency: 5.0
+        publish_frequency: 5.0
         global_frame: odom
         robot_base_frame: base_link
         use_sim_time: false
--- a/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
@@ -2,8 +2,10 @@ navigation:
   global_costmap:
     global_costmap:
       ros__parameters:
-        update_frequency: 5.0
-        publish_frequency: 5.0
+        # 2 Hz is plenty for a static-map costmap: the BT replans at 1 Hz and
+        # the only dynamic input (/scan) arrives at 4 Hz. Was 5 Hz of pure CPU.
+        update_frequency: 2.0
+        publish_frequency: 2.0
         global_frame: map
         robot_base_frame: base_link
         use_sim_time: false
@@ -138,8 +140,12 @@ mapfree:
   global_costmap:
     global_costmap:
       ros__parameters:
-        update_frequency: 5.0
-        publish_frequency: 5.0
+        # 2 Hz: the planner replans at 1 Hz and /scan arrives at 4 Hz; this
+        # rolling costmap only feeds path planning (the controller avoids
+        # obstacles via the 5 Hz local costmap). This stack runs even while in
+        # navigation mode (local_frame goals use it), so its idle cost matters.
+        update_frequency: 2.0
+        publish_frequency: 2.0
         global_frame: odom
         robot_base_frame: base_link
         use_sim_time: false
--- a/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
@@ -9,8 +9,8 @@ navigation:
         use_sim_time: false
         footprint: "[[0.25, 0.165], [0.25, -0.165], [-0.20, -0.165], [-0.20, 0.165]]"
         resolution: 0.05
-        track_unknown_space: true        # Now unknown areas are treated as obstacles
-        unknown_cost_value: 255          # Unknown cells set to a lethal cost
+        track_unknown_space: true        # keep unknown cells distinct (NO_INFORMATION)
+        unknown_cost_value: 255          # 255 = NO_INFORMATION (not lethal); planner treats it per allow_unknown
         rolling_window: false
         plugins: ["static_layer", "obstacle_layer", "inflation_layer"]
         static_layer:
@@ -36,49 +36,9 @@ navigation:
           cost_scaling_factor: 10.0
           inflation_radius: 0.3
         always_send_full_costmap: False
-        voxel_layer:
-          plugin: "spatio_temporal_voxel_layer/SpatioTemporalVoxelLayer"
-          enabled:               true
-          voxel_decay:           2.0     #seconds if linear, e^n if exponential
-          decay_model:           0      #0=linear, 1=exponential, -1=persistent
-          voxel_size:            0.05   #meters
-          track_unknown_space:   true   #default space is unknown
-          observation_persistence: 0.0  #seconds
-          max_obstacle_height:   0.36    #meters
-          mark_threshold:        0      #voxel height
-          update_footprint_enabled: true
-          combination_method:    1      #1=max, 0=override
-          obstacle_range:        3.0    #meters
-          origin_z:              0.0    #meters
-          publish_voxel_map:     false   # default off
-          transform_tolerance:   0.2    # seconds
-          observation_sources:   rgbd1_clear rgbd1_mark
-          rgbd1_mark:
-            data_type: PointCloud2
-            topic: /mars/main_camera/points
-            marking: true
-            clearing: false
-            obstacle_range: 2.5
-            min_obstacle_height: 0.05     #default 0, meters
-            max_obstacle_height: 0.36     #defaule 3, meters
-            expected_update_rate: 0.0    #default 0, if not updating at this rate at least, remove from buffer
-            observation_persistence: 0.0 #default 0, use all measurements taken during now-value, 0=latest 
-            inf_is_valid: false          #default false, for laser scans
-            clear_after_reading: false    #default false, clear the buffer after the layer gets readings from it
-            filter: "voxel"              #default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommended to have at one filter on
-            voxel_min_points: 5          #default 0, minimum points per voxel for voxel filter
-          rgbd1_clear:
-            enabled: false                #default true, can be toggled on/off with associated service call
-            data_type: PointCloud2
-            topic: /mars/main_camera/points
-            marking: false
-            clearing: true
-            min_z: 0.0                   #default 0, meters
-            max_z: 7.0                   #default 10, meters
-            vertical_fov_angle: 0.7      #default 0.7, radians
-            horizontal_fov_angle: 1.04   #default 1.04, radians
-            decay_acceleration: 1.       #default 0, 1/s^2. If laser scanner MUST be 0
-            model_type: 0                #default 0 (depth camera). Use 1 for 3D Lidar
+        # (A full voxel_layer block used to sit here unused — it was never in
+        # the plugins list above. The camera voxel layer lives in the local
+        # and mapfree costmaps.)
 
 
 local_costmap:
@@ -94,8 +54,8 @@ local_costmap:
       height: 6
       resolution: 0.05
       footprint: "[[0.25, 0.165], [0.25, -0.165], [-0.20, -0.165], [-0.20, 0.165]]"
-      track_unknown_space: false        # Ensure unknown space is considered as obstacles
-      unknown_cost_value: 255          # Unknown cells get a lethal cost value
+      track_unknown_space: false        # unknown space is treated as FREE in the rolling local costmap
+      unknown_cost_value: 255          # unused while track_unknown_space is false
       plugins: [
         "voxel_layer",
         "obstacle_layer",
@@ -120,6 +80,12 @@ local_costmap:
         cost_scaling_factor: 10.0
         inflation_radius: 0.25
       voxel_layer:
+        # NOTE: this layer is silently inert when the robot has no stereo
+        # calibration (/mars/main_camera/points never publishes) — nav is then
+        # lidar-only. mode_manager logs a warning when that is the case.
+        # expected_update_rate is left at 0 on purpose: a nonzero value marks
+        # the whole costmap not-current when the camera is down, which blocks
+        # planning entirely.
         plugin: "spatio_temporal_voxel_layer/SpatioTemporalVoxelLayer"
         enabled:               true
         voxel_decay:           2.0     #seconds if linear, e^n if exponential
@@ -209,6 +175,8 @@ mapfree:
           inflation_radius: 0.35
         always_send_full_costmap: False
         voxel_layer:
+          # NOTE: silently inert without stereo calibration — see the comment
+          # on the local costmap's voxel_layer.
           plugin: "spatio_temporal_voxel_layer/SpatioTemporalVoxelLayer"
           enabled:               true
           voxel_decay:           2.0     #seconds if linear, e^n if exponential
--- a/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
+++ b/ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
@@ -38,9 +38,6 @@ navigation:
           cost_scaling_factor: 10.0
           inflation_radius: 0.3
         always_send_full_costmap: False
-        # (A full voxel_layer block used to sit here unused — it was never in
-        # the plugins list above. The camera voxel layer lives in the local
-        # and mapfree costmaps.)
 
 
 local_costmap:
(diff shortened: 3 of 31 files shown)


ros2_ws/src/mars_bot/mars_nav/mars_nav/grid_localizer.py
```diff
@@ -501,6 +513,48 @@ def _publish_pose(self, x: float, y: float, theta: float):
...
+        pose_msg, retries_left = self._pending_seed
+        self._seed_amcl(pose_msg, retries_left - 1)
+
+    def _seed_amcl_done(self, future):
+        self._pending_seed = None
+        try:
+            future.result()
+            self.get_logger().info("Seeded AMCL via /set_initial_pose")
+        except Exception as e:
+            self.get_logger().warning(f"AMCL /set_initial_pose call failed: {e}")
```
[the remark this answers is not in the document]
theo-michel: Fixed in 307841fb: a failed call now schedules a retry, and the done callback closes over its own pose and verifies it is still the pending seed before clearing or rescheduling — a stale in-flight call can no longer clobber a newer pose's retry state. Verified live (seed delivered after map switch).

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -653,31 +714,39 @@ def change_map_callback(self, request, response):
...
+                if self.current_mode == "navigation":
+                    self.get_logger().info(f"Efficiently switching to new map: {requested_map}")
 
-                success, message = self._efficient_map_switch()
+                    success, message = self._efficient_map_switch()
 
-                if success:
+                    if success:
+                        self.save_last_map(requested_map)
+                        self._trigger_relocalization()
```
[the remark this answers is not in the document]
theo-michel: Fixed in 307841fb: relocalization now runs after the lock is released (the lifecycle switch is complete at that point), so concurrent mode/map service calls are no longer blocked by the up-to-9s localize wait. Verified live: switch + 'Re-localized on new map' both succeed.

ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
```diff
@@ -1,44 +1,54 @@
...
+      global_frame_id: "map"
+      lambda_short: 0.1
+      laser_likelihood_max_dist: 2.0
+      laser_max_range: 12.0   # RPLidar max usable range
+      laser_min_range: 0.2
+      laser_model_type: "likelihood_field"
+      max_beams: 60
+      # 3000 particles is plenty for rooms this size and roughly halves the
+      # filter CPU vs the old 6000 on an already saturated box.
+      max_particles: 3000
```
vignesh-anand: Finding is correct, the file wasn't being applied. But it's a regression, not dead config: the node was `name='amcl'` and matched the `amcl:` key until `cfdf21d4` (2026-01-03) renamed it to `navigation_amcl` without re-keying. It worked for ~10 months before that.

The values (max_particles, resample_interval, update_min_d/a, laser_max_range) come from a lot of hand-tuning across different environments. I wouldn't change them alongside the re-key without aggressive testing first.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
```diff
@@ -2,28 +2,18 @@
...
-        max_iterations: 1000000
-        max_on_approach_iterations: 1000
-        terminal_checking_interval: 5000
-        max_planning_time: 2.0
-        cost_travel_multiplier: 2.0
+        # NavFn instead of SmacPlanner2D: NavFn clears the robot's start cell
+        # before planning, so an AMCL pose that drifts into an inflated
+        # obstacle degrades gracefully instead of aborting instantly with
+        # "Starting point in lethal space" (Smac2D has no start tolerance).
+        plugin: "nav2_navfn_planner/NavfnPlanner"
```
vignesh-anand: Smac2D here wasn't a default, it was a deliberate choice for performance after a lot of testing. NavFn was on the table and we didn't go with it — I don't remember the specific reason anymore, but there was one.

The lethal-start abort is a real problem, but swapping back to NavFn undoes that tested decision. I'd rather fix the start tolerance within Smac2D than reverse the planner choice. I wouldn't make this swap without re-running the testing that led us to Smac2D in the first place.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
```diff
@@ -2,28 +2,18 @@
...
-        tolerance: 0.1
-        downsample_costmap: false
-        downsampling_factor: 1
-        allow_unknown: false
-        max_iterations: 1000000
-        max_on_approach_iterations: 1000
-        terminal_checking_interval: 5000
-        max_planning_time: 2.0
-        cost_travel_multiplier: 2.0
+        # NavFn instead of SmacPlanner2D: NavFn clears the robot's start cell
```
vignesh-anand: nit: "NavFn clears the robot's start cell" isn't quite the mechanism. NavFn doesn't clear anything — it just has no start-in-lethal rejection guard, whereas Smac2D explicitly checks the start cell and aborts with "Starting point in lethal space!". So NavFn tolerates a lethal start by omission, not by an active clearing step. Same outcome, but the wording would send someone looking for clearing logic that doesn't exist. Something like "NavFn has no start-in-lethal rejection, so it degrades gracefully where Smac2D hard-aborts" is more accurate.

ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
karmanyaahm, line 1: # AMCL config vs Nav2 Humble defaults

Compared against the defaults declared in `nav2_amcl/src/amcl_node.cpp` (Nav2 `humble` branch).

## Values that differ from the Nav2 Humble defaults

| Parameter | Ours | Default | Effect of our value |
|---|---|---|---|
| `alpha1` | 0.4 | 0.2 | More expected rotation noise from rotation |
| `alpha2` | 0.3 | 0.2 | More rotation noise from translation |
| `alpha3` | 0.4 | 0.2 | More translation noise from translation |
| `alpha4` | 0.3 | 0.2 | More translation noise from rotation |
| `alpha5` | 0.3 | 0.2 | Only used by the omni model — ignored with `DifferentialMotionModel` |
| `laser_max_range` | 12.0 | 100.0 | Caps beams at the RPLidar's real usable range |
| `laser_min_range` | 0.2 | -1.0 | Default -1.0 means "use the laser's reported min"; we override to 0.2 m |
| `max_particles` | 3000 | 2000 | *More* than default — the comment says it halves CPU vs an old 6000, but stock Nav2 is 2000 |
| `recovery_alpha_fast` | 0.2 | 0.0 | Default 0.0 **disables** recovery; the source's suggested "good value" is 0.1, so 0.2 is aggressive |
| `recovery_alpha_slow` | 0.005 | 0.0 | Default disables recovery; suggested good value is 0.001, ours is 5× that |
| `resample_interval` | 2 | 1 | Resample every 2nd filter update instead of every update |
| `update_min_d` | 0.15 | 0.25 | Filter updates after less translation (more frequent updates, slightly more CPU) |
| `always_reset_initial_pose` | true | false | On lifecycle reset, discard the last known pose instead of reusing it |
| `scan_topic` | `/scan` | `scan` | Absolute vs relative — identical unless the node is namespaced; ours pins it to the root |
| `map_topic` | `/map` | `map` | Same story |

## Values identical to the defaults (redundant lines)

Everything else in the file restates the stock default: `use_sim_time` (false), `base_frame_id`
(`base_footprint`), all four `beam_skip_*` / `do_beamskip`, `global_frame_id`, `lambda_short`,
`laser_likelihood_max_dist`, `laser_model_type`, `max_beams` (60), `min_particles` (500),
`odom_frame_id`, `pf_err`, `pf_z`, `update_min_a` (0.2), `robot_model_type`, `save_pose_rate`,
`sigma_hit`, `tf_broadcast`, `transform_tolerance` (1.0), `z_hit`/`z_max`/`z_rand`/`z_short`,
`set_initial_pose` (false), and `first_map_only` (false).

## Notes

- **The recovery alphas are the most consequential change.** Stock Nav2 ships with injection
  recovery *off* (0.0/0.0). Our 0.2/0.005 enables it at 2×/5× the values the source code itself
  suggests as "good" (0.1/0.001), which makes AMCL quicker to declare itself lost and inject
  random particles — combined with `always_reset_initial_pose: true`, this config leans heavily
  toward "recover from being lost" at the cost of occasionally destabilizing a good estimate.
- **One inaccuracy in the file's comments:** the note next to `resample_interval` implies the
  alternative was `update_min_a/d` of 0.0/0.0, and the `max_particles` comment compares against
  6000 — those describe the *previous* config, not Nav2 defaults (which are 0.2/0.25 and 2000).
  Fine as history, just don't read them as "what stock does."
- The inflated alphas (0.3–0.4 vs 0.2) say "trust odometry less" — reasonable for a differential
  robot with wheel slip, and consistent with the tighter `update_min_d`.

karmanyaahm: we should test localization thorughly

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
```diff
@@ -1,7 +1,13 @@
...
   ros__parameters:
     use_sim_time: false
-    controller_frequency: 20.0
+    # 10 Hz: 20 Hz MPPI was not achievable on this CPU (constant "Control loop
+    # missed its desired rate" spam), and every overrun past the velocity
+    # smoother's timeout zeroed cmd_vel -> stop-start stutter.
+    controller_frequency: 10.0
+    # Tolerate brief controller hiccups instead of aborting the whole goal on
+    # the first missed cycle (seconds; 0.0 = abort immediately).
+    failure_tolerance: 1.0
```
karmanyaahm: I'm not super sure about this, was this tested under pure robot conditions or with other stuff running at the same time. afaik it used to run fine back in the day but i might be misremembering

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
```diff
@@ -32,14 +38,15 @@
       
       # Motion model and basic parameters
       motion_model: "DiffDrive"
-      time_steps: 50                    # 50 steps * 0.05s = 2.5s prediction horizon (matches old sim_time)
-      model_dt: 0.05                    # 1/controller_frequency = 1/20 = 0.05s
-      batch_size: 2000                  # Number of trajectory samples
+      time_steps: 25                    # 25 steps * 0.1s = same 2.5s prediction horizon at half the compute
+      model_dt: 0.1                     # 1/controller_frequency = 1/10 = 0.1s
+      batch_size: 1500                  # Number of trajectory samples
```
karmanyaahm: how thoroughly has nav been tested with this?

ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
```diff
@@ -36,49 +38,9 @@ navigation:
...
-            clearing: true
-            min_z: 0.0                   #default 0, meters
-            max_z: 7.0                   #default 10, meters
-            vertical_fov_angle: 0.7      #default 0.7, radians
-            horizontal_fov_angle: 1.04   #default 1.04, radians
-            decay_acceleration: 1.       #default 0, 1/s^2. If laser scanner MUST be 0
-            model_type: 0                #default 0 (depth camera). Use 1 for 3D Lidar
+        # (A full voxel_layer block used to sit here unused — it was never in
+        # the plugins list above. The camera voxel layer lives in the local
+        # and mapfree costmaps.)
```
karmanyaahm:
```suggestion

```

ros2_ws/src/mars_bot/mars_nav/config/nav_through_poses.xml
```diff
@@ -1,7 +1,53 @@
...
+                    <!-- Costmap clears can't fix a bad AMCL pose (the static layer
+                         repaints map speckle immediately); a spin feeds AMCL fresh
+                         scan matches so localization can reconverge. -->
+                    <Spin spin_dist="1.57"/>
+                    <BackUp backup_dist="0.15" backup_speed="0.3"/>
+                    <Wait wait_duration="2"/>
+                </RoundRobin>
+            </ReactiveFallback>
+        </RecoveryNode>
+    </BehaviorTree>
```
karmanyaahm: Let's always return failure instead. This seems like a nice feature, but shipping something without testing doesn't seem like a good idea since users will assume it's a reliable primitive.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
```diff
@@ -2,28 +2,18 @@
...
+        # before planning, so an AMCL pose that drifts into an inflated
+        # obstacle degrades gracefully instead of aborting instantly with
+        # "Starting point in lethal space" (Smac2D has no start tolerance).
+        plugin: "nav2_navfn_planner/NavfnPlanner"
+        tolerance: 0.25
+        # Unknown cells are traversable at high cost; with speckled or partial
+        # maps a hard 'false' rejects plans through perfectly driveable space.
+        allow_unknown: true
+        use_astar: false
```
karmanyaahm: are there other differences in the planners? probably right. IMHO we should split up this pr into multiple smaller ones and test the effects of each separately.

ros2_ws/src/mars_bot/mars_nav/mars_nav/grid_localizer.py
```diff
@@ -501,6 +513,57 @@ def _publish_pose(self, x: float, y: float, theta: float):
...
+                future.result()
+                self.get_logger().info("Seeded AMCL via /set_initial_pose")
+                if not superseded:
+                    self._pending_seed = None
+            except Exception as e:
+                self.get_logger().warning(f"AMCL /set_initial_pose call failed: {e}")
+                if not superseded and self._seed_retry_timer is None:
+                    self._seed_retry_timer = self.create_timer(2.0, self._retry_seed_amcl)
+
+        self._amcl_seed_client.call_async(request).add_done_callback(_on_done)
```
vignesh-anand: **No retry if the service call is accepted but never completes.**

The retry machinery covers two cases — service not ready (arms a timer) and the call returning an exception (`_on_done` re-arms). It does *not* cover "the call was accepted but never replies."

`rclpy` `call_async` futures have no built-in timeout, and a client does not proactively fail pending futures when the server disappears. So if AMCL dies (or its lifecycle bounces) in the window *after* `service_is_ready()` returns true but *before* it replies — precisely the boot/mode-switch race this feature exists to survive — then `_on_done` never fires, `_pending_seed` stays set forever, and no timer is armed to poll it. Because AMCL subscribes `/initialpose` VOLATILE, the latched topic won't redeliver to a restarted AMCL either, so the seed is silently lost with no further attempt (on plain auto-localize boot there is no other trigger).

Severity moderate — it's a fallback-of-a-fallback and low-probability — but a genuine hole. Arming the retry timer on this path too (or a `wait_for`-style deadline) would close it.

ros2_ws/src/mars_bot/mars_nav/mars_nav/grid_localizer.py
```diff
@@ -197,6 +206,9 @@ def _cleanup_resources(self):
         if self._map_check_timer:
             self.destroy_timer(self._map_check_timer)
             self._map_check_timer = None
+        if self._seed_retry_timer:
```
vignesh-anand, line 215: **`_seed_retry_timer` isn't cancelled on deactivate.**

`_cleanup_resources` (here) destroys `_seed_retry_timer`, but `on_deactivate` (grid_localizer.py:184-197) cancels only `_auto_timer` and `_map_check_timer`. So across an Active→Inactive transition with a retry in flight, the 2 s timer keeps firing and `_retry_seed_amcl` → `_seed_amcl` will call AMCL's service from a deactivated node. `_pending_seed` is also never cleared on deactivate.

During a map switch (deactivate → reactivate grid_localizer), a stale retry carrying the *previous* map's pose can survive into the reactivated node and land in AMCL. The new `/localize` should supersede it via the identity check, but only if it arrives before the stale retry succeeds.

Cleanest fix: cancel `_seed_retry_timer` and null `_pending_seed` in `on_deactivate` alongside the other two timers.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -667,6 +692,38 @@ def _efficient_map_switch(self) -> tuple[bool, str]:
...
             return True, f"Map switched successfully to {self.current_map}"
 
+    def _trigger_relocalization(self):
+        """Ask grid_localizer for a fresh pose estimate after a map switch.
+
+        Without this the switch keeps AMCL's pose from the previous map. Waits
+        briefly so grid_localizer has received/processed the newly latched map
+        before searching. Failure is logged but never fails the map switch.
+        """
+        time.sleep(1.0)
```
vignesh-anand: **Map-switch relocalization is a timing heuristic with a hard ceiling.**

Two load-sensitive soft spots here:

- The fixed `sleep(1.0)` assumes grid_localizer has received *and processed* the newly latched map within 1 s. If delivery/processing is slower, `/localize` runs `_find_pose` against the **old** map — a wrong seed that still reports `success`. There's no confirmation grid_localizer actually swapped maps first.
- `_localize_cb` runs the blocking GPU search on grid_localizer's single-threaded executor. If that executor is already mid-search (e.g. an auto-localize tick), the request queues and the 8 s can expire — AMCL then silently keeps the old map's pose (the exact failure this is meant to prevent), logged only as a warning.

Functionally fine and gracefully degrading, but the two magic numbers (1 s, 8 s) are sensitive to load. Waiting on a positive "map swapped" signal from grid_localizer instead of a fixed sleep would harden this.

karmanyaahm (dismissed):
files without comments lgtm. IMHO it would be better if you leave all the obvious code stuff/cleanup in this PR, and move all the nav param changes to individual PRs for the planner vs controller vs localization

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -682,31 +739,46 @@ def change_map_callback(self, request, response):
...
+            # Serialize against mode changes: both callbacks drive lifecycle
+            # transitions on the same nodes, and interleaving them leaves a
+            # half-active stack (e.g. one thread deactivating bt_navigator
+            # while the other activates it).
+            relocalize_after_switch = False
+            with self._mode_change_lock:
+                previous_map = self.current_map
+                # _efficient_map_switch loads self.current_map, so set it for
+                # the attempt but only persist (and keep) it on success.
+                self.current_map = requested_map
```
vignesh-anand, line 806: **The serialization is incomplete — a third writer of `current_map` isn't covered by this lock.**

This block correctly serializes `change_map` ↔ `change_mode` on `_mode_change_lock`. But `delete_map_callback` also mutates `self.current_map` (line 832: `self.current_map = fallback`; then `save_last_map(fallback)`) with **no lock**, and its service is on the **default** callback group (line 157, no `callback_group=`) while `map_service` is on `_internal_callbacks_group` (line 146). Different groups → the executor can run them concurrently.

So while a locked `change_map` is mid-switch — having set `current_map = requested_map` here and calling `_efficient_map_switch` → `_load_map_on_server`, which reads `self.current_map` to build the map path (line 630) — a concurrent `delete_map` can overwrite `current_map` with its fallback. Result: the switch loads the wrong map, or persists a `current_map` that doesn't match what map_server actually loaded.

Narrow window (deleting during a switch is uncommon), but it's exactly the interleave class this fix targets — the boundary was drawn around two of the three writers. Suggest having `delete_map_callback` (and the `save_map` current-map path) take `_mode_change_lock` too, or at least guard the `current_map` mutation.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -682,31 +739,46 @@ def change_map_callback(self, request, response):
...
                 return response
 
-            # Update the current map
-            self.current_map = requested_map
+            # Serialize against mode changes: both callbacks drive lifecycle
+            # transitions on the same nodes, and interleaving them leaves a
+            # half-active stack (e.g. one thread deactivating bt_navigator
+            # while the other activates it).
+            relocalize_after_switch = False
+            with self._mode_change_lock:
```
vignesh-anand: **Blocking `acquire()` + busy-poll `call_service` on a shared 4-thread pool can self-starve under concurrency.**

mode_manager and the router share one `MultiThreadedExecutor(num_threads=4)` (line 1178). This is a **blocking** acquire and `_internal_callbacks_group` is reentrant, so N concurrent `change_map` requests occupy N threads: one runs the switch, the rest block on `.acquire()` while each holding a thread.

The thread running the switch drives lifecycle transitions via `call_service`, which **busy-polls `future.done()` with `time.sleep(0.1)`** (service_utils.py:34-40) — it doesn't spin, so it needs *another* executor thread to process the lifecycle responses. With 4 concurrent map changes: 1 running+polling, 3 blocked on the lock → zero threads free to service the responses → the running switch can't complete until its 8–17s timeouts fire.

The PR verifies **two** concurrent map changes (1 runs, 1 blocks, 2 threads free — fine). The safe-concurrency margin is just 2 spare threads, shared with the router. Not a blocker, but worth documenting the ceiling; longer-term the busy-poll `call_service` should await rather than sleep-poll.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -682,31 +739,46 @@ def change_map_callback(self, request, response):
...
-
-                if success:
+                    if success:
+                        self.save_last_map(requested_map)
+                        relocalize_after_switch = True
+                        response.success = True
+                        response.message = f"Successfully changed map to '{requested_map}'"
+                        self.get_logger().info(response.message)
+                    else:
+                        self.current_map = previous_map
```
vignesh-anand, line 821: **Rollback lives only in the `else` — an exception during the switch leaves `current_map` on the failed map.**

`previous_map` is captured and restored only on the `False` return path (here). If `_efficient_map_switch` / `_load_map_on_server` **raises** instead of returning `(False, …)`, control unwinds out of the `with` block to the outer `except` (line 783), where `previous_map` is out of scope and no rollback runs — `current_map` is left at `requested_map` in memory (unpersisted, since `save_last_map` wasn't reached).

In practice `transition_node` catches internally and returns `False`, so this is unlikely to fire today — but the rollback is structurally in the wrong place. Capturing `previous_map` and restoring it in a `try/except` around the switch would make the invariant hold unconditionally.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
theo-michel: okay interesting!

ros2_ws/src/mars_bot/mars_nav/mars_nav/grid_localizer.py
theo-michel: Fixed in e8aff1af: the in-flight call now has a 5 s watchdog timer — if the future never completes (AMCL died after accepting), the pending request is removed via remove_pending_request and the seed is retried through the normal path. The watchdog is armed before the done-callback so an immediately-completed future disarms it instead of racing it.

ros2_ws/src/mars_bot/mars_nav/mars_nav/grid_localizer.py
theo-michel: Fixed in e8aff1af: on_deactivate now destroys _seed_retry_timer and clears _pending_seed, so a retry carrying the previous map's pose cannot survive a deactivate/reactivate cycle.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
theo-michel: Fixed in e8aff1af: the fixed sleep + blind /localize is gone. mode_manager now subscribes to /localization/status and waits (up to 6 s) for a status newer than the switch start — grid_localizer re-localizes itself on every new /map (its _map_cb restarts auto-localization), so the primary path is pure confirmation and cannot race the map handoff. The explicit /localize Trigger only fires as a fallback when no status arrives at all. Verified live: switch completes in 1.6 s with 'Grid localizer re-localized on the new map (status: localized_low_confidence)' and the AMCL seed delivered.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
theo-michel: Fixed in e8aff1af: delete_map_callback is now on the same reentrant group and takes the same _mode_change_lock (non-blocking, busy response) before touching current_mode/current_map.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
theo-michel: Fixed in e8aff1af: change_map_callback now uses acquire(blocking=False) with a busy response, matching change_mode — concurrent map changes no longer pin executor threads waiting on the lock. (service_utils' polling wait is pre-existing and shared with the mode path; left as-is here.)

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
theo-michel: Fixed in e8aff1af: the switch body is wrapped in try/except that rolls current_map back to previous_map on any exception before re-raising to the outer handler, with the lock released in finally.

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
theo-michel: Tested under full production conditions: the complete stack running on the robot (cameras, depth, brain/skills server, webapp, WebRTC) at load average ~10-12. Under that load the 20 Hz loop missed its rate continuously (log spam every few hundred ms), and each overrun past the smoother's 0.2 s timeout zeroed cmd_vel. You may well be right that 20 Hz worked on a lighter stack back in the day — the difference is today's concurrent workload, not the controller itself. At 10 Hz there are still occasional misses under peak load, but failure_tolerance 1.0 + the 0.5 s smoother timeout absorb them: verified goal runs with zero aborts and no stutter.

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
theo-michel: The 0.4 cap itself changes nothing physically: bringup already hard-clamps every /cmd_vel at safety.max_speed 0.4 (bringup.py _cmd_vel_callback), so the motors never executed 0.45 — MPPI was just planning with a speed the base silently refused. Nav testing with the full retune: repeated navigate_to_position runs (map-frame and local-frame) plus a mid-drive cancel test on the robot, all passing. Happy to run a longer soak if you want more mileage before merging.

ros2_ws/src/mars_bot/mars_nav/config/nav_through_poses.xml
theo-michel: Done in e8aff1af — nav_through_poses.xml is now <AlwaysFailure/> with a comment explaining why (untested primitive, navigator can't be disabled on Humble). Goals now fail loudly instead of fake-succeeding.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
theo-michel: Main behavioral differences Smac2D -> NavFn: (1) NavFn clears the robot's start cell before planning, which is the property we need — a drifted AMCL pose inside inflated speckle degrades instead of instantly aborting (Smac2D has no start tolerance and throws 'Starting point in lethal space'); (2) NavFn is Dijkstra/A* potential-field without Smac's built-in smoother, so paths are slightly rougher (MPPI doesn't care much); (3) unknown-space traversal enabled here (allow_unknown). On splitting the PR: fair point — each change was verified individually on the robot, but if you'd rather land it piecewise I can split into (a) amcl/planner/localization chain, (b) controller/smoother retune, (c) router+mode_manager robustness. Say the word.

ros2_ws/src/mars_bot/mars_nav/config/costmap.yaml
theo-michel: Removed in e8aff1af.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
theo-michel: Fair — reverted in bc3f448c: the GridBased block is restored byte-for-byte from the tested config (only `costmap_update_timeout`/`service_introspection_mode` stay removed; the Humble binary never declared those, so they were inert).

For the record, fresh A/B data from the robot after the revert, with the localization chain from this PR active (AMCL cov ~0.06, seeding, spin recovery): map-frame goals on the current Test-map abort instantly again under Smac2D ('Starting point in lethal space', 12 aborts in one session), where NavFn had gone 6/6 on the same map minutes earlier. The mechanism is not localization anymore — the robot's believed pose sits at cost 99 inside an *inflated phantom speckle blob* (this map has ~600 scattered noise pixels and no coherent walls), so no pose-quality improvement can avoid it; Smac2D has no start tolerance on Humble and there's no knob for one.

So with Smac2D restored, the residual failure mode is map quality. Options, your call:
1. Treat it as a map problem: re-record Test-map (it's known-garbage) and accept that badly speckled maps can produce lethal-start aborts. The Spin recovery + retries in this PR soften but don't eliminate it.
2. Dual-planner BT: keep Smac2D as the primary `GridBased` and add NavFn as a fallback planner tried only when Smac2D fails — preserves the tested planner for every normal plan, uses NavFn purely as a lethal-start escape hatch. ~15 lines of config+BT if you want it.
3. If you can dig up the original reason NavFn was rejected, happy to re-run that testing.

The PR as pushed is option 1 (pure restore).

theo-michel: Reverted to old planner

ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
theo-michel: You're right, and thanks for the history — 'regression since cfdf21d4, not dead config' changes the calculus completely: those values have ~10 months of production behind them, which beats my theoretical concerns about update_min_d/a 0.0. Done in 3c5bb8b0: the re-key stays, and every value is restored byte-identical to the original file (verified live: navigation_amcl now reports max_particles=6000, update_min_d=0.0, resample_interval=1, always_reset_initial_pose=true — the original tuning applied for the first time since January). If a retune is ever wanted it should be its own tested change.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
theo-michel: The comment is gone with the NavFn revert (bc3f448c), but noted for if the fallback option is ever adopted. One small counterpoint on the mechanism: Humble's nav2_navfn_planner does actively clear the start — NavfnPlanner::makePlan calls clearRobotCell() on the robot's cell before computing the potential — in addition to having no explicit start-in-lethal rejection guard. So it's both: an active clear of the start cell AND graceful-by-omission for near-lethal surroundings. Happy to use your wording (plus the clearRobotCell note) if this ever comes back.

ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
theo-michel: Great audit — note it now describes an intermediate state: after vignesh's history note (the values are a long-serving hand-tuned set, orphaned by the cfdf21d4 rename rather than never-tested), 3c5bb8b0 restored the original file values byte-identical, so max_particles is back to 6000, update_min_d/a to 0.0, resample_interval to 1, laser_max_range to 10.0. Rows that still differ from Nav2 defaults (alphas, recovery_alpha_fast/slow 0.2/0.005, laser ranges, always_reset_initial_pose) are the historical tuning, deliberately untouched by this PR. Your point that the recovery alphas are aggressive vs the source's suggested 0.1/0.001 is worth keeping — but that belongs to a dedicated retune-with-testing change, not this re-key.

ros2_ws/src/mars_bot/mars_nav/mars_nav/grid_localizer.py
```diff
@@ -197,6 +206,9 @@ def _cleanup_resources(self):
...
+            self.destroy_timer(self._seed_retry_timer)
+            self._seed_retry_timer = None
```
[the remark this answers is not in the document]
theo-michel: Fixed in 3c5bb8b0: _cleanup_resources now clears _pending_seed alongside destroying the retry timer (exactly the suggestion), matching what on_deactivate already did.

ros2_ws/src/mars_bot/mars_nav/config/amcl.yaml
theo-michel: Agreed. Testing so far: repeated map-frame + local-frame goal runs, map switches with relocalization confirmation, and a mid-drive cancel, all on the live robot. Note the localization params are now byte-identical to the long-serving hand-tuned set (3c5bb8b0), so the only behavioral delta left in this PR is the seeding/relocalization machinery — which was live-verified. A dedicated localization soak (multiple environments, kidnapped-robot cases) is a good follow-up once the map is re-recorded.

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
theo-michel: Per the split request, this retune moved out of this PR entirely (54902867 reverts it here) and re-landed as #510 with the test evidence — this PR no longer touches controller params.

ros2_ws/src/mars_bot/mars_nav/config/controller.yaml
theo-michel: Moved to #510 with the rest of the param retunes (54902867 reverts it here); testing details are in that PR's description.

ros2_ws/src/mars_bot/mars_nav/config/planner.yaml
theo-michel: Split done: this PR now carries only code fixes, BT correctness, and dead-config cleanup (54902867); all param retunes re-landed as #510. No planner or localization param deltas remain anywhere (Smac2D and the AMCL values are byte-identical to main/original).

theo-michel:
@karmanyaahm split done as requested:

- **This PR (#508)** now contains only the code fixes, BT correctness, and dead-config cleanup: AMCL re-key (values byte-identical to the long-serving hand-tuned set), the seeding/relocalization machinery, router cancel race, mode/map/delete serialization, Spin recovery, `nav_through_poses` → AlwaysFailure, mobility busy-wait, and the removals of config the Humble binaries never read. Commit 54902867 reverts every param retune here.
- **#510** (stacked on this branch) carries all the nav param retunes — controller 10 Hz + tolerances, velocity caps at the hardware clamp, costmap update rates, scan throttle, `simulate_ahead_time` — one bullet per item with the on-robot evidence, so each can be tested or dropped independently.
- There are no planner or localization *param* changes left anywhere: Smac2D was restored verbatim (bc3f448c) and the AMCL values are byte-identical to the original tuning (3c5bb8b0), so no third/fourth PR is needed unless you want the recovery-alpha discussion from the audit thread turned into one.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -682,31 +768,61 @@ def change_map_callback(self, request, response):
...
+                self.current_map = previous_map
+                raise
+            finally:
+                self._mode_change_lock.release()
+
+            # Outside the lock: relocalization waits up to ~12 s on the grid
+            # localizer and must not block concurrent mode/map service calls —
+            # the lifecycle switch itself is already complete.
+            if relocalize_after_switch:
+                self._trigger_relocalization(since=switch_started)
```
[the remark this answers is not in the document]
theo-michel: Fixed in be2b7f61: _trigger_relocalization bails out (both from the status wait and before the fallback /localize) as soon as _mode_change_lock is taken again — a newer switch owns relocalization from that point, so a stale waiter can no longer claim the newer map's status as its own confirmation.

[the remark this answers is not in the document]
theo-michel: Fixed properly in beb85014 (the earlier lock.locked() guard had a TOCTOU hole): every lock holder now bumps a _switch_generation counter under the lock, and the relocalization waiter carries the generation it started with — any mismatch means superseded, regardless of whether the newer switch still holds the lock.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -667,6 +705,66 @@ def _efficient_map_switch(self) -> tuple[bool, str]:
...
+        Failure is logged but never fails the map switch.
+        """
+        deadline = time.monotonic() + 6.0
+        while time.monotonic() < deadline:
+            # A newer mode/map change supersedes this one: its localization
+            # (against the newest map) is the one that matters, and both the
+            # status feed and a fallback /localize would now refer to that
+            # newer map — stop confirming on behalf of a stale switch.
+            if self._mode_change_lock.locked():
+                self.get_logger().info(
```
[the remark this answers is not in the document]
theo-michel: Fixed in beb85014 with exactly the suggested mechanism: a switch generation counter. All three lock holders (map change, mode change, delete_map's current_map fallback) bump _switch_generation under _mode_change_lock; the waiter checks its captured generation in the poll loop and before the fallback /localize, so a newer switch that acquired and released the lock between polls is still detected. Verified live: a normal map switch still confirms ('Grid localizer re-localized on the new map').

PR #517, opened 2026-07-10
fix(nav): survive controller_server crashes — respawn + lifecycle watchdog

DavidDobas:
## Problem

`controller_server` crashes in the field and nothing brings it back — navigation is silently dead until reboot. Observed live twice in one sim session, same signature both times:

```
[controller_server]: Failed to make progress
[controller_server]: [follow_path] [ActionServer] Aborting handle.
[controller_server-6] double free or corruption (!prev)
process has died [exit code -6]
```

Trigger: repeated progress-checker aborts (the robot stuck/rotating near clutter — a state this stack reaches routinely). After the crash, every navigation turns into recovery theater: the BT times out waiting for `follow_path` to acknowledge, cycles clear/spin/backup/wait, and aborts — indistinguishable from "navigation mysteriously broken."

## Root-cause investigation (summary)

- **It's a thread race in upstream code, not ours**: valgrind memcheck observed **21 executions of the abort path with zero invalid memory operations** — memcheck serializes threads, and serialized execution can't race; free-running execution crashes. Under gdb the timing shift also hides it.
- The race surface is nav2_util's `SimpleActionServer`: the control loop runs in a detached `std::async` thread and calls `terminate_current()` on failure while the executor thread concurrently handles preempts/cancels/expiry on the same goal handles.
- It is **not** the already-fixed rclcpp action race (ros2/rclcpp#2250, backported long ago — we run rclcpp-action 16.0.19 which includes it). This is an additional, unfixed race somewhere in nav2_util / rcl_action / rmw_zenoh.
- This is a sibling of the crash family already documented in `mode_manager.py`'s `skip_cleanup_nodes` ("Zenoh RMW crashes during TF unsubscription in cleanup", planner "Segfaults during costmap cleanup").
- Real-robot exposure is identical: same ARM64 Humble debs (`ros-humble-nav2-controller 1.1.20`), same `rmw_zenoh_cpp 0.1.8`, same abort-generating behaviors. An upstream issue with the reproduction recipe is the path to a true fix; this PR makes the crash survivable in the meantime (and for whatever the next such bug is).

## Fix

Two parts, both generic:

1. **`respawn=True, respawn_delay=2.0`** on `controller_server` (`navigation.launch.py`) — launch brings the process back.
2. **Lifecycle watchdog in `mode_manager`** — a respawned lifecycle node comes back `UNCONFIGURED` and useless. A 5 s watchdog checks the current mode's nodes and re-drives anything `UNCONFIGURED` to where the mode needs it (`ACTIVE`, or `INACTIVE` for configure-only nodes). It arms only after the first successful mode startup (never races bringup) and yields to in-flight mode/map operations via the existing `_mode_change_lock`. It heals **any** respawn-configured nav node, not just the controller.

## Verification (live, full digital twin)

Simulated the exact crash (`SIGABRT` to `controller_server`) during an active session:

```
T0      SIGABRT
T0+2s   launch respawn: "process started with pid [56382]"
T0+6s   watchdog: "controller_server is UNCONFIGURED while navigation mode is
        active -- it likely crashed and respawned; re-driving its lifecycle"
        "controller_server restored to ACTIVE after respawn"
T0+40s  next NavigateToPose goal: SUCCEEDED
```

A crash that previously cost the rest of the session now costs ~6 seconds and leaves a loud log line.

## Real-robot testing notes

- To exercise the heal path on hardware without waiting for a natural crash: `pkill -ABRT -f nav2_controller` mid-session, then watch for the watchdog log lines and send a goal.
- The watchdog acts only on `UNCONFIGURED` nodes of the *current* mode, so normal mode switches, map switches and bringup are untouched (they hold `_mode_change_lock`, which the watchdog defers to).

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
@@ -381,6 +381,7 @@ class BrainClientNode(Node):
         primitive_name = payload.get("skill_name") or self._skill_name_for_id(skill_id)
         primitive_id = payload.get("primitive_id") or f"manual_{skill_id}_{int(time.time() * 1000)}"
         reason = payload.get("reason")
+        inputs = payload.get("inputs")
 
         self.get_logger().info(f"Manual skill event: {status} {primitive_name} ({skill_id})")
         if self.state.is_brain_active:
@@ -398,6 +399,7 @@ class BrainClientNode(Node):
             status=status,
             skill_id=skill_id,
             reason=reason,
+            inputs=inputs,
         )
         self.chat.history.append(
             {
@@ -408,6 +410,7 @@ class BrainClientNode(Node):
                 "primitiveId": primitive_id,
                 "skillId": skill_id,
                 **({"failureReason": reason} if reason else {}),
+                **({"inputs": inputs} if inputs else {}),
             }
         )
 
--- a/ros2_ws/src/brain/brain_client/brain_client/nodes/skills_server.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/nodes/skills_server.py
@@ -322,7 +322,7 @@ class SkillsActionServer(Node):
 
         name = self._skill_display_name(skill_type)
         run_id = uuid.uuid4().hex
-        self._publish_skill_status(run_id, skill_type, name, "running")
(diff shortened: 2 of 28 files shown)


DavidDobas:
Wrong head branch (created from the debug branch by mistake) — superseded by the PR from fix/nav-controller-crash-resilience.

PR #524, opened 2026-07-13
ci: publish multi-arch simulator images (amd64 + arm64)

DavidDobas:
## Why

The prebuilt simulator images are **amd64-only** (Cloud Build runs a plain `docker build` on an x86 machine) — so Apple-Silicon Macs, arm64 VMs, and Raspberry Pis have never used a prebuilt image: every `./innate-sim up` on them pays a ~25–30 min local Docker build (plus ~10 GB of build cache) that amd64 users skip with a ~5 min pull. Since Apple-Silicon Macs are likely the primary developer machine, that's the main audience.

This PR also publishes from **every branch tip**, not just main, so checking out any branch whose CI has run gets a pull instead of a local build.

## How

Four jobs; amd64 and arm64 build in parallel:

1. **tags** — content-hash computation (now via the shared `ci/compute_sim_image_inputs_hash.py`), exposed as job outputs.
2. **amd64** — Cloud Build, as before, via the shared build script, pushing arch-suffixed tags (`inputs-<hash>-amd64`, …).
3. **arm64** — the org's native `my-medium-runner-arm` larger runner (8 cores / 32 GB / 300 GB SSD; billed per-minute — see `.github/workflows/sim-images-build-cost-analysis.md`) running the *same* script with `ARCH=arm64`. Native, because QEMU-emulating the ros2_ws colcon build would take hours.
4. **manifest** — `ci/stitch_sim_manifests.sh` stitches each pair into a multi-arch manifest under the **unsuffixed tags every consumer already pulls** (`inputs-<hash>`, `sha-…`, `main`). One `imagetools create` per image carries all tags, so a cancelled run can't leave some tags multi-arch and others stale. Docker resolves the right architecture automatically — no launcher changes needed here.

Build logic is deduplicated into `ci/build_sim_images.sh`, called by both builders; the Cloud Build config shrinks to login + invoke.

## Every-branch publishing

- `on.push` now has no branch filter (paths only, and the path list covers exactly the hash inputs — including `.dockerignore`, previously missing). Non-main pushes publish `inputs-<hash>`/`sha-…` tags but never `main`.
- Per-ref concurrency with cancel-in-progress: only a branch's tip build completes; rapid pushes don't pile up.
- Volume estimate from `format.yml`'s pull_request-synchronize runs: ~400–600 publishes/month vs ~50 in the main-only era.

## Security

- Triggers are `push` + `workflow_dispatch` ONLY — never `pull_request`/`pull_request_target` (contract documented at the top of the workflow). PRs cannot run this workflow at all.
- Defense in depth: every job requires `github.repository == 'innate-inc/innate-os'` and a push/dispatch event, so a future trigger edit or a fork can't reach the pushing jobs.
- Permissions scoped per job: `id-token: write` only on the Cloud Build leg, `packages: write` only on arm64 + manifest.

## Transition & compatibility

- Build caches are per-arch (`buildcache-amd64`/`-arm64`), seeded from the old unsuffixed `buildcache` on first run (a one-time no-op pull of useless amd64 layers on the arm64 side; self-heals from run two).
- Unsuffixed tags simply gain an arm64 half; old and new launchers both work.

## Validated

Full end-to-end run on this branch ([29300662258](https://github.com/innate-inc/innate-os/actions/runs/29300662258)): tags 4 s → amd64 2.5 min ∥ arm64 10 min (fully cold) → manifest 15 s, all green, ghcr pushes authorized. Since this branch's inputs-hash equals main's, main's `inputs-…` tags are **already multi-arch** — the intended end state. Warm arm64 runs should land ~3–5 min.

Shared scripts were also mock-run with a stubbed docker: correct tag set per arch, cache fallback chain, and stitch commands (including dedup when `IMAGE_TAG` equals the sha tag) for both main and manual-dispatch cases.

## Related PRs

- #526 fixes the launcher's inputs-hash drift — **without it the launcher never matches any published tag** (pre-existing bug; every platform currently falls back to local builds). This PR makes the tags multi-arch; #526 makes launchers find them.
- #525 adds age/branch-aware ghcr retention sized for every-branch publish volume.

### DIFF
--- a/.github/workflows/cleanup-sim-images.yml
+++ b/.github/workflows/cleanup-sim-images.yml
@@ -25,6 +25,9 @@ concurrency:
 jobs:
   cleanup:
     runs-on: ubuntu-latest
+    # SECURITY: deletes ghcr package versions -- schedule/dispatch on this
+    # repo only, never any pull_request trigger.
+    if: github.repository == 'innate-inc/innate-os' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
     env:
       ORG: innate-inc
       # Retention policy (deps/ros; every branch tip publishes images):
--- a/.github/workflows/cleanup-sim-images.yml
+++ b/.github/workflows/cleanup-sim-images.yml
@@ -2,7 +2,7 @@ name: Cleanup Simulator Images
 
 on:
   schedule:
-    - cron: "17 3 * * *"
+    - cron: "17 3 * * 0"
   workflow_dispatch:
     inputs:
       dry_run:
@@ -15,7 +15,6 @@ on:
           - "false"
 
 permissions:
-  contents: read
   packages: write
 
 concurrency:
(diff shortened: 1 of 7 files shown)


karmanyaahm (approved):
yo

PR #527, opened 2026-07-14
Add camera calibration webapp page and manual-capture calibration flow

theo-michel:
## Summary
- Reworks the stereo calibrator's managed ROS2 action into a manual, keypress/button-driven capture loop — drops rosbag playback and the arm "insert board" handoff, which are no longer part of the intended calibration flow.
- Adds a capture-timeout watchdog (`capture_timeout_sec` ROS param, default 60s, live-adjustable via `ros2 param set`) that aborts an orphaned run if no capture arrives in time — guards against a stuck goal if the requesting webapp client disconnects. Starting a new calibration goal now also kills whatever run is currently active (waits for it to fully exit before accepting the new goal) instead of rejecting the new request.
- Feedback now reports the watchdog's live `capture_timeout_sec` on every tick (including one fired immediately at goal acceptance) and Result carries a structured `timed_out` flag — the webapp uses these to show a server-anchored countdown that re-syncs on every tick rather than running its own independent clock, and never declares the run expired itself (only the real Result does).
- Streams per-capture "coverage dot" debug visualizations (left/right JPEGs) back as action feedback, alongside progress counters and corners-detected status.
- Surfaces the EXCELLENT/GOOD/ACCEPTABLE/POOR RMS-threshold quality assessment (previously only logged server-side) in the Result, rendered as a color-coded stat in the webapp.
- Adds a new **Calibration** page to the webapp (`/calibration`), reusing the existing live WebRTC camera-feed widget, with Start/Capture/Stop controls, live feedback display, a result banner, and a depth-topic-based "calibration file detected" indicator (no new ROS surface needed for that).
- Switches the capture-trigger topic from `std_msgs/Empty` to `std_msgs/Bool`: publishing a zero-field message from a browser client through `rws` (our rosbridge-compatible server) crashes the subscriber node (Fast-CDR deserialize exception) — confirmed via live testing. Filed separately for a real fix in `ros-websocket` (it's an internal repo, not third-party); this switch is the workaround on our side.
- Mobile fix: the video feed collapsed/letterboxed on narrow viewports once the feedback panel grew taller than the video's ambiguous auto-sized grid row — now sized from the stream's actual 4:3 aspect ratio.
- "Save calibration when done" defaults to checked; default images-to-capture bumped 30 → 10 → 20 over the course of review.

## Test plan
- [x] `innate build mars_msgs mars_cam` — clean build
- [x] Verified new action/topic/service interface live (`ros2 action info`, `ros2 interface show`, `ros2 topic type`)
- [x] Exercised the full manual-mode flow live against the robot: goal accepted → capture triggered via the topic → feedback received with real coverage-dot JPEGs → stop path aborts cleanly (fixed an invalid `goal_handle.canceled()` transition this exposed) → capture-timeout watchdog fires and is live-adjustable via `ros2 param set`
- [x] Reproduced and fixed the `rws`-crashes-on-Empty bug by switching to `std_msgs/Bool`; re-verified consecutive capture triggers over rosbridge with no crash
- [x] Verified the READY feedback tick fires immediately at goal acceptance with the live `capture_timeout_sec`, and Result carries `timed_out: true` on an actual watchdog timeout
- [x] Verified starting a second goal while one is active kills the first (head restored, goal aborted) and the new one takes over cleanly, with no crash
- [x] Reproduced the mobile video-collapse bug and confirmed the fix with a real iPhone-viewport Playwright run (video box stayed a stable 4:3 before/after Start, where it previously collapsed)
- [x] Webapp page verified in a real browser against the live robot (Playwright): nav entry, live video, Start/Capture/Stop/result flow, coverage images decoding and rendering


karmanyaahm (approved):
i wrote this wrong account

theo-michel:
Great job me

PR #528, opened 2026-07-14
perf(sim): world-server memory diet -- 2.5GB -> 0.8GB, warm boots in 0.1s

DavidDobas:
> Stacks on #522 — the commit list collapses to the two `perf(sim)` commits once that merges.

## Why

The world server measured **2.3–2.5 GB RSS under Linux software GL** — on the 4 GB Raspberry Pi that made it the kernel's eviction victim (`VmRSS 68MB / VmSwap 1.0GB` measured live), putting the physics loop into permanent page-fault stalls ("dropping 0.5s of physics backlog" forever). It was half the machine, and half of the entire stack's memory demand.

Attribution (measured, macOS + Lima VM + Pi):

| Consumer | Cost |
|---|---|
| Five 4096×4096 baked-lighting atlases, decoded in the MjModel | ~252 MB |
| …uploaded **again** into the RGB renderer context (process RAM under software GL) | ~130–190 MB |
| …uploaded **again** into the depth renderer context (which never samples them) | ~180–245 MB |
| `MjSpec.compile` scratch (qhull over 1300+ hulls, texture decode) that glibc never returns | ~0.5–1 GB |

## What

1. **Texture cap** — visual-room textures are downscaled through a cached copy next to the original (assets dir is gitignored): **2048** at full render res — visually indistinguishable at 640×480, the parquet grain survives — and **1024** at half-res or below, where the difference physically can't reach the wire. `VIRTUAL_MARS_TEXTURE_MAX` overrides; `0` keeps the source resolution.
2. **`release_freed_heap()`** — glibc `malloc_trim` after model build and after the GL self-test (no-op off Linux), so compile/context scratch stops masquerading as resident memory.
3. **Content-keyed compiled-model cache** — the compiled world is saved as MuJoCo binary (`sim/assets/.model_cache/world-<key>.mjb`, ~121 MB) keyed by the generated MJCF + referenced asset mtimes/sizes + MuJoCo version. Warm boots load it in **~50 ms** instead of minutes of qhull on weak machines, and skip the compile debris entirely. Corrupt/stale cache falls back to compiling and is replaced; superseded entries are pruned.

## Measured

Full init including both renderers, render scale 2 (the weak-machine config):

| | before | after (cold) | after (warm) |
|---|---|---|---|
| Linux/llvmpipe (Lima VM) | 2495 MB | 1204 MB | **785 MB** |
| macOS (native GL) | 2023 MB | 798 MB | **669 MB** |
| model init time (warm) | minutes on a Pi | | **0.1 s** |

Identical frames (std 61.6 before/after), identical physics (`cmd_vel` drive test 0.60 m both). `sim/sandbox/test_driver_core.py` PASSes.

On the 4 GB Pi this takes the stack's total demand from ~4.6 GB (≈1 GB permanently in swap → physics thrash) to ~3.1 GB — resident, with page-cache headroom.

## Notes for reviewers

- Sim-only: everything lives under `mars_sim_driver`'s world/core/world_server, which never runs on robots.
- The browser viewer is untouched — it renders its own GLB bundle; the cap only affects the robot's simulated camera frames.
- The cache sits inside the gitignored assets dir, so asset-bundle refreshes naturally clear it.


DavidDobas:
Folded into #522 as its own commit (c97a7ac9) — same effort, one PR.

PR #533, opened 2026-07-14
fix(brain): report 'failed' to cloud when a primitive never starts (INN-711)

DavidDobas:
Fixes [INN-711](https://linear.app/innate/issue/INN-711/brain-client-reports-primitive-running-to-cloud-even-when-the-goal-was).

## Problem

`PrimitiveRunner.start_task()` sent the `running` lifecycle message and set `state.primitive_running` unconditionally — even when `_send_goal()` silently failed because the `execute_skill` action server didn't respond within its 1s `wait_for_server` window (typical right after a directive switch / skill reload).

Since no goal existed, no terminal result ever arrived, so the cloud agent believed the primitive was running forever and refused to re-trigger it ("should not stop running primitives").

## Fix

- `running` is announced only after the goal is actually dispatched (`_send_goal()` now returns whether it dispatched).
- Every never-started path now sends a terminal `failed` lifecycle message to the cloud via a new `report_start_failure()` helper, which also publishes the local app status and fires `on_task_finished` (gaze resume + deferred re-registration drain):
  - action server unavailable (`wait_for_server` timeout)
  - unknown skill ID from the cloud (`vision_output._start_next_task`)
  - unknown pending task after a stop+next_task sequence (`_maybe_run_pending`)
- A goal **rejected** by the action server now also sends the cloud a terminal `failed` — previously only the local app was told, leaving the cloud waiting on the `running` it had received at dispatch.

## Tests

New `test/test_runner_start_failure.py` (4 tests) pins: no `running` + terminal `failed` when the server is unavailable; `running` + state set when dispatched; terminal `failed` on goal rejection; unknown-skill reporting. Ran in the innate-dev container alongside the existing suite: 22 passed.

> [!NOTE]
> Replaces #530 — the branch was renamed to follow the `type/short-description` convention, which closed the original. All review discussion and the resolved+replied Greptile threads live there.

## Root cause fix (added after review discussion)

The reporting fix above is the safety net; the reason dispatch failed in the first place is that **registration never waited for the `execute_skill` server**. `skills_action_server` is a separate process, and `SkillCatalog.register()` told the cloud "you may trigger these skills" without confirming the server was discoverable — so a trigger fired immediately after a directive switch/activation raced ROS graph discovery against a single 1s `wait_for_server` poll.

`register()` now simply **waits for the server to be reachable before it registers**: if `server_is_ready()` is false it arms a short retry timer and returns; once ready it registers. No fail-open timer or fixed timeout — the wait always converges, because `skills_action_server` creates its action server *before* it publishes the `/brain/available_skills` roster, so any process healthy enough to trigger registration already has the server; only cross-process discovery latency (incl. the 2s respawn window) remains, which the retry resolves.

---

**Scope note (7fb8c1c2):** an earlier revision of this PR also hardened `_on_goal_response` against a deactivation-vs-goal-response race (surfaced in review because the safety-net code touched that method). On reflection that's a pre-existing concurrency issue unrelated to INN-711, so it's been reverted to keep this PR to its actual scope — the root-cause fix plus the minimal never-started safety net. If that race is worth fixing it'll get its own ticket.


ros2_ws/src/brain/brain_client/brain_client/skills/runner.py
```diff
@@ -160,21 +210,36 @@ def _handle_substep(self, substep: dict) -> None:
...
+                skill_id=skill_id,
+                reason="Goal rejected by action server",
+            )
+            # Only touch shared state if it is still this task's: a new task may
+            # already be running after the race that cleared ours.
+            running = self._state.primitive_running
+            if running is None or (running.get("primitive_id"), running.get("skill_id")) == (primitive_id, skill_id):
+                self._state.primitive_running = None
+                self._goal_handle = None
+                self._on_task_finished()
```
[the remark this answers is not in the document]
DavidDobas: Agreed — `running is None` at rejection time only happens after a deactivate/unregister race already tore the task down, and firing `on_task_finished` then would resume gaze on a deactivated brain. Fixed in 2ab17cdc with your suggested condition (match required, None no longer finishes); the terminal `failed` to the cloud is unchanged. `test_rejected_goal_reports_failed_even_after_state_cleared` now also pins `finished == []`.

ros2_ws/src/brain/brain_client/brain_client/skills/runner.py
```diff
@@ -160,21 +210,42 @@ def _handle_substep(self, substep: dict) -> None:
         if event == "completed" and output and output.strip():
```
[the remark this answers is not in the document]
DavidDobas: Right — the accepted branch had the same race, and worse: the orphaned goal kept executing untracked. Fixed in 1994080a: the staleness check is hoisted above both branches; a stale accepted goal is cancelled (fire-and-forget, mirroring `interrupt_for_deactivation`), the cloud gets the `interrupted` the teardown couldn't send (it had no handle yet), and `_goal_handle`/`_on_result` are never attached. Pinned by `test_accepted_goal_after_teardown_is_cancelled_not_adopted` plus a normal-path adoption test; suite at 31.

ros2_ws/src/brain/brain_client/test/test_registration_gate.py
```diff
@@ -0,0 +1,111 @@
...
+
+Registration is the cloud's licence to trigger skills, so it must not be sent
+before the execute_skill action server is discoverable — otherwise the very
+first trigger races graph discovery and fails. The gate defers registration on
+a retry timer and fails open (loudly) after a window so a wedged skills server
+degrades to per-trigger "failed" reports instead of a brain that never
+registers.
+"""
+
+import time
```
theo-michel: I think we should remove both tests because this is a temporary fix. We'll get the brain in a state where we don't have to do these kinds of failure recoveries and maybe simplify the process. I don't know if it's worth testing and keeping this type of functionning for a long time.

ros2_ws/src/brain/brain_client/brain_client/skills/registration.py
```diff
@@ -101,9 +118,46 @@ def _warn_dup(name, existing_id, new_id):
...
+            self._logger.info("execute_skill server not ready; deferring registration until it is.")
+        elif now - self._gate_started > REGISTRATION_GATE_FAIL_OPEN_SEC:
+            self._logger.error(
+                f"execute_skill server still not ready after {REGISTRATION_GATE_FAIL_OPEN_SEC:.0f}s — "
+                "registering anyway (skill triggers will fail until it comes up)."
+            )
+            self._clear_gate()
+            return True
+        if self._gate_timer is None:
+            self._gate_timer = self._node.create_timer(REGISTRATION_GATE_RETRY_SEC, self.register)
```
theo-michel: Not that nice looking

theo-michel (approved):
Rmv the tests as it's temporary fix until better brain and cloud system, and the code is bit hard to read in some sections, otherwise lgtm

ros2_ws/src/brain/brain_client/brain_client/skills/registration.py
DavidDobas: Cleaned up in dc81b9e7. Dropped the fail-open state machine (monotonic clock + 30s window + "register anyway") entirely — `register()` now just waits for the server: not ready → arm a short retry timer and return; ready → register. It reads top-to-bottom now (`_server_ready` / `_wait_for_server` / `_stop_waiting_for_server`). The fail-open turned out to guard a case that can't happen: `skills_action_server` creates the action server before it publishes the `/brain/available_skills` roster, so anything healthy enough to trigger registration already has the server — the retry always converges.

ros2_ws/src/brain/brain_client/test/test_registration_gate.py
DavidDobas: Done in dc81b9e7 — removed `test_registration_gate.py`. Agreed it's temporary mitigation until the brain/cloud handshake is reworked, not worth pinning. Kept `test_runner_start_failure.py` since that covers the substantive fix (never report a phantom "running"; emit a terminal "failed" so the cloud can recover), which stands regardless of how the handshake evolves — shout if you'd rather drop those too.

ros2_ws/src/brain/brain_client/brain_client/skills/runner.py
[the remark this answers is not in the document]
DavidDobas: Update: reverted this in 7fb8c1c2. Stepping back, this accepted/rejected teardown-race handling is a pre-existing concurrency issue (deactivation racing a goal response) that's independent of INN-711 — it just surfaced here because the safety-net code touched `_on_goal_response`. Keeping this PR to its actual scope; `_on_goal_response` is back to main's shape. If the race is worth fixing it'll get its own ticket.

ros2_ws/src/brain/brain_client/brain_client/skills/runner.py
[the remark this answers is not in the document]
DavidDobas: Update: reverted in 7fb8c1c2 along with the rest of the goal-response race handling — see the note on the accepted-goal thread. Out of INN-711 scope; separate ticket if warranted.

ros2_ws/src/brain/brain_client/test/test_registration_gate.py
DavidDobas: Dropped completely, rather solved root cause

PR #536, opened 2026-07-14
fix(cam): publish depth as 16UC1 (unsigned millimeters)

DavidDobas:
Depth in millimeters is never negative, and the signed 16SC1 encoding is
nonstandard enough that stock viewers refuse it -- Foxglove's image panel
errors out on the sim's (and robot's) depth topic. 16UC1 is the ROS
convention for millimeter depth.

Changed on both sides of the digital twin so the wire contract stays
identical: the robot's disparity->depth publisher (mars_cam publishing.cpp)
and the sim driver (mars_sim_driver node.py), plus the comments/docstrings
that named the old encoding. The clamp ceiling rises from 32767 to 65535,
which is moot below the 10m MAX_DEPTH cutoff.

Not changed: filterSpeckles' internal CV_16SC1 buffer -- OpenCV's API
requires that exact type and it never reaches the wire.

Consumers audited: the webapp calibration indicator only checks message
arrival; the controller app's DepthCheckScreen already decodes both 16U and
16S; nav consumes the points topic; /mars/main_camera/remote/depth is a
passthrough throttle.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

## Validation
- Sim-side conversion exercised against a live render: uint16, sane mm range, invalid=0 preserved.
- Foxglove image panel on the sim depth topic is the acceptance test for the sim half; a robot-side depth check (controller app DepthCheckScreen) covers the hardware half.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_cam/config/stereo_depth_estimator.yaml
+++ b/ros2_ws/src/mars_bot/mars_cam/config/stereo_depth_estimator.yaml
@@ -59,7 +59,7 @@ arm_camera_driver:
     # ── Topics ────────────────────────────────────────────────────────────
     left_topic: "/mars/main_camera/left/image_raw"
     right_topic: "/mars/main_camera/right/image_raw"
-    depth_topic: "/mars/main_camera/depth/image_rect_raw"     # 16SC1 mm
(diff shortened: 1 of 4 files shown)


karmanyaahm (approved):
lgtm

PR #540, opened 2026-07-15
fix(sim): make the prebuilt image tag reproducible (hash git-tracked files)

DavidDobas:
## Problem

`./innate-sim up` never finds a matching prebuilt Innate OS image on macOS and rebuilds locally every time, even for a commit CI *did* build and publish.

The launcher pulls a content-addressed tag `inputs-<hash>`, where the hash covers the sim build inputs. Both the launcher ([sim/launcher/config.py](../blob/main/sim/launcher/config.py)) and CI ([ci/compute_sim_image_inputs_hash.py](../blob/main/ci/compute_sim_image_inputs_hash.py)) enumerated `ros2_ws/src` with `rglob("*")` — a **filesystem walk**, so it also hashes untracked/gitignored files. On a Mac, the stray `.DS_Store` files under `ros2_ws/src` (and gitignored dirs like `live_agent/`) get hashed; CI's clean Linux checkout has none. Same commit → different hash → the launcher looks for a tag CI never published.

Concretely, for commit `4966356`:
- local hash (with `.DS_Store`): `d7acecfdd9b7…` → **not in registry**
- CI hash / tracked-only: `59cda85e7477…` → **published, exists**

## Fix

Enumerate **git-tracked** files under `ros2_ws/src` (`git ls-files`) instead of walking the filesystem, in both hash implementations (they must stay in sync). CI publishes `inputs-<hash>` from a clean checkout, so the tracked set is exactly what it built from — this makes the hash reproducible across machines.

Notes:
- **CI's hash is unchanged** (its checkout has no untracked files), so already-published images and `cleanup-sim-images.yml` pinning stay valid.
- Not `.dockerignore`-based: the ros image's dockerignore re-includes all of `ros2_ws/src/**`, so it would *still* hash `.DS_Store` and wouldn't fix this.
- The helper errors loudly if `git ls-files` fails rather than silently falling back to `rglob`, which would reintroduce the non-reproducibility.

## Verification

- Both implementations produce an identical hash on `main`'s tree.
- Planting `.DS_Store` files under `ros2_ws/src` no longer changes the hash.
- Tracked-only hashing reproduces CI's published `59cda85e7477` for `4966356` exactly.

### DIFF
--- a/ci/compute_sim_image_inputs_hash.py
+++ b/ci/compute_sim_image_inputs_hash.py
@@ -14,6 +14,7 @@
 # why this lives in one file. sim/launcher/config.py computes the same hash
 # independently; keep them in sync.
 import hashlib
+import subprocess
 from pathlib import Path
 
 root = Path(".")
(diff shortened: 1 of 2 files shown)


karmanyaahm (approved):
sure sg

PR #543, opened 2026-07-15
fix: eliminate boot/runtime errors from robot logs (webapp Debugging page)

theo-michel:
## Goal

No ERROR lines in the robot logs (as streamed to the webapp Debugging page) during normal functioning. Every fix targets the root cause; nav was deliberately left untouched.

**Verified live on R7-27**: restarted the affected panes (arm-recorder, ik-logger, cam-leader, brain-nav.0) and re-captured logs — **0 ERROR lines** post-restart in all of them, with warnings reduced to compact, actionable one-liners.

### fix(arm): remove unused MoveIt move_group stack and dead boot homing
- Two `move_group` nodes launched at every boot (`arm.launch.py` include + `ik.launch.py`) with **zero consumers** — live check showed `/move_action` with 0 clients. IK is served by the KDL `ik.py` node (reads the URDF directly); trajectory execution is the driver's own spline path. Each instance errored at boot (`No 3D sensor plugin(s) defined for octomap updates`, `No controller_names specified.`) plus ~6 warnings, and burned Jetson CPU/RAM. Deleted `planning.launch.py`, `moveit.yaml`, `kinematics.yaml` (referenced nowhere else).
- The constructor-time "go home" has **never worked**: `latest_joint_positions_` is only filled by the control timer and trajectory waypoints are only consumed by it, and neither runs before `executor.spin()`. It logged `No current joint state available` + `Failed to reach home position` on every boot since inception. Removed; boot behavior is physically unchanged and homing stays available via the arm services.

### fix(cam): log per-peer transport teardown at WARN, not ERROR
A peer vanishing mid-stream (tab close, wifi drop, failed DTLS handshake) posts a GStreamer bus error that `poll_pipeline_health()` already handles by releasing the peer — but it was logged at ERROR with full GStreamer internals, 24+ times per session while the marsnav laptop agent's retry loop was failing. Now WARN; shared encode/audio pipeline errors (real faults) stay ERROR.

### fix(brain): clock-safe skills wait, scanner and noise fixes
- `_startup`'s available-skills wait used wall-clock `time.time()`: an **NTP clock step mid-boot** (~150s jump observed) expired the 25s deadline instantly, so 29 agents were validated against an empty catalog (29 bogus warnings, each dumping the full catalog) and registration fired before the websocket connected. Now `time.monotonic()` with a 60s budget matching the skills server's real first-load time on the Jetson.
- Per-agent skill validation is skipped while the catalog is still empty (self-heals when the roster lands), and missing-skill warnings list only the dangling ids.
- Pre-connect websocket sends log at INFO — the registration payload is replayed on connect (`orchestrator.on_ws_status`).
- The physical-skill scanner no longer descends into `__pycache__`/hidden dirs; a stray empty `metadata.json` in `custom_skills/__pycache__/` errored every boot (junk file also deleted on the robot).
- `DetectOpponentMove`: absent/incomplete board calibration at discovery time is now a warning naming the missing corners; `execute()` still errors when the pose is actually needed.

### chore(webapp): green strict typecheck; drop dead sim-session thumbnail path
`tsc -p webapp/tsconfig.json` (strict + checkJs): **61 errors → 0**. Annotation-only fixes (missing `typeof` in `ReturnType` imports, `@param` gaps, narrowed-value captures) plus one real cleanup: `cameraSwitch`'s `session.thumbnailCanvas` branch referenced the sim session class deleted in the sim refactor — no session exposes it anymore. Webapp unit tests pass. Also gitignores `workspace/skill_storage/` (runtime per-skill KV stores).

## Known remaining noise (out of scope)
- ~20 warnings from stale test agents in `~/agents` / `workspace/custom_agents` referencing deleted local skills — user data on the robot, delete the agent files to silence.
- Nav2 costmap out-of-bounds spam when driven off-map (nav config, intentionally not touched).
- marsnav laptop agent (192.168.0.152) connection-refused warnings — external client.


DavidDobas (approved):
LGTM

PR #546, opened 2026-07-16
feat(logger): include robot_id and MAC address in telemetry vitals

theo-michel:
## What

The telemetry vitals payload now carries robot identity — `robot_id` and `mac_address`:

```json
{"robot_id": "R7-44", "mac_address": "E4:5F:01:2B:3C:4D", "commit": "0f66c925", "cpu_usage": 12.0, ...}
```

## Why

Vitals rows in `robot_logs.vitals` carried no robot identity — attribution was only possible via the auth token.

## How

**`robot_id`** — read once at startup from `data/robot_info.json`, resolved via the same `INNATE_OS_ROOT` → `~/innate-os/data/` convention `mars_bt_provisioner` uses. Read from the file rather than hardcoded, so each robot reports its own. Missing file/key → `"unknown"` plus a startup warning, so unprovisioned robots still boot.

**`mac_address`** — subscribed from `/robot/info`, which `mars_control` publishes at 1 Hz as JSON; the MAC is its `device_id` field (the Bluetooth adapter address from `bluetoothctl show`, see `app.cpp:get_bluetooth_device_id`). Cached on receipt, following the existing cached-latest pattern used for battery and diagnostics.

Two edge cases in `app.cpp` shaped the handling — `publish_robot_info_callback` omits `device_id` when no Bluetooth adapter is present, and publishes a bare `"{}"` if it hits an exception. So the callback keeps the last-seen MAC rather than clearing it, and vitals omit the key entirely until a MAC is observed (rather than sending null every 5s in sim, where `mars_control` isn't running at all).

## Verification

- `ruff check` / `ruff format --check` clean.
- `robot_id` helper exercised against a fixture `robot_info.json`: returns `R7-44` with the file present, `unknown` with `INNATE_OS_ROOT` pointed at a nonexistent path.
- `_on_robot_info` / `_log_vitals` driven directly (real module, ROS deps stubbed) across six cases using payload shapes taken from `publish_robot_info_callback`: `device_id` present → cached; absent → retained; malformed JSON → warned + retained, not clobbered; bare `"{}"` → retained; never-seen → key omitted from vitals; present → both ids in payload.

## Reviewer notes

- **`robot_id` key name is now confirmed** — `app.cpp:750` reads `robot_info["robot_id"]` from the same file, which resolves the open question from the first revision of this PR. It has not been read off a physical robot, but the C++ side agrees on the key.
- **Field naming**: the topic calls the MAC `device_id`; I send it as `mac_address`, since `device_id` alongside `robot_id` in a telemetry table reads ambiguously. Happy to rename for consistency with the topic if preferred.
- **Cloud side may need matching columns** in `robot_logs.vitals` if the logger service doesn't pass unknown fields through.

### DIFF
--- a/ros2_ws/src/cloud/innate_logger/innate_logger/logger_node.py
+++ b/ros2_ws/src/cloud/innate_logger/innate_logger/logger_node.py
@@ -72,11 +72,8 @@ class LoggerNode(Node):
         auth.wait_for_token(on_retry=self._log_auth_retry)
         self._client = TelemetryClient(url=telemetry_url, auth=auth)
 
-        # Git commit and robot identity at startup
+        # Git commit at startup
         self._git_commit: str = self._get_git_commit()
-        self._robot_id: str = self._get_robot_id()
-        if self._robot_id == "unknown":
-            self.get_logger().warning("robot_id not found in robot_info.json — reporting 'unknown'")
 
         # Initialise CPU baseline (first call always returns 0.0%)
         psutil.cpu_percent()
@@ -84,6 +81,7 @@ class LoggerNode(Node):
         # ── Subscriptions ───────────────────────────────────────────
         self._latest_battery: BatteryState | None = None
         self._latest_diagnostics: DiagnosticArray | None = None
+        self._robot_id: str | None = None
         self._mac_address: str | None = None
 
         self.create_subscription(BatteryState, "/battery_state", self._on_battery, 1)
@@ -98,23 +96,6 @@ class LoggerNode(Node):
 
     def _log_auth_retry(self, attempt: int, error: AuthError, next_delay: float) -> None:
         self.get_logger().warning(f"Auth not ready yet (attempt {attempt}): {error} — retrying in {next_delay:.0f}s")
-
-    @staticmethod
-    def _get_robot_id() -> str:
-        """Read robot_id from robot_info.json.
-
-        mars_control seeds the file with `"robot_id": null` (app.cpp
-        get_robot_info), so an unprovisioned robot has the key present but
-        empty — test the value, not just the key.
-        """
-        mars_root = os.environ.get("INNATE_OS_ROOT", os.path.join(os.path.expanduser("~"), "innate-os"))
-        robot_info_path = os.path.join(mars_root, "data", "robot_info.json")
-        try:
-            with open(robot_info_path) as f:
-                robot_id = json.load(f).get("robot_id")
-        except Exception:
-            return "unknown"
-        return str(robot_id) if robot_id else "unknown"
 
     @staticmethod
     def _get_git_commit() -> str:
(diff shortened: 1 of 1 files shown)


theo-michel:
Good catch — confirmed and fixed in 497074c2.

Verified the claim rather than assuming: `app.cpp:535` does seed `{"robot_id", nullptr}`, and `get_robot_info()` both creates the file with those defaults and ensures the default keys exist, so a freshly-flashed robot really does have `"robot_id": null` on disk. Reproduced the old behavior directly:

```python
>>> str(json.loads('{"robot_id": null}').get("robot_id", "unknown"))
'None'
```

So every unprovisioned robot would have reported `robot_id = "None"` with the warning silent — exactly as described.

Fixed by testing the value rather than leaning on the key default:

```python
with open(robot_info_path) as f:
    robot_id = json.load(f).get("robot_id")
except Exception:
    return "unknown"
return str(robot_id) if robot_id else "unknown"
```

I used a truthiness check instead of `is not None` — it covers the null default and an empty-string id in the same branch, and matches the `if device_id:` check already used on the `/robot/info` path. A falsy-but-valid id isn't reachable here (ids are strings like `R7-44`; an integer id still works, covered below).

My original test fixtures only covered a populated id and a missing file, which is why this slipped through. Extended to drive the real `_get_robot_id` against on-disk fixtures, including the exact default `mars_control` writes:

| case | result |
|---|---|
| provisioned `"R7-44"` | `R7-44` |
| **mars_control default (`null`)** | `unknown` |
| key absent | `unknown` |
| empty string | `unknown` |
| file missing | `unknown` |
| malformed JSON | `unknown` |
| non-string id `4471` | `4471` |

The `/robot/info` MAC path was already safe from this class of bug — it does `.get("device_id")` followed by a truthiness check — and its suite still passes.

ros2_ws/src/cloud/innate_logger/innate_logger/logger_node.py
```diff
@@ -93,6 +99,23 @@ def __init__(self) -> None:
...
+        """
+        mars_root = os.environ.get("INNATE_OS_ROOT", os.path.join(os.path.expanduser("~"), "innate-os"))
+        robot_info_path = os.path.join(mars_root, "data", "robot_info.json")
+        try:
+            with open(robot_info_path) as f:
+                robot_id = json.load(f).get("robot_id")
+        except Exception:
+            return "unknown"
+        return str(robot_id) if robot_id else "unknown"
+
```
karmanyaahm: get robot id from robot info msg as well no? just like device id

karmanyaahm (dismissed):

ros2_ws/src/cloud/innate_logger/innate_logger/logger_node.py
theo-michel: Shit told it to do so, bad bot

theo-michel:
Agreed — done in 1b015288. Both fields now come from `/robot/info` and `_get_robot_id` is gone, along with the `INNATE_OS_ROOT` path handling and file IO.

I checked the one thing that could have argued against it — whether the logger ever runs somewhere `mars_control` doesn't — and it doesn't: `logger_node` is launched only from `launch_ros_in_tmux.sh` (no match in `launch_sim_in_tmux.zsh`), and that same script launches `mars_control app.launch.py` in its first window. So the topic is always up wherever the node runs, and at 1 Hz it's cached well before the first 5s vitals tick.

It's also strictly better than what I had: the topic re-reads `robot_info.json` every second, so a robot **provisioned while the node is running** now gets picked up — the startup file read would have reported it unprovisioned until the next restart.

```python
robot_id = info.get("robot_id")
if robot_id:
    self._robot_id = str(robot_id)

device_id = info.get("device_id")
if device_id:
    self._mac_address = str(device_id)
```

Worth noting the null-`robot_id` bug greptile caught above is still live on this path — `app.cpp:753` explicitly publishes `robot_id: null` when unprovisioned — so the truthiness check is load-bearing, not stylistic. Both fields are now omitted from vitals until seen rather than sent as `"None"`. Kept an explicit regression test on it:

| case | result |
|---|---|
| both fields present | cached |
| **`robot_id: null` (app.cpp:753)** | **omitted, not `"None"`** |
| `device_id` absent (no BT adapter) | previous value retained |
| malformed / bare `{}` | previous value retained |
| never seen | both keys omitted |
| provisioned mid-run | picked up |

One consequence for the cloud side: `robot_id` can now be **absent** from a vitals row (unprovisioned robot, or the first tick after a `mars_control` restart) rather than always present as `"unknown"`. If `robot_logs.vitals` would rather have an explicit sentinel than a missing column, say so and I'll send `"unknown"` instead.

karmanyaahm (approved):
lgtm

PR #550, opened 2026-07-17
feat(pick): two-camera grasp verify + skill_lib vision/gemini + cancellation refactor

theo-michel:
Stacks on `theo/pick-wrist-servo` (where #548/#549 already merged). Consolidates the remaining pick work plus adjacent cleanups.

## pick_any_object
- **Two-camera verify**: `_grasp_verified` sends the head **and** wrist frames to Gemini in one call. The head view answers "is it still on the floor?"; the wrist view (flagged as mirrored in the prompt) can show the object held in the fingers, so a dangling sock isn't misread as a drop. Degrades to head-only if the wrist frame is missing. Verdict logic and the `verify` telemetry fields are unchanged (added a `cams` count).
- **Cancellation refactor**: `SkillCancelled` + a single `_checkpoint()` at loop tops replaces the `if self._cancelled: return None` plumbing. Stage return types are now honest (`None` = not found, not "cancelled"), and cancel-during-search/position is reported as `CANCELLED` instead of `FAILURE`. No checkpoint inside close/twist/lift — once the fingers commit, the grasp finishes.

## skill_lib
- Extracted the stateless vision math (frame decode, Gemini parse, LK tracking, color segmentation) into `skill_lib/vision.py` and the proxy vision call into `skill_lib/gemini.py`.
- `gemini.ask_image` accepts one image or a list (enables the two-camera verify).

## Folded in per "everything"
- **pyright**: return-type annotations on the `Skill` base class (`types.py`) + `pyrightconfig.json`; in-progress pyright cleanups across other `innate_skills/*` (from the parallel remediation task).
- **post_update.sh**: enforce `root:<user> 640` on `/etc/innate.env` so the non-root launch readers can load the service key (a hand-created 600 root:root file silently dropped the key).
- **webapp**: pick tuning panel renamed `pickTunePanel.js` → `pickOverlay.js` (+ `main.js`/`app.css`).

## Verification
Compiles + pyright-clean on the pick/skill_lib files; skills server hot-reloads to `ready`; request-shape test confirms head-then-wrist image ordering. Not yet exercised on a live robot — the robot's Gemini calls 403 until the cloud proxy PR (innate-cloud #80) deploys with the gemini service.

## Note
#548 and #549 are already **merged** into this base branch, so there's nothing to close on them.


theo-michel:
Folded directly into #542 — the commit is now on `theo/pick-wrist-servo` (cherry-picked as e89b0723), so #542 → main carries everything in one PR. No longer need the stacked branch.

PR #556, opened 2026-07-21
Favorite skills: robot-persisted per-user curation of the Skills menu

theo-michel:
## What

Users **star** skills; the Skills menu shows favorites by default (or the whole roster until the first star), with the rest behind **"Show all (N more)"**. Favorites live **on the robot** — not in any client — so the webapp and the mobile app share one list:

- `/brain/favorite_skills` — latched + heartbeated broadcast, String JSON `{"skills": [ids]}`
- `/brain/set_favorite_skills` — full-list replace from any client (star toggle publishes the whole updated list; the broadcast echo confirms)
- Persisted in `workspace/.preferences/favorite_skills.json` (gitignored), atomic writes, survives restarts and corrupt files

Also fixes row labels showing roster paths: `innate-os/pick_socks` now renders as **"Pick Socks"** (source prefixes are plumbing, not UI).

## History / design

The first commit curated via a hardcoded list in webapp JS; that was the wrong layer — curation must be robot-side so all clients agree — and shipped defaults shouldn't override user judgment. This supersedes it with user-owned favorites. The principle stands: exposure is a consumer/user list, never author-side metadata on skills (RFC included: `docs/SKILLS_AS_FUNCTIONS_RFC.md`, draft for discussion, not implemented here).

**Assumption:** clients are unauthenticated, so "per user" = per robot (one favorites list per machine). If we later grow user identities, the store keys naturally by user file.

**Mobile app:** adopts the same two topics via rosbridge — no innate-os changes needed when it does.

## UX details

- First star ever auto-flips to "Show all" so the list doesn't collapse to one row mid-browse
- Star sits beside the run button (never nested — a mis-tap must not launch the skill)
- Re-renders are membership-gated (latched heartbeats don't steal form focus)
- A favorite whose skill is mid-reload survives (store keeps unknown ids; the view just doesn't show them)

## Testing

- `pytest ros2_ws/src/brain/brain_client/test/test_favorite_skills.py` — 7 tests (sanitize, persistence round-trip, corrupt-file recovery, legacy bare-array format); added to the CI fast bucket in `ci/run_integration_tests.sh`
- `node webapp/tests/skillsMenu.test.js` — 6 tests on the pure helpers (favorites filter, first-run behavior, path-hiding labels)
- Browser end-to-end against the real `skillsMenu.js` with a stubbed rosClient: first-run shows all with outline stars → first star keeps the list expanded and publishes `{"skills":[...]}` → "Show fewer" shows exactly the starred rows → a robot-side list change (as the mobile app would make) updates the view → unstarring all returns to first-run
- `ruff` + `pre-commit` (CI config) pass on all touched Python


theo-michel:
Superseded — the favorites work (and the skill_lib cleanup) moved into #542 per review, so the whole skills story lands as one PR. Branch left in place.

PR #560, opened 2026-07-21
chore(setup): default to gemini cloud agent + better api key entering interface

karmanyaahm:
Improves the brain/LLM step of `innate-sim setup` (all in `sim/launcher/setup_wizard.py`): a fresh setup now defaults to running *with* an agent, key entry gives visible feedback, and re-running setup can rotate a key.

## 1. Default to a local cloud-agent when no keys are set

When neither `GEMINI_API_KEY` nor `INNATE_SERVICE_KEY` was detected, the backend prompt defaulted to **3 (None / no agent)**, so Enter-through-the-wizard landed on a sim with no brain. The no-keys default is now **1 (your own Gemini key)**. Defaults with a key present are unchanged: `has_gemini → 1`, `has_service_key → 2`.

## 2. Masked key entry with visible feedback

Pasting a key used to be fully hidden (`getpass`), so there was no way to tell whether the paste registered — or registered twice. Key entry now:

- echoes `*` per character with a live length count, so a double-paste is obvious
- supports **Ctrl-U** (clear and retry) and backspace
- ignores spaces and control characters (keys never contain them)
- handles Ctrl-C explicitly (raw mode swallows the signal)

When stdin/stdout is not an interactive TTY or `termios` is unavailable (piped input, Windows), it falls back to the old fully-hidden `getpass` behavior rather than echoing to a non-terminal.

## 3. Re-runs can replace an already-set key

A configured key used to short-circuit with "already configured", so rotating a key meant editing `.env` by hand. Setup now asks `<KEY> is already set. Replace it?` (default **No** — pressing Enter keeps the current key). The restore-from-commented-`.env` and found-in-shell paths still apply only when the key is not already configured.

## Wording

The section is retitled **Brain Backend → Cloud LLM Access** and the options are rephrased in plain terms ("Your own Gemini key", "Innate service key (from your robot)", "None") — no more "local brain vs. hosted brain" jargon on first contact.

## Notes

- `ruff check` and `ruff format --check` pass on the touched file.
- The masked reader restores terminal attributes in a `finally`, so Ctrl-C/EOF mid-entry leaves the terminal sane.

### DIFF
--- a/sim/launcher/setup_wizard.py
+++ b/sim/launcher/setup_wizard.py
@@ -70,11 +70,63 @@ def _prompt_yes_no(question: str, *, default: bool = False) -> bool:
 
 
 def _prompt_secret(question: str) -> str:
+    prompt = f"{YELLOW}{question}: {NC}"
     try:
-        return getpass.getpass(f"{YELLOW}{question}: {NC}", stream=sys.stdout).strip()
+        masked = _read_masked_secret(prompt)
+        # None: no interactive TTY (or raw mode unavailable) -- fall back to
+        # fully hidden input rather than echoing to a non-terminal.
+        if masked is None:
+            return getpass.getpass(prompt, stream=sys.stdout).strip()
+        return masked
     except (KeyboardInterrupt, EOFError):
         print()
         raise SystemExit(1)  # noqa: B904
+
+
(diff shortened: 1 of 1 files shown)


DavidDobas (approved):
LGTM

PR #564, opened 2026-07-22
Show launch-file labels for sim logs, matching the robot

DavidDobas:
## Problem

In the sim, the Logging page labels each source by its raw tmux `window.pane` id (`nav-brain.1`, `arm-ik.0`, `foxglove.0`), while on the robot the same page shows launch-file names (`mode_manager.launch.py`, …). Same webapp, different output.

## Root cause

The top-level label comes from a static lookup, `PANE_LAUNCH_LABELS` in [`webapp/js/constants.js`](../blob/main/webapp/js/constants.js), keyed by tmux window name → per-pane launch file, resolved in [`format.js`](../blob/main/webapp/js/logging/format.js):

```js
return (PANE_LAUNCH_LABELS[win] || [])[idx] || pane;  // ← falls back to raw "window.idx"
```

The map only contained the **robot's** windows (from `launch_ros_in_tmux.sh`). The sim's windows (from `launch_sim_in_tmux.zsh`) weren't there, so every sim pane fell through to the raw id.

## Fix

Add the sim's windows to the map with their real pane→launch-file layout. Purely additive data through the exact render path the robot already uses — no rename, no logic change, and nothing that touches `runtime.py`'s hardcoded `nav-brain.1` capture.

Note the sim's layout genuinely differs from the robot's (different window split, `.sim.launch.py` variants, `mode_manager` in pane 0 of `nav-brain`), which is exactly why a plain window-rename would have mislabeled panes — hence encoding the sim's actual layout instead.

## Test
- `node --check` passes on constants.js.
- Manual: bring up the sim, open the Logging page — sources now read as launch files.

### DIFF
--- a/webapp/js/constants.js
+++ b/webapp/js/constants.js
@@ -263,10 +263,13 @@ export const GET_RUN_LOGS_SERVICE = "/innate_training/get_run_logs";
 export const BRAIN_RELOAD_SERVICE = "/brain/reload";
 
 // Friendly subpane names per tmux window — index = pane (0 = left, 1 = right).
-// Mirrors innate-os/scripts/launch_ros_in_tmux.sh ROS_COMMAND_GROUPS; keep the
-// two in sync. Labels the Logging page's subpane picker by launch file.
+// Robot windows mirror innate-os/scripts/launch_ros_in_tmux.sh ROS_COMMAND_GROUPS;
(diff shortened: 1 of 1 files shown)


DavidDobas:
Folded into #563 — consolidating the innate-os changes into a single PR.

PR #568, opened 2026-07-22
feat(teleop): record camera video to MP4 from the webapp

theo-michel:
Restores work that had been parked in a stash (`wip: video recorder + record button`) and rebuilds it on current `main` — it applied cleanly, no conflicts.

## What

A record toggle in the teleop right rail brackets an on-robot MP4 capture of the main and arm cameras, so you can grab clips of the robot driving around without screen-recording the browser.

- **`video_recorder.py`** (`mars_bringup`, launched by `bringup_core`) — `/video_recorder/start` and `/video_recorder/stop` (`std_srvs/Trigger`), plus a latched `/video_recorder/status` `Bool` with a 1 Hz heartbeat so a reloaded page (or a node restart mid-recording) converges on what the robot is actually doing.
- **`recordButton.js`** — right-rail toggle that follows the latched status. Hidden under `simControls`, since the sim doesn't run robot bringup.

Frames are written on a fixed-rate 30 Hz timer — latest frame per tick, duplicated on stalls — so the file plays back in real time regardless of camera or subscriber drops. If no frames ever arrive, stop reports failure and removes the empty directory instead of leaving a zero-byte MP4.

## Output

`<INNATE_OS_ROOT>/data/recordings/video_<timestamp>/{main,arm}.mp4`, `mp4v`, 30 fps.

The writer is sized from the first frame, so resolution is whatever the camera topics publish — today **640x480** for both (`/mars/main_camera/left/image_raw` is the left half of the 1280x480 stereo capture; `/mars/arm/image_raw` is 640x480 native). Nothing here caps it: raise the driver's publish size and the recordings follow.

## Testing

Not yet exercised on the robot since being restored — `python3 -m py_compile` passes and the webapp APIs it calls (`rosClient.subscribe` returning an unsubscribe, `callService(service)`) are unchanged on `main`. Wants a live pass: record while driving, confirm both MP4s play back at real-time length.

## Notes

- `bringup_core` gains an always-running node. It idles with no subscriptions until start is called, so the cost is one node, not one camera subscriber.
- The status subscription doesn't pass an explicit message `type`. That's fine while the node is in `bringup_core` (it publishes at construction), but if the node is ever made opt-in, rws will error on every retry until it comes up.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_bringup/CMakeLists.txt
+++ b/ros2_ws/src/mars_bot/mars_bringup/CMakeLists.txt
@@ -40,6 +40,7 @@ ament_python_install_package(${PROJECT_NAME}
 install(PROGRAMS
   mars_bringup/bringup.py
   mars_bringup/rosbag.py
+  mars_bringup/video_recorder.py
   DESTINATION lib/${PROJECT_NAME}
 )
 
--- a/ros2_ws/src/mars_bot/mars_bringup/launch/bringup_core.launch.py
+++ b/ros2_ws/src/mars_bot/mars_bringup/launch/bringup_core.launch.py
@@ -23,11 +23,20 @@ def generate_launch_description():
         output="screen",
     )
 
+    # On-demand camera→MP4 recorder driven by the webapp's record button.
+    video_recorder_node = Node(
+        package="mars_bringup",
+        executable="video_recorder.py",
+        name="video_recorder",
+        output="screen",
+    )
+
     # base_link -> base_footprint static TF is now published by
     # robot_state_publisher via the URDF (base_footprint_joint).
 
     return LaunchDescription(
         [
             bringup_node,
+            video_recorder_node,
         ]
     )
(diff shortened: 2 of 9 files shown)


theo-michel:
**Update: recordings are now 1080p, live-tested on R7-27.**

`0de46bfa` raises the main-camera capture to the sensor max (3840×1080 MJPG, decode/rotate on NVDEC/VIC) and taps the full-res frames via `/mars/main_camera/stereo`, subscriber-gated so the 12 MB frames only serialize out while a recording is running. All existing consumers are untouched — `left`/`right` stay 640×480 for policies, datasets, VP8 teleop.

Measured on the robot after a full camera-container + bringup restart:

| Check | Result |
|---|---|
| `main.mp4` | **1920×1080** (left eye cropped from stereo) |
| `arm.mp4` | 640×480 (unchanged; arm driver is YUYV-bound, out of scope) |
| `/mars/main_camera/left/image_raw` | 15.0 fps — no regression |
| `/mars/arm/image_raw` | 30.0 fps — no USB2 starvation at full capture width |
| Recorder CPU | ~70% of one core while recording, idle otherwise |
| Camera container | ~155% idle (includes the now-4K-wide BGRx→BGR videoconvert) |
| File size | ~30 MB for 8.5 s (~29 Mbps mp4v) |

Start/stop exercised end-to-end via the Trigger services; both files play back at real-time length. If the container CPU cost ever matters, dropping `width/publish_stereo_width` to 2560×720 in `stereo_depth_estimator.yaml` gives 1280×720 recordings and proportionally less convert work — one knob, no code change.

theo-michel:
Superseded by #569, which carries this same work plus the 1080p capture path as one reviewable unit. Closing so the recording pipeline lives in a single PR.

PR #572, opened 2026-07-23
fix(webapp): fetch /config.json once per load, not ~5 times

karmanyaahm:
## Problem

Six modules each `fetch("/config.json", { cache: "no-store" })` independently — `router.js`, `shell.js`, `robotSession.js`, and the `teleop`/`agent`/`collect` section entries. Since the router dynamically imports a section on top of the shell + session, a single page load hits the proxy **~5 times**. And `cache: "no-store"` bypasses the HTTP cache entirely, so each one is a **full download** — it doesn't even benefit from the proxy's cheap `stat()`→304.

## Fix

A shared `js/config.js` that memoises the promise:

```js
let _config;
export function getConfig() {
  return (_config ??= fetch("/config.json", { cache: "no-store" })
    .then((r) => (r.ok ? r.json() : {}))
    .catch(() => ({})));
}
```

All six sites now `await getConfig()` (or reuse its promise). One request per page load, still `no-store` so a fresh load reflects env-driven flags — just deduped in-page. Behaviour is otherwise identical (each site got `{}` on failure before and still does).

## Verified
`node --check` passes on all seven files; the only remaining `/config.json` fetch is in `config.js`.

### DIFF
--- a/webapp/js/agent/main.js
+++ b/webapp/js/agent/main.js
@@ -11,6 +11,7 @@
 
 import { ros } from "../rosClient.js";
 import { mountPage } from "../pageMount.js";
+import { getConfig } from "../config.js";
 import { robotSessionFactory } from "../robotSession.js";
 import { createVideoStage } from "../teleop/videoStage.js";
 import { createTelemetry } from "../teleop/telemetry.js";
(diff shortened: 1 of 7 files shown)


theo-michel (approved):
LGTM

PR #573, opened 2026-07-23
ci: stream Cloud Build logs into the integration-test job output

karmanyaahm:
## Why

The integration-test job submits the build to Cloud Build and then sits silent until it finishes — the build logs never show up in the Actions output, only a Cloud Logging URL.

Root cause: the build uses `logging: CLOUD_LOGGING_ONLY` (required to pick an explicit logging config when the build runs under a user-specified service account), and the GA `gcloud builds submit` can only stream logs from a GCS logs bucket, not from Cloud Logging.

## What

- Switch the submit step to `gcloud beta builds submit` — the beta command streams logs from Cloud Logging live into the step output. Same single blocking command; its exit code still fails the job.
- Install the `beta` component in the setup-gcloud step.
- Comment the `logging: CLOUD_LOGGING_ONLY` line in the cloudbuild config to explain the arrangement.

No new cloud resources (no logs bucket, no trigger migration).

## Caveat

Streaming needs the GitHub WIF service account to be able to read logs (`roles/logging.viewer` or equivalent). If it can't, nothing breaks — gcloud prints the log URL and waits, same as today, and the step output will say why. This PR's own integration-test run is the test.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -36,6 +36,10 @@ jobs:
 
       - name: Set up Google Cloud SDK
         uses: google-github-actions/setup-gcloud@v3
+        with:
+          # `gcloud beta builds submit` (used below for log streaming) needs the
+          # beta component.
+          install_components: beta
 
       - name: Submit integration tests to Cloud Build
         run: |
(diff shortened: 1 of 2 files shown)


theo-michel (approved):
LGTM

PR #575, opened 2026-07-28
Remove motor stress protection

theo-michel:
Removes the motor stress protection feature from `mars_arm` entirely.

## Why

The leaky-integrator overload cooldown shipped disabled by default (`stress_enabled: false`) and was never enabled on any robot. Rather than keep a dead setting in the settings UI, the whole feature goes away — runtime behavior is identical to the default (no cooldown, torque is never cut on load).

## What's removed

- `updateMotorStress()` and its control-loop call site (`arm_control.cpp`)
- `MotorStressTracker` and the 6 `stress_*` members (`arm_types.hpp`, `arm_node.hpp`)
- The 5 `stress_*` ROS parameters (`arm_node.cpp`, `arm_config.yaml`)
- The "Motor stress protection" knob in the webapp settings Arm section (`webapp/js/settings/catalog.js`)
- The `settings.yaml.template` line and the `docs/PARAMETERS.md` table entry

Deletion-only: 118 lines out, 1 in.

## Compatibility

A leftover `stress_enabled:` line in a robot's own `config/settings.yaml` is harmless — rclcpp ignores YAML overrides for parameters the node doesn't declare.

## Verification

- `pre-commit run -c .config/pre-commit-config.yaml` on the changed files passes (clang-format, settings/nav2 guards).
- `node --check` on `catalog.js` passes; the module imports cleanly.
- No local ROS build — Docker wouldn't start on my machine, so the C++ build is on CI. Grepped all seven removed identifiers across `ros2_ws/`, `webapp/`, `config/`, `docs/`: zero remaining references.

### DIFF
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -64,7 +64,6 @@
 # mars_arm:
 #   ros__parameters:
 #     max_jerk: 150.0               # rad/s^3 trajectory jerk limit (0 disables)
-#     stress_enabled: false         # motor-protection cooldown
 
 # ── Camera: main (stereo; hardware only) ──────────────────────────────
 # Resolution is staged: the sensor captures the full side-by-side stereo frame at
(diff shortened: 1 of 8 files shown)


DavidDobas (approved):
LGTM

karmanyaahm:
yeah for context i was experimenting w/ this and set up an autoresearcher on iron bird but it didn't decide on a good model for the dynamixel torque overload thingy - not sure how this got mergde

PR #589, opened 2026-08-01
fix(skills): qualify colliding physical-skill refs instead of dropping them

theo-michel:
## Problem

Two physical skills whose ids share a leaf — `local/wave` and `innate-os/wave` — both derive the class name `Wave`. Until now the loser was **dropped**:

- `physical_skills/__init__.py` emitted `# skipped innate-os/wave: class name Wave already taken by local/wave` and defined no class for it, so a live skill silently lost its typed ref;
- `render_dir_shims` returned `None` for its folder, **deleting** the committed `innate_skills/wave/__init__.py` — dirtying git and breaking `from innate_skills.wave import Wave` for everyone on that pack.

All of that from a flat-namespace clash the skill itself had no way to see.

## Change

The loser is re-claimed under a qualified name instead — `Wave` taken → `WaveInnateOs` (`class_name_for` applied to the id prefix), mirroring how the catalog already publishes the display name as `wave (innate-os)`. Every plain name is claimed before any fallback is considered, so a fallback can never displace a plain name.

The folder shim keeps exporting its **plain** name and aliases the import:

```python
from physical_skills import WaveInnateOs as Wave

__all__ = ["Wave"]
```

So the pack-local spelling `from innate_skills.wave import Wave` keeps working, a collision **rewrites** a committed shim rather than deleting it, and the shim reverts byte-for-byte once the collision goes away. Non-colliding shims are unchanged, verbatim.

Only entries with no derivable identifier at all are still skipped (`# skipped local/123: no usable Python class name`) / removed.

Also factors the repeated claim-order sort key out into `_claim_order`, so `_claimed` and `render_refs` can't drift apart.

## Verification

Exercised `_claimed` / `render_refs` / `render_dir_shims` directly (pure, no ROS) over a roster with a three-way `wave` clash plus a name-less id:

- claims: `Wave <- local/wave`, `WaveInnateOs <- innate-os/wave`, `WaveVendor <- vendor/wave`, `PickSocks <- innate-os/pick-socks`
- all three wave folders keep a shim; the two losers import qualified-as-`Wave`
- the single-skill (no-collision) shim is byte-identical to what the previous code produced
- `local/123` still yields a `None` shim and a skip comment

`ruff format --check` and `ruff check` pass — the alias spelling survives `ruff format` untouched, which matters since these shims are committed and CI formats them.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/skills/physical_refs.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/skills/physical_refs.py
@@ -47,7 +47,8 @@ _SHIM_MARKER = "# AUTO-GENERATED physical-skill ref"
 # Dir shims land in workspace/ and get committed, so they have to survive
 # `ruff format` untouched: hence "{name}" and not {name!r}, whose single
 # quotes ruff rewrites and CI then fails on. Safe because class_name_for()
-# only ever returns a valid identifier.
+# only ever returns a valid identifier. {imported} is just "{name}", or
+# "{qualified} as {name}" when another skill took the plain name.
 
 _DIR_SHIM = (
     _SHIM_MARKER
@@ -56,10 +57,10 @@ _DIR_SHIM = (
 
 Same class either way:
 
-    from physical_skills import {name}
+    from physical_skills import {imported}
 """
 
-from physical_skills import {name}
+from physical_skills import {imported}
 
 __all__ = ["{name}"]
 '''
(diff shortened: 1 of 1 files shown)


theo-michel:
Good catch — confirmed and fixed in 40f4863.

I reproduced three shapes where the single qualified attempt still dropped a live skill:

| Roster | Before | After |
|---|---|---|
| `local/wave` + `local/wave-innate-os` + `innate-os/wave` | `innate-os/wave` dropped | `WaveInnateOs2` |
| `local/wave` + `innate-os/wave` + `innate_os/wave` | `innate_os/wave` dropped | `WaveInnateOs2` |
| `local/wave` + bare `wave` | dropped | `Wave2` |

The third isn't reachable in practice — physical-skill ids are always built as `f"{skill_id_prefix_for(path)}/{basename}"`, so they always carry a pack prefix — but the counter covers it anyway, since `base` then degenerates to the taken plain name.

The fallback now counts up until a name is free, so every entry with a derivable identifier gets claimed and only a genuinely nameless id is skipped:

```python
for name, entry in losers:
    base = name + class_name_for(entry["id"].rpartition("/")[0])
    candidate, n = base, 2
    while candidate in claims:
        candidate, n = f"{base}{n}", n + 1
    claims[candidate] = entry
```

One trade-off worth flagging for reviewers, now called out in the `_claimed` docstring: plain and singly-qualified names are derived only from the skill's own id, so a committed shim never changes because some other skill appeared. Numbered names are roster-dependent — adding a skill can renumber one and rewrite its shim. That only applies to this doubly-colliding tail, and it's deterministic (losers are numbered in `_claim_order`), which seemed a better deal than deleting a live skill's committed shim.

Plain and singly-qualified output is byte-identical to the previous commit; the no-collision shim is still byte-identical to `main`.

DavidDobas (approved):
LGTM

PR #590, opened 2026-08-01
fix(agents): import-based discovery so broken agents stay visible

theo-michel:
## Summary

Agents were the last user of the pre-#542 file-exec loader. This fixes the two things that were **objectively broken** about that — not the authoring API, which is a separate, optional conversation (see *Deliberately not in scope*).

**1. A broken agent vanished.** `DynamicLoader.file_errors` was populated in four places and read by none. A typo in an agent file, a missing `id` (→ abstract → filtered), or an `id` written as a plain attribute instead of a property (→ `_validate_class` rejects it) all removed the agent from the picker with nothing but a log line. #542 fixed exactly this for skills and left agents behind.

**2. Ordinary Python didn't work in agent directories.** Agent modules were exec'd by file path and then popped from `sys.modules`, so a helper module next to an agent, a folder agent, or a subpackage import didn't resolve — and the class's `__module__` pointed at a module that no longer existed, which is why agents needed their own parallel reload path.

Both have the same root cause, so both are fixed by the same move: **agents register the way skills do.**

## What changed

- **`Agent.__init_subclass__` registry** — defining an `Agent` subclass in an imported module *is* the registration, same model as `Skill._registry`.
- **Real imports.** Agent packages are imported by name (`innate_agents.demo_agent`) via the skills machinery. `import_packages` + `live_class` were factored out of `workspace_import.py` with the skills path delegating to them — shared, not forked.
- **Broken entries are exact, not guessed** — a module that raises rosters by module name with its real error; a module that imports clean and registers nothing is a helper.
- **`AgentLoader` deleted** (317 lines → two functions). Net diff is **−16 lines** across the change.

**Broken agents in the UI:** published on `get_available_directives` in the **meta** dict (`directives[1]`), *not* the agent list — so a client that doesn't know the field can never offer one for selection. The webapp renders them as disabled options with the error as tooltip, mirroring the broken-skill rows in the skills menu.

**Load-time probing:** `build_agent_instances` exercises every surface the node reads later (`id`, `display_name`, `get_prompt`, `skill_ids`, `input_names`, icon). Besides turning late failures into visible broken entries, this fixes a latent bug on main — an agent whose `get_skills()` returned a bad ref crashed `_svc_get_directives` for **every** agent, not just itself.

**Hot reload** collapses onto the same model: agent dirs are watched as recursive roots, and any change evicts the workspace subtree and re-imports — identical to the SAS catalog. `reload_agent_by_name` and its filename-stem matching are gone. The current directive survives by id with its active-skill set intersected (a mid-session edit can't silently widen skills the user had narrowed); if its file no longer loads, it falls back to the default idle agent rather than continuing to run a stale instance whose file no longer says what it does.

**Preserved:** custom-over-innate id precedence, `source` stamping, icon loading, the `empty_directive` default, warning-only missing-skill validation, and every agent id (no namespacing — see below).

## Deliberately not in scope

- **The authoring API.** `id`/`display_name` derived from the class name, docstring-as-prompt, `skills = [...]` instead of `get_skills()` — all real ergonomics (7 shipped agents restate ~15 lines of boilerplate each), all *taste*. Cheap to add on top of this now that the registry hands over the class; not a justification for touching the load path.
- **Id namespacing** (`innate-os/demo_agent`). Consistent with skills, but its only real benefit is packs shipping agents — which needs `innate_agents`/`custom_agents` out of `NON_PACKAGE_DIR_NAMES` — and its breakage surface is persisted ids in the webapp and mobile app plus two hardcoded `"empty_directive"` literals. Revisit together, if at all.
- **Input devices.** Still on `DynamicLoader`, which is why it survives. Natural next step; after it, that module dies.

## Validation

- [x] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up`. — *N/A: no simulator behavior, config, or assets touched.*
- [x] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`. — *N/A: no asset changes.*

Verified in the sim image against the real ROS workspace: all 7 shipped agents load through the new path with correct ids, skill and input refs, icon data, and `source` stamping, plus a second pass exercising the evict/re-import reload cycle (fresh instances, identical roster). Separately, a synthetic workspace covering broken modules, helper-module and folder-agent imports, id conflicts, probe failures, and break→fix reload transitions.

`ruff check` + `ruff format --check` clean across `brain_client`; `node --check` on both webapp files.

> **Now with committed coverage** — `test/test_agent_loading.py`, 16 no-ROS pytest cases against a synthetic `$INNATE_OS_ROOT` workspace: broken-module surfacing with file:line, helper/folder/multi-agent imports, abstract/probe-failing/function-local agents as visible rows, custom-over-innate precedence across consecutive loads, every key-collision pairing, break→fix reload, and multi-agent carry-through. Verified green in the `innate-os-test` image.

## Notes for review

- **Reload is now all-or-nothing.** Under the import model there's no per-file agent reload, so a single agent edit re-imports every agent package (the skills catalog already works this way). Fine at current scale; worth knowing if the agent count grows a lot.
- **The eviction is the whole workspace subtree**, not just agent dirs — agent modules import skill classes and the generated `physical_skills` refs, so a narrower eviction would pin whatever those held at the previous load.
- The `hasattr(directive, "uses_gaze")` defensive check in `gaze_control.py` is now provably unnecessary, but I left it alone — out of scope.


DavidDobas (approved):
lgtm

PR #592, opened 2026-08-01
fix(webapp): inset select chevron and truncate long labels

ugtthis:
Problem: 
* native `<select>` chevron look crammed in the right with spacing that is inconsistent
* label clips if it does not fit within given area

Solution:
* add better padding and don't use native chevron
* truncate labels if they don't fit

## Comparing UI

https://github.com/user-attachments/assets/c338458e-9132-44ce-b40b-29ab22bdfdd2

Why 30px right padding?

* 30 = 10px (icon width) + 10px (icon's edge inset) + 10px (gap between label and icon)
* makes it so the right padding of `<select>` matches the left padding(10px)

Ideally in the future this could have a custom select if it becomes a priority but not within the scope of this PR

<br>

## Truncating label

https://github.com/user-attachments/assets/b674b9d5-1cad-4659-b271-af51a51aa424

### DIFF
--- a/webapp/css/app.css
+++ b/webapp/css/app.css
@@ -95,11 +95,11 @@ button {
 /* Native select chevron cannot be inset reliably.
    30px end padding = 10px inset + 10px icon + 10px text gap. */
 .agent-directive,
-.record-skill,
 .log-select,
+.modal-select,
+.record-skill,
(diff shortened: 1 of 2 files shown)


theo-michel (approved):
Great works thank you!

PR #595, opened 2026-08-03
feat(teleop): copy leader joint positions to the clipboard

theo-michel:
## What

Adds a small copy icon button next to **Engage follow** in the teleop arm panel. Clicking it copies the six live leader-arm joint ticks to the clipboard as `[2048, 1024, ...]` — the shape you paste into a skill when scripting a pose.

## Details

- Uses the existing `copyToButton` / `ICON_COPY` helpers from `webapp/js/clipboard.js`, so it flashes a check on success (✕ on failure) and falls back to the legacy textarea path on insecure origins (`http://<robot-ip>`).
- Engage and the copy button now share an `.arm-engage-row`; the row is what hides when the leader arm isn't reading, so the copy button is only offered when there are positions to copy.

## Testing

`tsc --noEmit` over `webapp/` reports no new errors (`armPanel.js` and `clipboard.js` clean; the remaining errors are pre-existing in other files). Not exercised on hardware — the row only appears with a leader arm connected over WebSerial.

### DIFF
--- a/webapp/css/app.css
+++ b/webapp/css/app.css
@@ -1609,6 +1609,29 @@ button {
   color: var(--accent);
 }
 
+/* Engage sits next to a small icon button that copies the joint ticks. */
+.arm-engage-row {
+  display: flex;
+  gap: 8px;
+}
+
+.arm-engage-row .arm-engage {
+  flex: 1;
+}
+
+.arm-copy {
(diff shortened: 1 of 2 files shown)


karmanyaahm (approved):
ok

PR #596, opened 2026-08-03
feat(arm): one blocking, raising arm SDK

theo-michel:
## What

Replaces the arm interface with a single raising SDK — blocking by default, non-blocking on request — and migrates every shipped skill to it. The pre-0.7 (0.6.0) surface is removed outright.

**The new `Manipulation` core**
- `move_to` / `move_by` / `follow` / `move_joints` / `rest`, `gripper_open` / `gripper_close` — by default every motion blocks until the arm settles and raises `ArmFailed` / `ArmUnhealthy` instead of returning `bool`.
- **Non-blocking motions**: every motion method takes `block=False` to return as soon as the command is accepted, so the skill keeps running (drive the base, watch a camera) while the arm travels. `wait()` joins the in-flight motion — returning the settled pose or raising on failure — and `moving` reports whether one is still in flight. Any new command supersedes an unjoined motion; non-blocking motions are unverified until joined.
- Reads return the typed `Arm` dataclass (the same type as the ambient `arm:` feed), now with `rpy`.
- **Standing grip target**: the interface tracks the last *commanded* j6 and every motion carries it. Under current-based position control the standing position error *is* the grip force, so re-seeding j6 from the measured position drops a held object — that footgun is gone, along with the per-call `gripper=` threading every trajectory used to need.
- Adds `Waypoint` for `follow()` trajectories, a `/mars/arm/status` subscription so torque state survives out-of-band toggles from the webapp, and `halt()` as a documented no-op framework hook.

**Skill migrations** — `arm/*`, `arm_rest_position`, the chess skills, and `pick_any_object`. `arm_circle_motion` becomes one smooth `follow()` per loop instead of sixteen stop-and-go moves, which also removes a sim/hardware divergence. `arm_utils` is untouched: `torque_on` / `torque_off` / `reboot_servos` keep their bool contract.

**Compatibility** — none kept: the 0.6.0 shims (`move_to_cartesian_pose`, `move_cartesian_trajectory`, `move_to_joint_positions`, `solve_ik`, `open_gripper`, `close_gripper`, `get_current_end_effector_pose`, `get_current_orientation_rpy`, `spin_node_to_refresh_topics`) are removed, along with their deprecation machinery and the file-wide pyright pragma they required. `test_manipulation_surface.py` now pins the *new* API surface (signatures, defaults, constants) and asserts the removed shims don't creep back.

Also on this branch: Mad mode folds the arm into a bracing pose on the rising edge of the speed-scale write, since a collision at 2.0x can break an extended arm.

## Testing

- `test/test_manipulation_unit.py` + `test/test_manipulation_surface.py` — **82 passed** (full fast-unit CI block: 149 passed, 1 skipped). Both run without ROS via `test/ros_stubs.py`.
- `ruff check` / `ruff format --check` clean; the file-level pyright suppression is gone and the file is clean under pyright (modulo environment-only missing-import noise without ROS).
- The bracing pose and the gripper trip thresholds (`GRIPPER_MAX_STRENGTH = 0.6`; 0.7 and 0.8 both tripped) come from hardware captures — see the individual commit messages for what was verified live and how.


ros2_ws/src/brain/brain_client/brain_client/state/arm.py
```diff
@@ -2,9 +2,25 @@
...
+
+    sinp = 2 * (qw * qy - qz * qx)
+    pitch = math.copysign(math.pi / 2, sinp) if abs(sinp) >= 1 else math.asin(sinp)
+
+    siny_cosp = 2 * (qw * qz + qx * qy)
+    cosy_cosp = 1 - 2 * (qy * qy + qz * qz)
+    yaw = math.atan2(siny_cosp, cosy_cosp)
+    return (roll, pitch, yaw)
+
+
```
karmanyaahm, line 11: thought about whether or not this should be abstracted out into a helper geometry class, but the import gymnastics are probably not worth

karmanyaahm: pushed the abstraction.

ros2_ws/src/brain/brain_client/brain_client/state/arm.py
```diff
@@ -38,3 +54,23 @@ def position(self) -> tuple[float, float, float]:
     def orientation(self) -> tuple[float, float, float, float]:
         """(qx, qy, qz, qw)."""
         return (self.qx, self.qy, self.qz, self.qw)
+
+    @property
```
karmanyaahm: made this a cached property

ros2_ws/src/brain/brain_client/brain_client/robot/manipulation.py
```diff
@@ -192,73 +437,314 @@ def _arm_state_callback(self, msg: JointState):
...
+        x: float,
+        y: float,
+        z: float,
+        *,
+        roll: float = 0.0,
+        pitch: float = 0.0,
+        yaw: float = 0.0,
+        duration: float = 1.5,
+        grip: float | None = None,
+        tol_xy: float | None = 0.05,
```
karmanyaahm: its not obvious that tol means target tolerance. It's also not directly in the docstring

workspace/innate_skills/chess/detect_opponent_move.py
karmanyaahm, line 1: are these integration tested

ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
```diff
@@ -806,6 +810,41 @@ class AppControl : public rclcpp::Node {
...
+        }
+        request->time = 2.0;
+        RCLCPP_INFO(this->get_logger(), "Mad mode: folding the arm to its bracing pose");
+        arm_goto_client_->async_send_request(
+            request, [this](rclcpp::Client<mars_msgs::srv::GotoJS>::SharedFuture future) {
+                const auto response = future.get();
+                if (!response || !response->success) {
+                    RCLCPP_WARN(this->get_logger(), "Mad mode: arm safe-fold goto was rejected");
+                }
+            });
```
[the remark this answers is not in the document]
karmanyaahm: nbd

ros2_ws/src/brain/brain_client/brain_client/robot/manipulation.py
karmanyaahm: fixing

karmanyaahm:
moved auto recovery to a different PR https://github.com/innate-inc/innate-os/pull/598

karmanyaahm (approved):
yay!

axelpey (approved):
Tested all skills involving manipulation, works

DavidDobas (approved):
Great

DavidDobas:
Maybe we can expose some of them as skills in arm folder, so it’s easy to use from webapp?

workspace/innate_skills/chess/pick_up_piece_simple.py
```diff
@@ -115,71 +115,71 @@ def _d(self, seconds: float) -> float:
...
+        self.manipulation.move_to(
+            x,
+            y,
+            z,
+            roll=self.FIXED_ROLL,
+            pitch=pitch,
+            yaw=yaw,
+            duration=self._d(duration),
+            tolerance_xy=None,
+            tolerance_z=None,
```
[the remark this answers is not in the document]
theo-michel: Should be fine

ros2_ws/src/mars_bot/mars_arm/mars_arm/arm_trajectory.cpp
```diff
@@ -312,11 +312,15 @@ void MarsArmNode::armGotoJSTrajectoryCallback(const std::shared_ptr<mars_msgs::s
                 }
             }
             waypoints.insert(waypoints.begin(), start);
-            if (!durations.empty()) {
-                durations.insert(durations.begin(), durations[0]);
-            } else {
-                durations.insert(durations.begin(), 0.5);
+            // One duration per waypoint means durations[0] already paces this
+            // prepended approach; legacy waypoints-1 callers get a copy of the
+            // first segment instead.
```
axelpey, line 309: too many comments

workspace/innate_skills/arm/arm_circle_motion.py
```diff
@@ -2,55 +2,62 @@
 # Copyright (c) 2026 Innate Inc
 import math
 
-from innate import Manipulation, Skill, SkillReturn
+from innate import ArmFailed, ArmUnhealthy, Manipulation, Skill, SkillReturn, Waypoint
```
DavidDobas, line 7: Shouldnt we move ArmFailed, ArmUnhealthy to innate.exceptions or something like that? Would be categorised a bit nicer

ros2_ws/src/brain/brain_client/brain_client/nodes/arm_sdk_server.py
```diff
@@ -0,0 +1,265 @@
...
+so the page is always available; the webapp front door proxies /armsdk/api/*
+here, which keeps the raw arm-control port on-box and same-origin.
+
+The Manipulation feeds cost ~600 msg/s of callback dispatch while live, so
+they only run while the page does: constructed lazy, started on the first
+request, and parked again after IDLE_PARK_S without one (the page polls every
+250 ms, so an open tab keeps them warm).
+
+Not skill code — plain time.sleep and blocking calls are fine here.
+"""
```
DavidDobas, line 18: Why rest API when elsewhere we use rosbridge?

From Claude:
You can keep the logic on the robot and keep rosbridge: wrap Manipulation in a ROS action server instead of an HTTP server. MoveTo.action with goal / feedback / result, and the page drives it over the socket that's already open.

And this codebase already does exactly that pattern. [brain_messages/action/ExecuteSkill.action](vscode-webview://1imqf2e6j65mvbluerc9ul1peehalu7q2pq70vd0g0d8sva9brud/ros2_ws/src/brain/brain_messages/action/ExecuteSkill.action) is how the webapp runs skills — long-running, robot-side logic, driven from the browser over rosbridge. RunStereoCalibration does the same for a whole capture session. The precedent is right there.

Actions are also a genuinely better fit than HTTP for this specific shape. move_to blocks for seconds, so you want cancellation and progress — an action gives you both. The HTTP design has to hand-roll it: torque_off is special-cased to bypass motion_lock so the in-flight motion fails fast on a de-energized arm. That's an abort path built out of de-energizing the hardware because the transport has no cancel. Meanwhile [constants.js:202](vscode-webview://1imqf2e6j65mvbluerc9ul1peehalu7q2pq70vd0g0d8sva9brud/webapp/js/constants.js#L202) notes the skill action "cancels the running skill regardless of which client sent its goal" — cross-client arbitration, free, which is precisely the guarantee motion_lock tries to provide and can't.

workspace/innate_skills/chess/pick_up_piece_simple.py
```diff
@@ -288,17 +304,42 @@ def execute(
...
+                    f"so its position is unknown: {safe_pose_error}"
+                )
+            raise
+
+        safe_pose_error = self._return_to_safe_pose()
+        if safe_pose_error is not None:
+            self.fail(
+                f"Moved piece from {square} to {place_square}, but the arm failed to return "
+                f"to safe pose and is in an unknown position: {safe_pose_error}"
+            )
```
DavidDobas: The failure handling here doesn't look that nice, couldnt we do try catches elsewhere so that skills stay clean?

DavidDobas:
Tried the `/armsdk` page in the sim today. It surfaced a pre-existing arm bug worth knowing about — not introduced by this PR, but the jog console is what makes it easy to hit.

Jogging `+x` from a side-yawed pose sends the end-effector *down*.

The cause is a joint1-dependent floor on joint2 — the shoulder can't fold deep while the base yaw points the arm at the body. Sensible guard:

https://github.com/innate-inc/innate-os/blob/8b0ecc7e9460e9b59127f02659f6e28fe9d98e25/ros2_ws/src/mars_bot/mars_arm/mars_arm/arm_control.cpp#L373-L396

But the IK node solves with `ChainIkSolverPos_LMA`, which is fully unconstrained — it doesn't enforce URDF joint limits, never mind the joint1/joint2 coupling:

https://github.com/innate-inc/innate-os/blob/8b0ecc7e9460e9b59127f02659f6e28fe9d98e25/ros2_ws/src/mars_bot/mars_arm/mars_arm/ik.py#L56

So the SDK plans a pose, `move_to` sends those joints, and the driver silently overwrites joint2 on the way to the encoders. **What the arm reaches is not what IK solved for**, and the only trace is `err_xy` in the log. The sim mirrors the same ramp in [`core.py#L65-L77`](https://github.com/innate-inc/innate-os/blob/8b0ecc7e9460e9b59127f02659f6e28fe9d98e25/ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/core.py#L65-L77), so it reproduces there.

Verified by streaming explicit joint vectors with `j2 = -1.45`, bypassing IK entirely. Settled joint2 tracks the computed floor to ~0.004 rad:

| j1 commanded | j2 floor | j2 settled | |
|---|---|---|---|
| 1.40 | −1.571 | −1.451 | honored |
| 1.30 | −1.571 | −1.451 | honored |
| 1.20 | −1.307 | −1.307 | clamped |
| 1.10 | −0.778 | −0.775 | clamped |
| 0.90 | −0.250 | −0.261 | clamped |

Replayed as 2 cm `+x` jogs from `joint1 = 1.445` — just past the 1.25 threshold, which is why the shoulder was allowed down at −1.39 to begin with:

```
j1 1.2845  j2 -1.2818  z 0.0584
j1 1.1400  j2 -0.9958  z 0.0446   (-1.4 cm)
j1 1.0131  j2 -0.4378  z 0.0271   (-1.75 cm)
```

3.1 cm of unrequested descent across two forward steps, joint2 pinned to the floor each time.

No action needed in this PR. But it's an argument for surfacing clamp events — the C++ logs a throttled warning, the sim clamps silently, and `Manipulation` sees neither, so a jogged move that got overridden looks like a successful move with a bit of error.

theo-michel:
Addressed the open review threads in d3218b90 / 28f65fbd (replying here — I have a draft review pending, which blocks inline replies):

**`innate.exceptions` (@DavidDobas, arm_circle_motion.py)** — agreed and done in 28f65fbd: `ArmFailed`/`ArmUnhealthy` now live in `brain_client.robot.exceptions` and are re-exported together with `SkillFailed`/`SkillCancelled` from `innate.exceptions` (and at the `innate` top level). Old import paths keep working.

**Chess failure handling (@DavidDobas, pick_up_piece_simple.py:326)** — agreed it read badly. d3218b90 folds the dual-failure logic into an `_ending_at_safe_pose()` context manager, so `execute()` is back to one `with` line around `_move_piece()`.

**Cancellation discards cleanup failure (greptile, :316)** — real; fixed in the same refactor. A Stop whose safe-pose cleanup then fails now re-raises `SkillCancelled` carrying the stranded-arm message (the server surfaces `str(e)` as the CANCELLED result), instead of reporting a clean cancel while the arm sits over the board. All six exit paths (ok/fail/cancel × cleanup ok/miss) are exercised and behave.

**REST vs rosbridge (@DavidDobas, arm_sdk_server.py)** — the action-server route is the principled rosbridge answer, and the pasted argument is right that actions give cancel + cross-client arbitration for free. The reason this page is HTTP anyway: its purpose is to exercise the *actual Python `Manipulation` class* — block-until-settled, FK-verify + `recover()` retry, the standing-grip-target rule — so the logic must run robot-side in Python either way; the wire to it is page-private RPC, and HTTP is the thinnest wrapper that didn't need a `mars_msgs` schema for a jog vocabulary. Wrapping Manipulation in a `MoveTo.action` is the right move the moment arm control needs more than this one page (cancel from other UIs, arbitration across clients) — I'd take that as a follow-up rather than block this PR on it. If you feel strongly it should happen now, let's talk scope.

theo-michel:
Update on the REST-vs-rosbridge thread: flipped in this PR after all (ffa1deb3). The page now drives a `brain_messages/action/ExecuteArmCommand` action into the same Python Manipulation server, sliders stream over a topic it subscribes (the velocity-clamped slew stays SDK-side), and pose/joints/torque ride the driver topics directly. The localhost HTTP server, the front-door proxy route + origin gate, and the page's 4 Hz poll are gone — a page that's just watching the arm now costs the robot nothing.

Verified live on R7-27 end-to-end (browser → rws → action server → arm; jogs settling with mm readout). One operational note: **rws loads message typesupport once per process**, so after pulling this you need a stack restart (or at least kill rws and let it respawn) before the page's action goals resolve — otherwise you'll see \"typesupport library could not be found\" in the console.

@DavidDobas your sim bringup (a47eb122) merged in cleanly on top — the launch addition and model-serving fix carry over unchanged; only the 8090/proxy part of its commit message is now historical. With the flip, the sim page also stops needing anything HTTP-side.

PR #601, opened 2026-08-05
Sim challenge

karmanyaahm:
## Summary

-

## Validation

- [ ] I ran the relevant local validation, or explained why it was skipped.
- [ ] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up`.
- [ ] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`.

PR #602, opened 2026-08-05
feat(sim): droppable scenario props — human, soccer ball, labrador

karmanyaahm:
## Summary

Split out of #601 (part 1 of 3). Adds three free-jointed props to the apartment world and a way to place them, with **no challenge/judging layer** on top — that lands in the stacked follow-up.

**Driver**
- `world.py`: `OBJECT_KINDS`/`OBJECT_MESHES`/`OBJECT_PARK`/`OBJECT_DROP_Z`, one body per prop in `build_world_xml`, parked off-map until dropped. The human is a posed scan colliding as its own convex hull; the ball is a sphere primitive (`condim 6`, so a kicked ball rolls and settles) and the dog collides via CoACD pieces — both with the textured mesh drawn over them, so the robot cameras see what the viewer draws.
- `core.drop_object()` teleports a prop above `(x, y)` with a yaw and lets physics settle it; `object_poses()` publishes ground truth for everything dropped. Meshes are optional — older asset bundles fall back to bare primitives.
- `world_server` publishes `objects` in the observer state stream, and `serve_state`'s receive side now takes scenario commands (`{"op": "drop_object", ...}`). World manipulation stays on the observer channel and out of the robot's RPC surface.

**Assets**
- `sim/tools/convert_objects.py` converts the Blender exports to the MuJoCo convention (Z-up meters, bbox-centered), pulls the basecolor out of the viewer GLB, and CoACD-decomposes the dog into collision pieces plus a wireframe soup for the browser overlay.
- `publish_assets.py` bundles `humans/` and `objects/`; lock repinned; the human model is credited per CC BY 4.0.

**Viewer**
- `scene.ts` draws a GLB per dropped prop at its MuJoCo pose, normalized into the same body frame physics uses, with collision wireframes on the existing "collisions" toggle.
- `simStage.ts` gains a drop picker: pick a kind, press to mark the spot, drag to point the yaw, release to drop.

## Stack

1. **this PR** — props + drop command
2. #604 — challenge engine (based on this branch)
3. #603 — unrelated dev-experience docs, independent

Based on the same commit as #601 (`504f0d0b`), so the diff is exactly the prop half of that PR. `origin/main` has since merged #577, which touches the same sim files — a rebase is still needed and is not part of this split.

## Known open review items (carried over from #601, deliberately not fixed here)

- [ ] `SIM_ASSET_UNITS` in `sim/launcher/runtime.py` gains `work/humans` but **not** `work/objects`, while `publish_assets.py` publishes both. `ensure_sim_assets` only moves listed units out of staging, so the ball/dog meshes and collision pieces are silently discarded — MuJoCo falls back to a bare sphere/box while the browser still draws the real GLBs.
- [ ] `scene.placementMode` is assigned by `armDrop()` but only read inside `setView()`, so orbit controls are never actually paused during a placement drag (and can get stuck disabled if the view changes while armed).

## Validation

- [x] `ruff check` / `ruff format --check` clean (three E741s in `convert_objects.py` fixed — they were failing `format` on #601).
- [x] `tsc --noEmit` clean in `sim/viewer`.
- [ ] Not yet run against a live `./innate-sim up`.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/core.py
+++ b/ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/core.py
@@ -178,11 +178,17 @@ class VirtualMars:
         visual_dir = ASSETS_DIR / "apartment_visual"
         visual_rooms = world.find_visual_rooms(visual_dir) if visual_dir.is_dir() else {}
 
+        # Prop meshes are optional: older asset bundles lack them (the human
+        # body is then absent; ball/dog fall back to bare primitives).
+        human_obj = ASSETS_DIR / world.HUMAN_MESH
+        object_meshes = {k: p for k, rel in world.OBJECT_MESHES.items() if (p := ASSETS_DIR / rel).exists()}
         xml = world.build_world_xml(
             rooms,
             include_placeholder_robot=False,
             visual_rooms=visual_rooms,
             texture_max=_texture_cap(self._render_w),
+            human_obj=human_obj if human_obj.exists() else None,
+            object_meshes=object_meshes,
         )
         # Lidar rays hit only the textured visual meshes (true surfaces, like
         # a real lidar) when available -- without them, fall back to all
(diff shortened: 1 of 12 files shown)


karmanyaahm:
**Why there are no `format` / `integration-test` checks here:** this branch conflicts with `main`, so GitHub can't build the PR merge ref and never schedules `pull_request`-triggered workflows. #601 is `CONFLICTING` for the same reason — the split didn't introduce it.

The conflict is real, not textual. `main` merged #577 after this branch was cut, which added its own manipulation-prop system (`GRASP_OBJECTS`, `GRASP_PARK_XY`, prop MJCF) to the same `world.py`/`core.py`/`scene.ts`/`simStage.ts`, plus a competing `drop_objects` viewer button. The two prop systems need reconciling by hand, and `sim/sim-assets.lock` pins two different published bundles (`sim-assets-65dcba740aba` on main vs `sim-assets-ed27954801fd` here) — resolving that needs a republished bundle containing both main's assets and `humans/`+`objects/`, which only someone with the source-asset checkout can produce.

Lint and types were verified locally against this base: `ruff check` / `ruff format --check` clean, `tsc --noEmit` clean in `sim/viewer`.

PR #604, opened 2026-08-05
feat(sim): challenge engine — declarative tasks judged against ground truth

karmanyaahm:
## Summary

**Stacked on #605** — review that one first; this PR's diff is only the challenge layer.

A challenge is a Python module in `sim/challenges/` exporting `CHALLENGE = Challenge(...)`: a scene setup (which props to drop where), an ordered goal checklist of predicates, and an optional time limit — the same sidecar shape as #605's `sim/props/`.

The engine is hosted by the world server, which owns ground truth, so the robot stack never sees any of it — that is what makes the verification honest. `tick()` runs on every observer state broadcast and returns the `challenge` block embedded in the state stream, so any frontend is a thin renderer; `start`/`abort` arrive as observer commands alongside the prop ops; results persist to `workspace/challenges.json`. Skill completions stream in from rosbridge (`/brain/skill_status_update`) best-effort — with rosbridge down, world-state challenges still work and `SkillDone` goals simply never fire.

Predicates: `Near`, `InCircle`, `InRect`, `Hold` (dwell), `SkillDone` (optionally guarded), `AllOf`/`AnyOf`. A broken challenge file or predicate degrades that challenge, never the sim.

Four challenges ship: `victory_lap`, `lifesupport`, `rescue`, `shepherd`.

**Webapp**: a top-left challenge panel on the Agent page renders the roster, goal checklist, timer and pass/fail banner, and sends start/abort. Sim-only — it keys off `session.onChallenge` existing — and the telemetry card moves into a shared top-left column with it.

**Docs**: `sim/README.md` gains a user-facing Challenges section (what they are, how to run one, how to write one) next to skills and agents, plus a `challenges.py` section in "How it works". Its observer-stream description used to end with "future challenge UIs/graders are just more clients" — that sentence is now the present tense.

## Rebased onto the prop sidecars (#605)

Was stacked on #602; #605 supersedes it, so the engine now speaks the sidecar API:

- `Drop("human", x, y)` names a **prop** (`sim/props/*.py`), placed with `drop_prop_at`. A name no sidecar claims is skipped with a warning instead of silently doing nothing.
- The local `OBJECT_CENTER_OFFSET` table and its quaternion math are gone. Distances measure to a prop's visual centre via `PropRegistry.center_xy`, which reads the sidecar's own `center_offset` — one place knows that the human scan stands feet-at-origin. `VirtualMars.object_centers()` (9 lines, the only addition outside the challenge layer) gathers those centres under the sim lock so `tick()` stays pure evaluation with no sim access.
- `start_challenge`/`abort_challenge` join `_run_scenario_command`, which already wraps each command in its own `try` — so the malformed-command review item is resolved by construction.
- `load_challenges` and the challenge filenames follow the same conventions as the prop sidecars (see below).

## Stack

1. #605 — props behind sidecars
2. **this PR** — challenge engine

(#603, dev-experience docs, is independent.)

## Review items fixed here

- [x] **`ChallengeEngine.start()` race.** It published the active challenge, *then* took the sim lock to reset the world and drop props, so a physics tick landing in that window judged goal 0 against the previous run's world and computed `elapsed_s` from a stale `started_t` — on a server up longer than the time limit, a fresh run was born `failed`. It now deactivates, builds the scene, and publishes the whole run in one atomic step. Starting over a live run records it as aborted instead of dropping it, which is what Abort already did, so both go through `_deactivate()`.
- [x] **`AllOf`/`AnyOf` short-circuited**, so a `Hold` behind a decided sibling never got its `update()` and silently restarted its dwell every tick. Both now judge every child before combining.
- [x] **Roster on every broadcast.** What a challenge *is* (title, brief) now ships once per connection in the same frame as the prop roster; only what can change rides the stream. 714 bytes once, then 294 per broadcast while a run is going and 95 while idle, against ~950 before. Progress ships every tick rather than only when it changes — the stream is latest-wins, so a client that skipped the one frame carrying an update would keep a stale roster. `SimSession` merges the halves, so the panel's view is unchanged.
- [x] **`load_challenges` takes a list of roots**, like `load_props`, so an asset bundle can ship a challenge pack next to the props it needs.
- [x] **Filenames renumbered** by what the challenge asks for rather than when it was written (10s skills, 20s people, 30s moving things around), since `sorted(glob)` is the roster order the user sees.
- [x] **Two regression tests** in `sim/sandbox/test_driver_core.py` — neither needs GL or rosbridge. The race one reproduces the physics thread landing on both sides of the setup window deterministically, by ticking from inside a patched `reset` and a patched drop with the sim clock set past the time limit; against the pre-fix engine it fails on the first tick.
- [x] `tick()`'s docstring claimed the physics thread; `publish_state()` runs it on observer threads too.
- [x] `handle_observer_command` used to drop the whole observer connection on a malformed command — #605's `_serve_scenario_commands` catches per command.
- [x] The agent page built the telemetry card with an empty class name and appended it after the parts list; the column's DOM order now reads in the order it renders.

**`victory_spin` is not missing.** The user writes it during the onboarding tutorial, so `victory_lap` is that tutorial's payoff — the first thing their own skill gets used for. `10_victory_lap.py` now says so, since its absence from `workspace/innate_skills/` reads like a bug otherwise.

## Follow-ups (not in this PR)

The engine is deliberately the smallest thing that judges honestly. What it opens up:

- **Hand the brief to the agent.** Starting a challenge should be able to send its `brief` straight into the agent chat instead of the user copying it across — the panel already keeps the text selectable for exactly that reason. Wants a small "post a message" path from the observer channel to the brain, which is the one place this design deliberately does not reach yet.
- **Randomised placement.** `Drop` takes fixed coordinates, so a challenge is memorisable: same spot, same route, every run. A `DropAnywhere(kind, region=...)` that samples a free spot from the nav map (as the placement probing in #605 already does by hand) would make "find the person" mean finding, and make repeat runs comparable rather than identical.
- **Run history and run data.** `challenges.json` keeps one row per challenge (passed, attempts, best time); nothing keeps the *runs*. A per-run record — when, how long, which goals fell and at what time, what the robot's own skill trace was — is what turns this from a pass/fail light into something you can compare two agent versions with. The panel would grow a history view; the same data is what a leaderboard or a regression gate would read.
- **More predicates.** Orientation (`FacingToward`), velocity/stillness, "prop is off the floor" (a real grasp check rather than a proximity proxy), and region-relative composition. Each is a small class; the reason to wait is that the right set falls out of writing more challenges.
- **Authoring loop.** Sidecars load at world-server startup, so iterating on a challenge means a full `down && up`. A reload command on the observer channel would make writing one interactive.
- **Broad predicate coverage.** Only the two regressions above are tested; `Near`/`InCircle`/`InRect`/`SkillDone`-guard/ordered-goal latching are exercised by hand.

Smaller things, noted so they aren't rediscovered:

- `_record()` writes `challenges.json` synchronously while holding `_mutex`, from whichever thread ticked. Once per state transition, so not hot, but it is disk I/O on the physics thread.
- `attempts` is only persisted when a result is recorded, so a clean shutdown mid-run loses the increment. The in-memory roster the panel reads is always right.

## Validation

- [x] **Played end to end** against a live `./innate-sim up` — challenges start from the panel, the world resets and the props land, goals tick over, and the run scores.
- [x] `ruff check` / `ruff format --check` clean; `tsc --noEmit` clean in `sim/viewer`; `node --check` on the new webapp modules; `npm run build:lib` succeeds after the wire split.
- [x] `uv run sandbox/test_driver_core.py` passes, including the two new regressions; both were confirmed to fail against the pre-fix engine (the race on the first tick of the setup window, the dwell reporting `False` after 6 s).
- [x] Headless end to end through `WorldServer`: `{"op": "start_challenge", "id": "rescue"}` drops the human, the state payload carries the `challenge` block, goals latch in order, the dwell passes the run at 6.0 s, and `challenges.json` records the pass and best time. `abort` clears the active block.
- [x] All four challenges load; `Near` / `Hold` / guarded `SkillDone` behave, including a guard that correctly fails at distance.
- [x] The human's centre correction survives the port to the sidecar API: body origin `(-5.690, 4.590)` → centre `(-4.826, 4.590)` under the challenge's −90° yaw, i.e. 0.864 m along the body's head axis.
- [x] A challenge naming a prop no sidecar claims warns and keeps running rather than failing the sim.


karmanyaahm:
**Why there are no `format` / `integration-test` checks here:** this branch conflicts with `main`, so GitHub can't build the PR merge ref and never schedules `pull_request`-triggered workflows. #601 is `CONFLICTING` for the same reason — the split didn't introduce it.

The conflict is real, not textual. `main` merged #577 after this branch was cut, which added its own manipulation-prop system (`GRASP_OBJECTS`, `GRASP_PARK_XY`, prop MJCF) to the same `world.py`/`core.py`/`scene.ts`/`simStage.ts`, plus a competing `drop_objects` viewer button. The two prop systems need reconciling by hand, and `sim/sim-assets.lock` pins two different published bundles (`sim-assets-65dcba740aba` on main vs `sim-assets-ed27954801fd` here) — resolving that needs a republished bundle containing both main's assets and `humans/`+`objects/`, which only someone with the source-asset checkout can produce.

Lint and types were verified locally against this base: `ruff check` / `ruff format --check` clean, `tsc --noEmit` clean in `sim/viewer`.

Additionally: while this PR targets `sim-drop-objects` rather than `main`, `format.yml` and `integration-test.yml` only trigger on `branches: [main, develop]`, so they won't run here regardless. GitHub will auto-retarget this to `main` when #602 merges, and the checks will run then.

theo-michel (approved):
Tried it it's cool thanks!

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/world_server.py
```diff
@@ -130,10 +134,23 @@ def publish_state(self) -> None:
             x, y, yaw = self.sim.pose()
             joints = self.sim.joint_positions()
             objects = self.sim.object_poses()
+            # Prop CENTRES for the judge (props.py center_offset): a distance
+            # to the human has to mean its body, not the feet its origin sits
+            # at. Gathered here because the judge runs without the sim.
+            centers = self.sim.object_centers()
             sim_time = float(self.sim.data.time)
+        # Judged outside the sim lock: pure evaluation over the gathered state.
+        challenge = self.challenges.tick(sim_time, (x, y, yaw), centers)
```
theo-michel: **Race: a tick whose state was gathered *before* `start()` can be delivered *after* the run is published.**

The state (`sim_time`, pose, centers) is gathered under `self.lock`, the lock is released, and only then does `tick()` run. If the physics thread is preempted in that gap and an observer thread runs `start_challenge` to completion — `sim.reset()` zeroes `data.time`, so `started_t ≈ 0` — the stale tick then arrives with the pre-reset clock and poses: `elapsed_s = 900 − 0 > time_limit_s` instantly records the brand-new run as **failed** (burning an attempt in `challenges.json`), and goal 0 is judged against the previous run's prop layout.

This is the same bug class the PR's `start()` fix closes, but from the other side of the window: `check_challenges` in `test_driver_core.py` injects ticks *during* `start()` (while `active` is still `None`), so it can't catch a tick gathered before `start()` and delivered after. A monotonic-`t` guard in `tick()` (drop ticks with `t < started_t` or `t <` last seen `t`), or a generation counter bumped by `start()` and stamped into the gathered state, would close it.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -0,0 +1,483 @@
...
+        self.started_t = 0.0
+        self.elapsed_s = 0.0
+        if self.challenges:
+            print(f"[challenges] loaded: {', '.join(self.challenges)}", flush=True)
+
+    # -- persistence --
+
+    def _load_progress(self) -> dict:
+        try:
+            return json.loads(self.progress_path.read_text())["challenges"]
```
theo-michel: **A wrong-typed (but valid-JSON) `challenges.json` kills the physics thread.**

This catches parse errors but not type surprises: `{"version":1,"challenges":null}` (hand-edited, or written by another tool) makes `["challenges"]` return `None` without raising, so `self.progress` is `None`. The first `tick()` → `_block()` → `self.progress.items()` then raises `AttributeError` *outside* the per-goal `try` — and `physics_loop` has no guard, so the daemon physics thread dies and state broadcasts stop for every viewer. That breaks the module's own contract ("a broken challenge file degrades that challenge, never the sim"), and it fires even with no challenge active.

Cheap fix: validate the loaded value, e.g.
```python
data = json.loads(self.progress_path.read_text())["challenges"]
return data if isinstance(data, dict) else {}
```

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -0,0 +1,483 @@
...
+
+    def tick(self, t: float, pose: tuple[float, float, float], centers: dict[str, tuple[float, float]]) -> dict:
+        """Advance the active challenge and return the state-stream block."""
+        with self._mutex:
+            challenge, events, self._events = self.active, self._events, []
+            if challenge is not None and self.state == "running":
+                state = WorldState(t=t, robot=pose, centers=centers)
+                self.elapsed_s = max(0.0, t - self.started_t)
+                try:
+                    idx = self.goal_done.index(False)
```
theo-michel: **Drained skill events meant for later goals are discarded, not deferred.**

`tick()` drains the whole `_events` batch but hands it only to the first unmet goal's predicate; whatever that goal doesn't match is dropped on the floor. Concretely, in `20_lifesupport`: the agent waves from 1.6 m (guard fails, goal 0 stays unmet), then dutifully emails — the `send_email` completion is consumed by a goal-0 tick and lost. When the robot later closes in and waves, goal 1 can never latch (the agent won't re-send), and the run dies on the clock despite both tasks being done, with nothing telling the user why.

The strictly-ordered contract is fine, but the same drain also loses goal N+1's event when two completions land in one batch (physics stall, rosbridge burst on reconnect) — which defeats even the intended sequential semantics. Buffering unmatched skill events (or letting each goal's `SkillDone` see the batch and latch a pending flag it only *reports* once prior goals are done) would fix both.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -0,0 +1,483 @@
...
+
+    # -- evaluation (physics thread, after state gathering; no sim access) --
+
+    def tick(self, t: float, pose: tuple[float, float, float], centers: dict[str, tuple[float, float]]) -> dict:
+        """Advance the active challenge and return the state-stream block."""
+        with self._mutex:
+            challenge, events, self._events = self.active, self._events, []
+            if challenge is not None and self.state == "running":
+                state = WorldState(t=t, robot=pose, centers=centers)
+                self.elapsed_s = max(0.0, t - self.started_t)
```
theo-michel, line 476: **An external sim reset mid-run leaves a zombie run whose timer never fires.**

`/virtual_mars/reset` (or the observer `reset` op) calls `sim.reset()` directly: props re-park and `data.time` zeroes, but the engine is never told. It keeps `state="running"` with the old `started_t` — say 900 — so `max(0.0, t − started_t)` clamps `elapsed_s` to 0.0 for the next 900 sim-seconds. The panel shows a permanently-running 0:00 challenge over an empty scene; the goals can't pass (their props are parked) and the time limit can't fire either. Only a manual abort ends it.

Either route the reset ops through the engine (abort-on-reset), or treat `t < self.started_t` in `tick()` as evidence the world was reset out from under the run and fail/abort it — which would also double as the monotonicity guard for the gather/tick race flagged in `world_server.py`.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -0,0 +1,483 @@
...
+            return False
+        # Nothing is judged while the scene is being built. The world reset and
+        # the drops take the sim lock, which the physics thread keeps grabbing
+        # between ticks, so a tick lands in the middle of this -- and it must
+        # not see a challenge whose start time isn't known yet (elapsed_s
+        # against a stale started_t instantly "times out" a fresh run) or judge
+        # goal 0 against the world the last run left behind. So: deactivate,
+        # build the scene, then publish the whole run in one atomic step.
+        self._deactivate()
+        with self.sim_lock:
```
theo-michel: **Two concurrent `start()` calls interleave — the published run can be judged against the other challenge's world.**

`start()` is three separate critical sections (`_deactivate` under `_mutex`, scene build under `sim_lock`, publish under `_mutex`) with nothing held across them. Each observer connection gets its own command thread, so two tabs starting Shepherd and Rescue near-simultaneously can interleave: A drops the ball+dog, B's `reset()` re-parks them and drops the human, and if A's publish block runs last, `active=Shepherd` is judged against a world containing only Rescue's human — `Near("robot", "soccer_ball", …)` can never fire and the run dangles until timeout.

A single engine-level start lock (or serializing `start`/`abort` through one queue in the world server) closes it.

theo-michel (commented):
Reviewed the challenge layer. The engine's own race fix and the atomic-publish design are sound as far as they go, but there are still a few real holes in the concurrency story plus one crash path — inline comments below on the five that matter, a mobile UI regression, and one cheap hardening.

Minor cleanups, no inline comment needed:
- `load_challenges` duplicates `props.load_props` near-verbatim (~13 of ~17 lines; its own docstring says "the same shape as props.load_props") — worth a shared `load_sidecars(roots, attr, tag)` so the two contracts can't drift.
- `webapp/js/agent/main.js` line 86: the `overlay()` helper is dead now that its only call site was inlined — delete it or keep calling it.
- `.overlay-stack-top-left > .overlay` (app.css:643) applies the sim-justified 66% dark glass to the real-robot telemetry card too, diverging from the identical cards on teleop/collect; consider scoping the dark variant to sim or sharing a modifier with `.overlay-sim`, whose values it copies.

webapp/js/agent/main.js
```diff
@@ -45,13 +46,24 @@ function buildAgentView(root) {
...
+  // in the same corner position as before.
+  const cornerStack = document.createElement("div");
+  cornerStack.className = "overlay-stack-top-left";
+  root.append(cornerStack);
+  // Column order is DOM order: the challenge panel (sim only) first, then the
+  // telemetry card under it. The stack does the positioning, so the card
+  // carries no corner class of its own.
+  const challengePanel = typeof session.onChallenge === "function" ? createChallengePanel(cornerStack, session) : null;
+  const telemetryOverlay = document.createElement("div");
+  telemetryOverlay.className = "overlay";
```
theo-michel, line 60: **Mobile regression: the telemetry card is no longer hidden on phones.**

Dropping the corner class here orphans the mobile rule at `app.css:5976` (`.overlay-top-left { display: none }` — "The Robot/Link/Agent telemetry card costs too much of a small feed"), and no rule hides `.overlay-stack-top-left`. On a ≤640 px screen the Agent page now renders the telemetry card — plus, in sim, the fixed-width 252 px challenge panel at z-index 6 — over the full-width camera strip, while Teleop and Collect still hide theirs.

Adding `.overlay-stack-top-left { display: none }` to the same media block restores parity (or hide just the telemetry child if the challenge panel should survive on phones). Any future rule written against `.overlay-top-left` (safe-area insets, z-index) will likewise skip the agent page now that it's on a second corner system.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -0,0 +1,483 @@
...
+                with connect(self.url, open_timeout=5) as ws:
+                    ws.send(json.dumps({"op": "subscribe", "topic": self.TOPIC, "type": "std_msgs/String"}))
+                    if not announced:
+                        print(f"[challenges] skill events connected ({self.url})", flush=True)
+                        announced = True
+                    for message in ws:
+                        frame = json.loads(message)
+                        if frame.get("topic") == self.TOPIC:
+                            self.engine.post_event(json.loads(frame["msg"]["data"]))
+            except Exception:  # noqa: BLE001,S110 -- rosbridge down/restarting; retry
```
theo-michel, line 589: **Hardening: one bad frame costs 5 s of skill events.**

Both `json.loads` calls run inside the `for message in ws` loop with no per-message `try`, so a single malformed frame on `/brain/skill_status_update` — an open `std_msgs/String` bus anyone can `ros2 topic pub` onto — raises here, tears down the whole connection, and sleeps 5 s. Completions published in that window are gone (rosbridge doesn't replay), so a `SkillDone` goal the robot actually met silently never latches. Today's in-repo publishers always emit valid JSON, so this is latent — but a `try/except` around the per-message parse (keeping the connection) confines the damage to the one bad frame and costs two lines.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
karmanyaahm: Fixed in `817813fc`.

A single `_run_lock` now spans all three critical sections of `start()` — deactivate, build the scene, publish the run — so a second starter cannot begin until the first has published. The later start then simply wins: it aborts the run in progress and builds its scene on top, so the published run and the world always agree. `abort()` takes the same lock, or it could clear an `active` that the starting thread is about to overwrite and leave a scene on the floor with nothing judging it.

Lock order is `_run_lock -> sim_lock -> _mutex`, and no two are ever held together: `tick()` and `post_event()` take `_mutex` alone, and nothing acquires `_run_lock` while holding another, so there is no cycle. That ordering is now stated in the `ChallengeEngine` docstring so the next change has to keep it true.

Regression test in `sim/sandbox/test_driver_core.py::check_concurrent_start` races a second start from inside the first one's scene build: it asserts the run being built is never torn down (the teardown count stays at 1) and never overtaken, then that the winning run's props are exactly what the world contains. Against the pre-fix engine the teardown count reaches 2 before the first start has published.

One scope note: this closes the concurrent-start half only. The gather/tick race on `world_server.py:143` is a separate mechanism (a tick gathered *before* `start()` and delivered *after* it) and is still open.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -0,0 +1,456 @@
...
+            self.elapsed_s = 0.0
+            self._events.clear()
+            for goal in challenge.goals:
+                try:
+                    goal.predicate.reset()
+                except Exception:  # noqa: BLE001,S110 -- challenge bug; judged as-is
+                    pass
+            entry = self.progress.setdefault(challenge_id, {"attempts": 0, "passed": False, "best_time_s": None})
+            entry["attempts"] += 1
+        with self.sim_lock:
```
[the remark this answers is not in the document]
karmanyaahm: Fixed in `817813fc`.

A single `_run_lock` now spans all three critical sections of `start()` — deactivate, build the scene, publish the run — so a second starter cannot begin until the first has published. The later start then simply wins: it aborts the run in progress and builds its scene on top, so the published run and the world always agree. `abort()` takes the same lock, or it could clear an `active` that the starting thread is about to overwrite and leave a scene on the floor with nothing judging it.

Lock order is `_run_lock -> sim_lock -> _mutex`, and no two are ever held together: `tick()` and `post_event()` take `_mutex` alone, and nothing acquires `_run_lock` while holding another, so there is no cycle. That ordering is now stated in the `ChallengeEngine` docstring so the next change has to keep it true.

Regression test in `sim/sandbox/test_driver_core.py::check_concurrent_start` races a second start from inside the first one's scene build: it asserts the run being built is never torn down (the teardown count stays at 1) and never overtaken, then that the winning run's props are exactly what the world contains. Against the pre-fix engine the teardown count reaches 2 before the first start has published.

One scope note: this closes the concurrent-start half only. The gather/tick race on `world_server.py:143` is a separate mechanism (a tick gathered *before* `start()` and delivered *after* it) and is still open.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/world_server.py
karmanyaahm: Fixed in `757ad14a`.

`world_epoch` names which build of the world a snapshot came from. `start()` bumps it as the first statement inside its scene build; `publish_state()` reads it in the same lock hold that gathers `t`/`pose`/`centers`. Both sides hold the sim lock, so a snapshot can only ever carry the epoch of the world it actually describes. `tick()` judges when the snapshot's epoch matches the active run's, and otherwise renders without judging.

An epoch rather than the monotonic-`t` guard you offered as the alternative, for two reasons: a run with `reset_world=False` never rewinds the clock, so a time comparison silently does nothing exactly where it is still needed; and the epoch also fixes the stale-props symptom, which a timer guard does not. Comparing `active` identity would not work either — `Challenge` instances are reused out of `self.challenges`, so a restart of the same challenge is the same object.

Related: a skipped tick no longer drains `_events`, or a stale tick would swallow the new run's skill completions.

New regression test `check_stale_tick` in `sim/sandbox/test_driver_core.py`. It needs no threads — the gap is between two calls the test makes itself, so snapshot / start / stale-tick in that order reproduces it deterministically. Verified by removing only the epoch clause and re-running:

```
AssertionError: stale snapshot killed the fresh run:
{'id': 'probe', 'state': 'failed', 'reason': 'time limit', 'elapsed_s': 900.0, 'time_limit_s': 60.0, ...}
```

The other challenge tests still pass with the clause removed, so that test is specifically what catches this.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
karmanyaahm: Fixed in `1b8db7e9` — per-message `try/except` around both `json.loads` calls, so a malformed frame costs that frame and not the connection, and the 5s reconnect sleep with it. The bad frame is logged and skipped.

Since then the same reasoning has been applied one level up in `c5b5a6b8`: `publish_state` now wraps the whole `challenges.tick()` call, because it runs on the physics thread and `physics_loop` has no guard — so a predicate bug or a hand-edited `challenges.json` drops the challenge block for a frame instead of freezing the world for every viewer.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -0,0 +1,516 @@
...
+        # build the scene, then publish the whole run in one atomic step.
+        #
+        # _run_lock holds those three steps together against a SECOND starter.
+        # Each observer connection commands on its own thread, so two tabs
+        # starting different challenges could otherwise interleave -- A drops
+        # its props, B's reset re-parks them and drops its own, and whichever
+        # publishes last is judged against the other's world, with goals that
+        # can never fire. Serialized, the later start simply wins: it aborts
+        # the earlier run and builds its scene on top.
+        with self._run_lock:
```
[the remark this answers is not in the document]
karmanyaahm: yeah thats fine

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
karmanyaahm: yeah thats fine it should be strict order

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
karmanyaahm: Fixed in `0fa17751` — the second of your two options.

`tick()` now reads a backwards sim clock as evidence the world was rebuilt under the run, and fails it with reason `"the sim was reset"`. The panel renders that through the existing `Failed — {reason}` banner, so no wire or UI change was needed.

A clock check rather than routing the reset ops through the engine, because it catches all three paths: `/virtual_mars/reset`, the observer `reset` op, and `core.step()`'s NaN recovery — that last one fires inside the physics step itself, so nothing above it could call an abort hook, and an explicit hook would also make `core.py` aware of challenges. The margin is `CLOCK_REWIND_S = 0.5`, comfortably clear of the ~25ms of legitimately out-of-order `t` that two `publish_state` callers (physics thread and an observer command) can produce, and far below the whole-uptime drop a reset causes.

Failing rather than silently aborting was the call: a reset really did destroy the attempt's scene, and the user gets told why instead of watching the panel snap back to the roster with no explanation.

On your note that this could double as the monotonicity guard for the `world_server.py:143` race — it turned out cleaner to keep them separate. That race is fixed with a `world_epoch` stamped into the snapshot under the sim lock, because a run with `reset_world=False` never rewinds the clock, so a time-based guard would have silently done nothing there. The two checks now sit next to each other in `tick()`: epoch mismatch means "this snapshot is from another world, don't judge it", clock rewind means "this world was rebuilt under me, end the run".

New regression test `check_external_reset`. Verified by removing only the tripwire and re-running:

```
AssertionError: a reset left the run running at 0.0s
```

The rest of the suite still passes with it removed, so that test is specifically what catches this.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
karmanyaahm: Done in `02ff5c67` — your two paragraphs turned out to want different answers.

**Paragraph 2 (two completions in one batch) was a straight bug and is fixed.** `tick()` now cascades: after a goal latches, the same batch is offered to the next, so a batch that satisfies several goals in sequence advances all of them. That is still strictly ordered — goal N+1 is only judged once N is done, just within the same tick rather than waiting for the next one. Ticks only happen when the physics thread stepped, so a stall or a rosbridge reconnect really does hand one tick several completions, and dropping the later one when the robot did things in the right order was a hole in the contract rather than the contract.

**Paragraph 1 (the lifesupport story) stays as designed, but stops being silent.** A completion that arrives while an *earlier* goal is open is still discarded: goals are ordered and the robot is expected to do them in order. It is now logged with the skill and the goal still blocking —

```
[challenges] lifesupport: dropped completion(s) send_email -- goals are ordered and 'Wave to check if they respond' is still open
```

— so an author stops watching a run die on the clock with both tasks apparently done and nothing anywhere explaining why, which was the part of your report that really stung. `Challenge.goals` now states the rule too, since that is where an author meets it.

We considered deferring unmatched events and replaying them, and an `observe()` hook letting each `SkillDone` watch the stream and latch a pending flag. Both fix the lifesupport story, and both were rejected deliberately: replay has to carry the `WorldState` of the arrival tick or a guarded `SkillDone` silently changes meaning ("bark WHILE near the dog" would pass later, when the robot happens to be near), and the hook is a Predicate API every composite must forward for a case we would rather authors design around.

Known wart of the cascade: nothing consumes an event, so two consecutive identical goals — `[SkillDone("wave"), SkillDone("wave")]` for "wave twice" — both latch on one wave. No shipped challenge does that, and a consumption protocol was a lot of machinery for a case that does not exist yet.

New test `check_event_ordering` covers both halves. Verified by reverting only the cascade:

```
AssertionError: one batch did not advance both ordered goals:
[{'label': 'wave', 'done': True}, {'label': 'email', 'done': False}]
```

theo-michel (approved):
LGTM

PR #607, opened 2026-08-05
fix(brain): hold a task through skill registration instead of bouncing it

karmanyaahm:
## The symptom

`test_fake_cloud_loop`'s `test_loop_continues_after_completion` failed intermittently with:

```
AssertionError: Loop did not complete two primitives.
Types: [... 'primitive_failed', 'pose_image', 'primitive_activated',
        'primitive_feedback', 'primitive_completed', 'pose_image' x ~90]
```

Two things were wrong with that, before anything about the brain.

It is **the wrong diagnosis**: the loop did not stop. One of the two primitives came back `primitive_failed`, and the test counted only completions.

And it **discarded the evidence**. The reason lived in `payload["reason"]`; the assertion printed a list of message types, which cannot say why anything failed.

It also burned the full 45 s doing it. A failure is unrecoverable here — the fake cloud pops its queue on dispatch and never requeues (`fake_cloud._handle_perception:265`), so one failure left the test asking for two completions from a single remaining task. The assertion became unsatisfiable and the robot streamed perception until the deadline. That is the tail of ~90 `pose_image` frames, and it reads as a hang in the continuation path while being nothing of the sort.

## What the split found

Splitting the one test into its two actual claims, and asserting *with* the payload, produced the cause on the first failing run:

```
FAIL: test_dispatched_primitive_completes_rather_than_failing
AssertionError: 'primitive_failed' != 'primitive_completed'
 : innate-os/send_email failed instead of completing.
   reason='The robot dropped this task: skills were mid (re-)registration.'
```

- **`test_dispatched_primitive_completes_rather_than_failing`** (new) owns the failure and asserts with the reason the skills server attached.
- **`test_loop_continues_after_completion`** keeps its own subject and redispatches around a failure, bounded at `MAX_REDISPATCHES = 3`, so it measures continuation and fails for its own reasons. It passed in the failing run — the other half of the point.

## The cause

`vision_output.handle_message` answered any task landing inside a registration round trip with a terminal failure:

```python
if not self._state.primitives_registered:
    self._bounce_dropped_task(msg.payload, "skills were mid (re-)registration")
    return
```

Worth recording what this is *not*, because it is the plausible-looking answer: the skill catalog is never the gate. `catalog.reload_all` builds the new dicts outside the lock and swaps them inside it, so a lookup during a reload sees the complete old dict. The gate is the brain's own `primitives_registered` flag.

## The fix

A task arriving mid-registration is not one the robot *cannot* run — it is one the robot is not ready for **yet**, and the round trip is one the robot itself started. The cloud was being told a skill failed when it would have succeeded moments later.

It is now held and run once registration completes (`Orchestrator.handle_registered` → `replay_deferred`), for at most `DEFERRED_TASK_TIMEOUT_S`, polled from the agent loop, after which it bounces exactly as before.

That cap is load-bearing rather than defensive garnish: the cloud marks a primitive running when it *sends* it, and only a terminal message clears that, so silence would pin the agent on a tool call forever — precisely the failure `_bounce_dropped_task` was written to prevent. Latest-wins while holding (the cloud runs one task at a time), and a superseded task still gets its terminal message rather than being dropped silently.

## Evidence

| commit | integration runs | result |
|---|---|---|
| `cf294a07` (tests only) | 2 | **1 failure** (the trace above), 1 pass |
| `76f13a43` (+ fix) | 4 | **4 passes** |

Reproduced within two runs before the fix; four consecutive green after, via `workflow_dispatch`.

This is not proof. It is a race, absence of a flake is weak evidence, and four runs is a small sample. What changed is that the mechanism is now identified from a real trace instead of inferred from message ordering, and the fix addresses that mechanism directly. More dispatches before merge are cheap if you want a larger sample.

## Known limitation

`expire_deferred` is polled from `agent_loop`, which early-returns while the brain is inactive. If the brain goes inactive while a task is held, nothing expires it until the loop resumes — bounded in practice, but not by construction.

## Validation

- [x] `ruff check` + `ruff format --check` clean across `brain_client` (caught a late-binding closure in the retry loop; the predicate binds `seen_failures` explicitly).
- [x] Integration tests: 4/4 green on the fix commit, against a failure that reproduced within 2 runs before it.
- [ ] Not exercised on hardware; the race is in the brain, which is shared, so behavior should be identical.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/core/orchestrator.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/core/orchestrator.py
@@ -59,6 +59,7 @@ class Orchestrator:
 
         # Set via set_lifecycle() after construction (mutual cycle with BrainLifecycle).
         self.lifecycle = None
+        self.vision_output = None
 
         self._active_inputs_pub = active_inputs_pub
         self._memory_positions_pub = memory_positions_pub
@@ -73,6 +74,12 @@ class Orchestrator:
     def set_lifecycle(self, lifecycle) -> None:
         """Inject BrainLifecycle, which can only be built after this object (mutual cycle)."""
         self.lifecycle = lifecycle
+
+    def set_vision_output(self, vision_output) -> None:
+        """Inject VisionOutputHandler, built after this object. Registration
+        completing is what releases a task it held (see its
+        _defer_until_registered)."""
+        self.vision_output = vision_output
 
     # ================= perception loop =================
     def check_lidar_health(self) -> None:
@@ -111,6 +118,8 @@ class Orchestrator:
             # at 10 Hz, so keep this at debug to avoid flooding the log. Matches
             # the "Brain not active" skip above.
             self._logger.debug("[BrainClient] Primitives not registered. Skipping agent_loop.")
+            if self.vision_output is not None:
+                self.vision_output.expire_deferred()
             return
         if not (self._state.ready_for_image and self._camera.has_image):
             return
@@ -283,6 +292,8 @@ class Orchestrator:
             self._logger.error("Failed to register primitives and/or directive with server")
             return
         self._state.primitives_registered = True
+        if self.vision_output is not None:
+            self.vision_output.replay_deferred()
         self._logger.info(
             f"Successfully registered {msg.payload.get('count', 0)} primitives and directive: "
             f"{msg.payload.get('directive_registered', False)}"
--- a/ros2_ws/src/brain/brain_client/brain_client/core/vision_output.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/core/vision_output.py
@@ -11,12 +11,21 @@ from __future__ import annotations
 
 import json
 import math
+import threading
+import time
 import traceback
 
 from brain_client.perception import pose as pose_math
 from brain_client.transport.messages import VisionAgentOutput
 
 _NAV_TO_POSITION = "innate-os/navigate_to_position"
+
+# How long a task may wait for a registration round trip before the cloud is
+# told the robot dropped it. Registration is a single request/ack the robot
+# itself initiates, so this is generous; the cap exists only so a registration
+# that never completes cannot pin the agent on a tool call forever -- the exact
+# failure _bounce_dropped_task was written to prevent.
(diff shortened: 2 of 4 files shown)


karmanyaahm:
## The split did its job on the first failing run

```
FAIL: test_dispatched_primitive_completes_rather_than_failing
AssertionError: 'primitive_failed' != 'primitive_completed'
 : innate-os/send_email failed instead of completing.
   reason='The robot dropped this task: skills were mid (re-)registration.'
```

So it **is** the registration race, and my earlier reading was wrong in a specific way worth recording: `catalog.reload_all` really does swap atomically, so the catalog was never the gate. The gate is the brain's own `primitives_registered` flag — `vision_output.handle_message` answered any task landing inside a registration round trip with a terminal failure.

Note the old test could never have told us this. It printed a list of message types and reported "the loop did not complete two primitives"; the reason lived in a payload it discarded.

## The fix (`76f13a43`)

A task arriving mid-registration is not one the robot *cannot* run — it is one the robot is not ready for **yet**, and the round trip is one the robot itself started. It is now held and run when registration completes (`Orchestrator.handle_registered` → `replay_deferred`), for at most `DEFERRED_TASK_TIMEOUT_S`, polled from the agent loop, after which it bounces exactly as before. That cap matters: the cloud marks a primitive running when it *sends* it and only a terminal message clears that, so silence would pin the agent on a tool call forever — the very failure `_bounce_dropped_task` was written to prevent. Latest-wins while holding, and a superseded task still gets its terminal message.

## Runs

| commit | runs | result |
|---|---|---|
| `cf294a07` (tests only) | 2 | **1 failure** (the trace above), 1 pass |
| `76f13a43` (+ fix) | 4 | **4 passes** |

Four consecutive green on the fix, against a failure that reproduced within two runs before it. Not proof — it is a race, and absence of a flake is weak evidence — but the mechanism is now identified from a real trace rather than inferred, and the fix addresses that mechanism directly.

`test_loop_continues_after_completion` passed in the failing run too, which is the other half of the point: the continuation path is no longer collateral damage when a primitive fails.

karmanyaahm:
closing this since the switch to local agent will almost certainly override this very soon

PR #609, opened 2026-08-05
Barge-in detection: let humans interrupt the robot mid-sentence

DavidDobas:
## What this does

The robot can now be interrupted mid-sentence. Say something while it's talking and it stops, drops the rest of that reply, and listens.

Standard AEC is not an option here: the mic lives in the arm's gripper camera, centimetres from the speaker, so the robot's own voice arrives ~33 dB above the room floor and (before this PR) clipped the ADC outright. Instead this exploits the one advantage we have — **the robot knows exactly what it is saying** — and looks for speech the TTS reference cannot explain.

```
TTS PCM ──tee──> /tts/ref_audio ─┐
   │                             ├─> BargeInDetector ──> /barge_in ──> stop_current()
   └──> aplay (speaker)          │      align → predict echo → score residual
                          mic ───┘
```

## How it works

- **`transport/tts.py`** tees the exact PCM being piped to `aplay` onto `/tts/ref_audio` (WAV header stripped, pipe shrunk to 16 KB so the reference tracks playback), and gains `stop_current()` — the previously missing ability to kill speech mid-utterance.
- **`audio/barge_in.py`** (pure numpy, runs in the existing mic thread on chunks ducking used to discard): aligns reference to mic by envelope cross-correlation (the lag varies 0.3–1.4 s per utterance with TTS time-to-first-byte), predicts the echo per mel band via adaptive gains plus a measured reverb-decay envelope, and triggers when the loudest bands exceed the prediction for long enough.
- **`audio/echo_model.py`** optionally layers a learned echo predictor over the adaptive gains — a small conv net trained offline (GPU) from robot-recorded TTS/mic pairs, exported to npz and run as precomputed GEMMs (**2.8 ms/frame, no torch on the robot**). It contributes via elementwise max, never below the adaptive estimate — alone it false-fires when the camera's AGC drifts.
- **`scripts/audio/`** makes the model reproducible: `collect_echo_pairs.py` drives the production TTS path while recording aligned pairs; `train_echo_model.py` fits and exports. Models are per-robot/per-room and intentionally not committed; without one the detector runs on adaptive gains alone.

## Measured on MARS, and what it changed

Every constant here came from measurement, not intuition:

| Observation | Consequence |
|---|---|
| Mic clips through TTS peaks at 100 % capture gain | Gain now ducks to 50 % **only while speaking** (`barge_in_tts_mic_percent`), restored instantly after — the weak gripper mic keeps full STT sensitivity |
| Echo is ~40 dB weaker at 4–8 kHz | Scoring covers the full range, not just 200–4000 Hz. This was the single biggest sensitivity win; a user noticed interruptions "worked better at high frequencies" before the data confirmed it |
| Servo motion measures −23 dBFS at the gripper mic vs a −50 floor | `/joint_states` gates detection during motion (+350 ms) |
| Room reverb decays 0.87/frame (median of recorded tails) | Predictor follows the ring-out instead of firing on it (`barge_in_reverb_decay`) |

## Fixes found along the way

Three of these are pre-existing bugs this work surfaced, worth reviewing on their own:

- **The robot answered itself.** Ducking released the instant `aplay` exited, so the reverb tail and inter-utterance gaps reached STT, which hallucinated full sentences from fragments and fed them back as user chat. Fixed with a 400 ms hold-off plus a self-echo transcript filter (the TTS text rides along on the reference stream).
- **A dropped websocket killed the mic thread.** The OpenAI reconnect loop tore down and respawned the audio consumer, silently disabling capture mid-utterance. It's now one persistent thread; reconnects only swap the client.
- **Odd-length HTTP chunks desynced the reference.** Cartesia's chunk boundaries aren't sample-aligned; a single odd byte made `audioop.ratecv` reject every subsequent chunk for that utterance.
- **Interrupting one sentence wasn't enough.** Replies stream as several utterances, so the tail spoke seconds later and the interruption felt ignored. A barge-in now holds all speech for 5 s, released early once the user's words reach the brain.

## Testing

- 8 unit tests (`test/test_barge_in.py`) over synthetic acoustics: echo-only silence, loud and quiet interrupters, TTS-pause interruptions, clipping, resampling, odd-byte chunks, and an under-predicting model.
- Live on the robot with an automated harness: ElevenLabs voices played from a laptop across the room as spatially realistic interrupters, with per-round ground truth. Best round detected 8/8 interrupter clips across voices and volumes.
- Confirmed end-to-end in real conversation — human speech stops the robot mid-sentence.

## Notes for review

- Defaults are the values validated on MARS. Expect per-robot retuning; `barge_in_debug_dir` writes one npz trace per utterance for exactly that, and every knob is live in `settings.yaml` without a rebuild.
- The synthetic test harness pins its own threshold rather than riding the shipped default — its convolved-noise reverb leaves hotter residuals than real speech echo, so field values would fail it for reasons that don't exist in a real room.
- **The precision/recall trade-off is not fully settled.** Tuning it further needs daytime labelled sessions in a known-quiet room; my overnight negative trials can't distinguish a false positive from a real sound in the flat.
- `barge_in_flush_tail` (replaying post-trigger audio to STT) is implemented but unvalidated, and off by default.
- Unrelated bug found while testing: after a full stack restart the input manager sometimes comes up with dead zenoh subscriptions, which silently breaks TTS ducking. A per-node restart heals it. Tracked separately, not addressed here.

### DIFF
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -152,7 +152,7 @@
 #     openai_realtime_model: "gpt-4o-realtime-preview"
 #     openai_transcribe_model: "gpt-4o-mini-transcribe"
 #     barge_in_enabled: true        # stop TTS when a human talks over the robot
-#     barge_in_threshold_db: 6.0    # residual-over-echo excess to count a frame as speech
+#     barge_in_threshold_db: 10.0   # residual-over-echo excess (top-5 bands) to count a frame as speech
 #     barge_in_min_ms: 250          # cumulative overlap-speech evidence before triggering
 #     barge_in_flush_tail: false    # replay post-trigger mic audio to STT after TTS stops
 #     barge_in_model_path: ""       # npz from scripts/audio/train_echo_model.py; empty = adaptive gains
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -151,6 +151,10 @@
 #   ros__parameters:
 #     openai_realtime_model: "gpt-4o-realtime-preview"
 #     openai_transcribe_model: "gpt-4o-mini-transcribe"
+#     barge_in_enabled: true        # stop TTS when a human talks over the robot
+#     barge_in_threshold_db: 6.0    # residual-over-echo excess to count a frame as speech
+#     barge_in_min_ms: 250          # sustained speech required before triggering
+#     barge_in_flush_tail: false    # replay post-trigger mic audio to STT after TTS stops
 
 # ── Vision-language navigation (UniNavid) ─────────────────────────────
 # (The cloud endpoint URL is connectivity and lives in .env as UNINAVID_WS_URL.)
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -156,6 +156,7 @@
 #     barge_in_min_ms: 250          # cumulative overlap-speech evidence before triggering
 #     barge_in_flush_tail: false    # replay post-trigger mic audio to STT after TTS stops
 #     barge_in_model_path: ""       # npz from scripts/audio/train_echo_model.py; empty = adaptive gains
+#     barge_in_tts_mic_percent: 50  # capture gain while the robot speaks (100 = never duck the mic gain)
 
 # ── Vision-language navigation (UniNavid) ─────────────────────────────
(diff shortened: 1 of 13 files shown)


theo-michel:
Reopening under the correct author account (opened by mistake from a shared machine's default `gh` login; commits were always authored by Theo Michel).

PR #613, opened 2026-08-06
docs(sim): move notebook expected-output images out of the ipynb

DavidDobas:
## Why

`sim/sandbox/virtual_mars_demo.ipynb` was 722 KB — matplotlib PNGs were stored as base64 cell outputs inside the notebook JSON, making it heavy to diff and search.

## What

- Extracted the four output images (RGB cameras, depth, occupancy grid, lidar) to `sim/sandbox/virtual_mars_demo_assets/` as normal git-tracked PNGs.
- Added a collapsible **Expected output** markdown cell after each demo cell that embeds the corresponding image via a relative path — GitHub renders these, so readers still see the expected results before running anything.
- Cleared all execution outputs and counts.

The notebook drops from **722 KB to 10 KB**, and re-running + clearing outputs no longer destroys the documentation images.

## Notes

`rgb_cameras.png` is 461 KB (natural-image render, PNG compresses it poorly), but it now lives as a regular binary file instead of bloating the notebook JSON.


theo-michel (approved):
LGTM (after greptile comment addressing)

PR #614, opened 2026-08-06
Add CI file-size guard via innate-inc/git-file-size-limit

karmanyaahm:
## What

Adds a CI check that stops large files from being committed accidentally. Any file added or changed in a pull request that exceeds **100 KiB** fails the check. Files that are intentionally large are exempted by listing them in a root **`.sizelimitignore`** (gitignore-style patterns) — that edit shows up in the diff and gets reviewed like any other change, which is the "explicit approval" gate.

## Why

The motivating case lives in this repo: `sim/sandbox/virtual_mars_demo.ipynb` is ~739 KiB because it was committed **with saved outputs**. It's smaller than legitimate binary assets (meshes, readme GIFs) yet clearly shouldn't be in git.

## How it works

Runs [**`innate-inc/git-file-size-limit`**](https://github.com/innate-inc/git-file-size-limit) — a standalone, open-source GitHub Action built for this, so the checker lives in one place and can be reused by any repo (not just this org) instead of being copy-pasted per repo.

- Diffs the PR against its merge-base and checks only **changed** files (three-dot diff, matching GitHub's "Files changed" tab). Pre-existing large files (meshes, readme assets) are never flagged unless a PR touches them.
- Exemptions in `.sizelimitignore` are matched with **real gitignore syntax** via `git check-ignore --no-index` — not an approximation. (`--no-index` matters: without it, git refuses to flag already-tracked files as ignored, which would silently defeat every exemption since changed files are always tracked. Verified this before shipping.)
- `.sizelimitignore` is seeded with `uv.lock` and `package-lock.json`.

Currently pinned to `@main`; will move to a released tag (`@v1`) once one exists on the action repo.

## Testing

The action ships its own test suite (14 assertions covering violations, exact + glob exemptions, edits, deletions, the pass path, `fail_on_violation=false`, and size-unit parsing) plus an end-to-end composite-action smoke test, all passing in its own CI.

https://claude.ai/code/session_01LF1W2HQY2X8Say1QqQqCpn


theo-michel (approved):
LGTM

PR #616, opened 2026-08-07
fix(brain): stop go_to_point approaches short of what the lidar sees

theo-michel:
## Summary

- `go_to_point_in_view` drove the gripper into a wall on R7-44 (2026-08-06): the 0.35 m standoff was measured at `base_link`, but the outstretched arm reaches up to 0.49 m ahead of it and nav2's goal checker adds ±0.15 m — the goal itself put the gripper past the pointed-at spot.
- Replace the constant with a lidar **corridor guard**: aim the original 0.35 m short (spot in the gripper's working range), then clamp the travel so `front_extent + goal tolerance + clearance` stays short of every scan return in a robot-width corridor toward the spot. Close on open floor, margin only where a wall actually is.
- Front extent and half-width come live from `/footprint` (the arm-inclusive hull nav2's costmaps already consume); `ScanHealthMonitor` additionally serves the latest returns in `base_link` (new `x_laser` param, URDF `base_laser` offset). Self-hits from the arm crossing the scan plane are filtered.
- Degradation is one-sided: stale/missing scan or footprint falls back to a blind worst-case standoff — stops early, never over-drives.

## Validation

- [ ] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up` — *not run: no sim changes; in sim the missing `/footprint`/scan feeds take the conservative fallback path, covered by unit tests.*
- [ ] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json` — *n/a.*

`pytest` brain_client: 117 passed (4 new: open-floor approach, wall clamp, self-hit/side/behind filtering, no-scan fallback). Hardware validation on R7-44 against the crash wall, arm outstretched (live front extent 0.43 m):

| scenario | result |
|---|---|
| wall spot 0.75 m ahead | travel clamped to 0.06 m; nav holds at the guard boundary, ~0.19 m real gripper-to-wall, repeated pointing cannot creep in |
| open-floor spot 1.2 m | full travel, spot ends 0.35 m from base_link |
| replay of the crash geometry through the old constant | gripper tip 20 cm past the wall face — reproduces the collision |

Field note recorded on `CLEARANCE_M`: a tape check found the lidar reading the crash wall ~10 cm long; widen the margin before trusting the guard on tighter work.


axelpey:
Closing the stranded stacked PR. Its parent architecture has merged, the current diff is no longer reviewable, and the lidar guard retains unresolved frame and Nav2-tolerance safety issues. Re-cut the fix from current main.

PR #617, opened 2026-08-07
feat(brain): local spatial memory — recorded on tour, visible on the map, recallable as a skill, Gemini-cached

theo-michel:
## Summary

The complete local spatial memory for MARS, in the spirit of [Mobility VLA](https://arxiv.org/abs/2407.07775): while driving well-localized the robot **remembers what it sees**; a VLM then **recalls** the remembered frame that answers a request ("go to the kitchen", "find my airpods", "I am hungry" → the counter with the bananas) and the agent navigates to its pose. Stacked on #565 (`feat/local-gemini-brain`) — this is the "memory system … to be rebuilt locally later" from that PR's summary. Built in themed layers (a merge duplicated some early commits; the sections below are the review map).

### 1 · The memory core (`feat(brain): local spatial memory…` + the 12 h TTL bump)

- **Recording** (`brain_client/memory/`, new): an always-on `MemoryRecorder` (webapp teleop tours build memory too, not just brain-driven runs) ticks at 1 Hz and records only what a *well-localized* robot saw: AMCL covariance must hold below the webapp's "confident" thresholds (`max(var_x, var_y) ≤ 0.10 m²`, `var_yaw ≤ 0.18 rad²`) for **3 s straight**, in `navigation` mode, with a fresh frame and the head not pitched at the floor.
- **Storage**: a plain JSON index + one JPEG per memory under `data/spatial_memory/<map>/` — thread-safe, atomically written, persistent across restarts. (The cloud era's networkx pose graph is gone: it built edges retrieval never used.) Memories key off `/nav/current_map`, and the index stores a SHA-256 of the map's `.pgm` — **re-mapping under the same name wipes memories whose coordinate frame died**.
- **Admission** is a small pure policy (`memory/selection.py`): a viewpoint earns a slot only if no kept frame is both near (<1 m) *and* similarly oriented (<100°); revisiting a known viewpoint after 30 min refreshes its image in place (memories age — the banana moves); at the 50-frame cap the older half of the closest pair is evicted.
- **Retrieval** (`brain/memory_search.py`, new): every remembered frame goes to Gemini labeled with id, capture time, and map pose; a structured verdict (`{found, frame, explanation}`) picks the best match. Frames ride in an explicit **Gemini context cache** (12 h TTL — ~$0.02/h storage for a ~20 k-token blob), pre-warmed in the background once recording has been quiet 30 s, so recall re-uploads nothing and restarts re-cache on their own.

### 2 · Made visible (`feat(webapp): the spatial memory made visible…`)

- `/brain/memory_positions` carries `id` + `stamp` per memory and a `cache` health field; every finished search mirrors its verdict onto a latched `/brain/memory_search`; each memory's JPEG is served same-origin at `GET /memory/image` (`webapp/proxy`, same fenced-route pattern as `/map/preview`, `?v=` stamp as cache-buster) — images never cross rosbridge.
- **"Memories" map layer** (Nav page + teleop PiP): one violet view cone + dot per viewpoint, opacity fading with age; a white ring blooms the instant a memory is recorded — you drive the tour and watch the robot learn. Hover → the remembered image; click → a pinned card with a **Go here** button (existing NavigateToPose path).
- **The recall moment**: on a fresh verdict the matched memory breathes with pulse rings, a dashed thread drifts robot→memory, and a card floats in with the query, the model's explanation, the frame, and `1.2 s · cached ⚡`. Stale latched verdicts (page reloads) deliberately show nothing.
- **"Memories" sidebar panel**: count, recall-cache chip (*instant recall ready / warming… / uncached*), and a reel of remembered views newest-first (hover lights the cone; click centres the map). All canvas animation runs on a rAF ticker that only spins while something animates.

### 3 · Made callable (`feat(skills): recall as a capability…`)

The capability-server pattern the codebase already uses twice (Nav2 behind `navigate_to_position`, `arm_sdk_server`): state and secrets live in a server; skills are thin, cancel-aware orchestrators.

- **`SearchMemory` action** (`brain_messages`) served by `brain/search_server.py` *inside the brain process beside `MemorySearch`* — cache, transport, credentials never leave it; own node + spin thread so a multi-second search can't stall a turn; cancels rejected by design (a caller that stops waiting abandons the goal; the verdict is dropped).
- **Result images, typed**: `SkillOutput` grows `image=jpeg_bytes`, `ExecuteSkill.Result` grows `image_b64`, the runner decodes it into the completion event — any skill can now *show* the agent its evidence (mirrors the feedback lane's `image_b64` precedent; additive msg change).
- **SDK accessor**: `memory: SpatialMemory` injects like `mobility: Mobility`; `begin(query)` returns a reader for `self.wait_for` (cancel-aware, no polling), concluding in a typed `RecallVerdict`.
- **The skill** (`workspace/innate_skills/search_memory.py`, ~15 lines) is in the basic + demo directives. The brain's built-in tool is removed — the declaration comes from the skill's docstring. `verdict_text()` is the single source of the model-facing sentence for every door the search comes through.

### 4 · Made cheap (`feat(brain): files tier…`)

The cache blob used to carry the frame *bytes*, so any memory change meant re-uploading all ~50 frames (~3 MB). Three invariants now hold, each degrading without changing answers:

- **Bytes cross once, at record time** (`brain/frame_files.py`): a background chore on the recorder's tick uploads each JPEG to the Gemini Files API; requests reference it as `fileData`. The registry is a `files.json` sidecar — restarts re-upload nothing — and entries are trusted only while their stamp matches the live memory. Server expiry 48 h; re-upload past 40 h, references stop at 47 h.
- **A search is one round trip and never waits for maintenance**: cache builds live only in `warm()`'s thread (which no longer holds the search lock). Fresh cache → question only; **stale cache → the stale `cachedContent` + a delta** (new/re-captured frames with supersede notes, retire notes for evicted ids); no cache → all frames as references; a not-yet-uploaded frame rides `inlineData`.
- **Clean degradation**: no upload endpoint → permanent latch, frames inline; no cache endpoint → reference-only requests; both → byte-identical to the plain inline behavior. Frames are labeled by **stable store id** (not position), so delta notes and verdicts survive additions/refreshes/evictions; the cache itself builds from `fileData`, making rebuilds reference-only POSTs.

### 5 · Made agnostic (`refactor(brain): the agent forgets memory exists`)

An earlier revision dispatched recall in-process and slot-free so the model could search while driving. That kept memory-specific machinery in the agent (a hardcoded skill id in tool building *and* dispatch), which taxes exactly the thing this system should invite: swapping the retrieval mechanism. The refactor (17 insertions, 161 deletions) makes `search_memory` an **ordinary blocking skill** — it occupies the slot for the seconds a cached search takes, and its verdict returns as a normal completion event carrying the matched frame.

- `agent.py` now has **zero memory references**: no skill-id special cases, no MEMORY event kind, no memory collaborator. The system-prompt paragraph moved into the skill's docstring — the same text the model reads as the tool description, now traveling with the skill.
- The brain *process* still hosts the capability (recorder, store, `/brain/search_memory` server) the way it hosts the camera; the reasoning loop doesn't know it exists.
- **Experimenting with a different retriever = editing one workspace file.** A new skill can call a different server, a different model, or no cache at all — the brain needs no changes.
- Traded away: recall isn't offered while another skill runs (stop it or wait, like any skill). The natural flow — search first, then navigate to the match — is unaffected.

## Validation

  - **180 brain tests pass** (`test_spatial_memory.py` reads as the spec: admission policy, store round-trip/fingerprint-wipe/eviction, the sustained-confidence gate, the full cache lifecycle — a search never builds the cache, a stale cache serves with a delta, warm rebuilds+deletes — and the files tier: upload-once, registry survives restart, aging-upload refresh, expired-reference inline fallback, the 404 latch vs transient retry, cache-from-references, mixed reference+inline requests, delta shape, **search-never-waits** while an upload hangs, id-stable labels/verdicts. `test_memory_skill.py`: action mapping, the accessor's unblock-on-every-failure-path contract, result-image plumbing. `test_local_brain.py`: skill completion events carry result images; the agent itself has no memory tests left to write — it has no memory code.)
  - **Proxy: 21 tests pass** (`/memory/image` serve + traversal fences). **Webapp**: `node tests/memories.test.js` + existing node tests pass; `tsc` reports zero errors in touched files (the repo's 47 pre-existing elsewhere unchanged).
  - `ruff` clean; `basedpyright brain_client/` at 13 errors vs the pre-stack base's 14 (a stale-install import got fixed by the `brain_messages` rebuild; nothing new, and every new module checks clean).
  - `brain_messages` rebuilt locally (new `SearchMemory` action + `Result.image_b64`), so suites run against real generated types. Nodes need the usual full rebuild to agree on the new type hash.
  - **Not yet driven live** — see the drive-test list below.
- [ ] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up`. — *No sim-specific changes: same topics; the layer stays empty until memories exist; the brain constructs the search server only when it has Gemini access.*
- [ ] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`. — *No asset changes.*

## What only a live drive test can answer

Tour SmallRoom on the Nav page (chip green, rings bloom), then ask the brain to find something, watching the node log:

1. **Proxy passthroughs**: does the service proxy forward `cachedContents` and `/upload/v1beta/files`? The proxy client also cannot attach the raw-upload header yet, so proxied robots likely log `backend has no file-upload endpoint` and latch to inline (direct `GEMINI_API_KEY` setups should work as-is) — a server-side proxy change would enable both tiers.
2. Whether `cachedContents` ingests `fileData` content at creation (cache outlives file expiry) — the 4xx fallback covers if not.
3. Retrieval quality on the four use-case shapes: "go to the kitchen", "find my airpods", "I'm hungry", "exit the room" — first-search vs cached-search latency shows in the recall card.

## Notes for reviewers

- **What's consciously traded** by recall-as-a-skill: the tool declaration is directive-gated (basic + demo include it), an empty memory rejects with an outcome message rather than a hidden tool, and recall isn't offered while another skill runs (section 5 — the price of a retrieval-agnostic brain).
- A search's empty-store answer comes from the search itself now (an honest "no memories yet" verdict), not an agent-side gate.
- Memory hue `#b48cff` (violet) is new to the map palette on purpose; mapping mode forces the layer off (marks are frames of the *finished* map) and restores it after, like the costmap chip.
- Cache cost: ~50 frames ≈ 15–50 k cached tokens; 12 h of storage ≈ $0.24 — cheap insurance against cold starts (TTL bump commit).

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/README.md
+++ b/ros2_ws/src/brain/brain_client/brain_client/README.md
@@ -6,9 +6,10 @@ code, start from what it *does*:
 | Folder | What lives here |
 |---|---|
 | `nodes/` | The runnable ROS entry points (the only files with `main()`). Thin composition roots — they build collaborators, wire them, and spin. No behaviour. |
-| `brain/` | The local agent loop: `agent` (look → think → act as one cancellable coroutine), `loop` (the dedicated-thread asyncio runtime it runs on), `context` (bounded Gemini conversation), `tools` + `transport` (declarations and the wire), pure `grounding` (pointed pixel → floor target), and the system `prompt`. |
+| `brain/` | The local agent loop: `agent` (look → think → act as one cancellable coroutine), `loop` (the dedicated-thread asyncio runtime it runs on), `context` (bounded Gemini conversation), `tools` + `transport` (declarations and the wire), pure `grounding` (pointed pixel → floor target), `memory_search` (recall over the spatial memory, context-cached), and the system `prompt`. |
 | `core/` | The activate/deactivate/reset state machine and directive switching (`lifecycle`), typed `config`, and the shared `state`. |
 | `perception/` | Turning sensors into what the agent sees: `camera`, `pose`/`pose_tracking`, `scan_health`, `gaze`. |
+| `memory/` | Persistent per-map spatial memory: `store` (JSON index + JPEGs on disk), pure `selection` (which viewpoints earn a slot), `recorder` (the always-on ROS adapter that captures them). |
 | `skills/` | The skill system: `registry`, `roster` (available + directive-active sets), `runner` (action lifecycle), `loader`, `hot_reload`, and the public `types` SDK base classes. |
 | `agents/` | Directives/behaviours: `loader`, `initializer`, and the public `types` SDK base class. |
 | `inputs/` | Input-device subsystem and its public `types` SDK base class. |
@@ -21,7 +22,7 @@ code, start from what it *does*:
 **1. Dependency direction is one-way.**
 
 ```
-nodes  ->  {brain, core}  ->  {perception, skills, agents, inputs, transport, robot}  ->  common
+nodes  ->  {brain, core}  ->  {perception, memory, skills, agents, inputs, transport, robot}  ->  common
 ```
 
 A module never imports "upward" (e.g. `perception` never imports `core`).
--- a/ros2_ws/src/brain/brain_client/brain_client/README.md
+++ b/ros2_ws/src/brain/brain_client/brain_client/README.md
@@ -6,7 +6,7 @@ code, start from what it *does*:
 | Folder | What lives here |
 |---|---|
 | `nodes/` | The runnable ROS entry points (the only files with `main()`). Thin composition roots — they build collaborators, wire them, and spin. No behaviour. |
-| `brain/` | The local agent loop: `agent` (look → think → act as one cancellable coroutine), `loop` (the dedicated-thread asyncio runtime it runs on), `context` (bounded Gemini conversation), `tools` + `transport` (declarations and the wire), pure `grounding` (pointed pixel → floor target), `memory_search` (recall over the spatial memory, context-cached), and the system `prompt`. |
+| `brain/` | The local agent loop: `agent` (look → think → act as one cancellable coroutine), `loop` (the dedicated-thread asyncio runtime it runs on), `context` (bounded Gemini conversation), `tools` + `transport` (declarations and the wire), pure `grounding` (pointed pixel → floor target), `memory_search` (recall over the spatial memory, context-cached) + `search_server` (the `/brain/search_memory` action skills call), and the system `prompt`. |
 | `core/` | The activate/deactivate/reset state machine and directive switching (`lifecycle`), typed `config`, and the shared `state`. |
 | `perception/` | Turning sensors into what the agent sees: `camera`, `pose`/`pose_tracking`, `scan_health`, `gaze`. |
 | `memory/` | Persistent per-map spatial memory: `store` (JSON index + JPEGs on disk), pure `selection` (which viewpoints earn a slot), `recorder` (the always-on ROS adapter that captures them). |
--- a/ros2_ws/src/brain/brain_client/brain_client/brain/agent.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/brain/agent.py
@@ -31,15 +31,9 @@ from typing import TYPE_CHECKING
 from brain_client.brain import grounding
 from brain_client.brain.context import Decision, GeminiContext, ToolCall
 from brain_client.brain.loop import LoopThread
+from brain_client.brain.memory_search import verdict_text
 from brain_client.brain.prompt import build_system_prompt
-from brain_client.brain.tools import (
-    GO_TO_POINT_IN_VIEW,
-    SEARCH_MEMORY,
-    STOP_SKILL,
-    WAIT,
-    assign_tool_names,
-    build_tools,
-)
+from brain_client.brain.tools import GO_TO_POINT_IN_VIEW, STOP_SKILL, WAIT, assign_tool_names, build_tools
 from brain_client.brain.transport import pick_transport
 from brain_client.brain.utils import (
     Event,
@@ -60,7 +54,7 @@ if TYPE_CHECKING:
 
     from rclpy.node import Node
 
-    from brain_client.brain.memory_search import MemorySearch
+    from brain_client.brain.memory_search import MemorySearch, SearchVerdict
     from brain_client.core.config import BrainConfig
     from brain_client.core.state import BrainState, RunningSkill
     from brain_client.perception.camera import CameraCapture
@@ -74,6 +68,7 @@ if TYPE_CHECKING:
     from innate_proxy import ProxyClient
 
 _NAV_TO_POSITION = "innate-os/navigate_to_position"
+_SEARCH_MEMORY = "innate-os/search_memory"
 _FRESH_FRAME_SEC = 3.0  # an older camera frame means the feed is broken; don't think blind
 _MAX_EVENTS_QUEUED = 30  # the oldest beyond this are dropped as stale stimuli
 _MAX_EVENT_IMAGES = 4  # newest event images sent per turn; older ones arrive as text only
@@ -448,12 +443,7 @@ class BrainAgent:
         # the name the model calls always resolves to the skill it was declared for.
         named = assign_tool_names(skills)
         self._tool_map = {name: meta["id"] for name, meta in named}
-        return build_tools(
-            named,
-            None,
-            can_go_to_point_in_view=_NAV_TO_POSITION in active_ids,
-            can_search_memory=self._memory is not None and self._memory.frame_count > 0,
-        )
+        return build_tools(named, None, can_go_to_point_in_view=_NAV_TO_POSITION in active_ids)
 
     # ================= act =================
     def _act(self, decision: Decision, speaker: SpeechStreamer, context: GeminiContext) -> list[tuple[ToolCall, str]]:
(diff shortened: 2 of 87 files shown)


karmanyaahm:
On the ROS action fields — should the image be `sensor_msgs/CompressedImage` rather than a base64 `string`?

`SearchMemory.action` and `ExecuteSkill.action` (both result and feedback) declare the image as a `string`:

```
# SearchMemory.action (result)
string image_b64    # the matched remembered frame ("" when none)
```
```
# ExecuteSkill.action (result + feedback)
string image_b64     # optional evidence image the skill attached to its result
```

In rclpy a `string` field is a UTF-8 `str`, so raw JPEG bytes can't go in one at all — every producer base64-encodes and every consumer decodes:

| encode | decode |
|---|---|
| `brain/search_server.py:87` | `robot/spatial_memory.py:125` |
| `nodes/skills_server.py:419` | `skills/runner.py:356` |
| `nodes/skills_server.py:387` (feedback) | `skills/runner.py:238` |

That's ~33% size inflation plus an encode and a decode on every hop, for what should be an opaque byte payload. [`sensor_msgs/msg/CompressedImage`](https://docs.ros.org/en/humble/p/sensor_msgs/msg/CompressedImage.html) is the standard type for exactly this: a `header`, a `format` string, and `uint8[] data`.

### This repo already does it the other way, in the browser

`mars_msgs/action/RunStereoCalibration.action:26` carries `sensor_msgs/CompressedImage[] images` **in action feedback**, and the webapp consumes it over rosbridge — `webapp/js/calibration/main.js:260` (`imageDataUrl`) turns `{format, data}` into a `data:` URL for an ``. So the browser-facing case for `CompressedImage` in an action isn't hypothetical here; it's shipping.

### The browser argument doesn't cut the other way either

I couldn't find a stated rationale for the `string` choice anywhere — neither introducing commit (`ea0e7457`, `397641f7`) gives one, and `webapp/js/constants.js:211` only describes the field's shape. If the reasoning was "a `uint8[]` reaches the browser as a bloated array of integers," that isn't what happens: rosbridge lists `uint8[]` in `ros_binary_types` and its default `binary_encoder` is `standard_b64encode` (`rosbridge_library/internal/message_conversion.py:100,135-136`; inbound it decodes base64 back to bytes at line 396). CBOR mode sends raw binary instead. The comment at `calibration/main.js:250-256` says the same thing.

So a browser subscribing to a `CompressedImage` field receives `{header, format, data: "<base64>"}` — the same base64 payload as today, plus a format tag and a stamp.

Worth noting the current consumers are all Python↔Python anyway: nothing in `webapp/` reads `image_b64` (one grep hit — the comment at `constants.js:211`), and `SearchMemory`'s image never reaches the browser at all, since as this PR's summary says, "images never cross rosbridge" — remembered JPEGs are served over HTTP at `/memory/image`.

### Bonus

- `header.stamp` could subsume `SearchMemory`'s separate `float64 seen_stamp`.
- `format` makes the currently-implicit "it's a JPEG" assumption explicit.

### Scope

- `brain_messages` rebuild (both `.action` files)
- ~6 Python touch points (the table above, plus `skills/types.py:1097` feedback signature)
- `test/test_memory_skill.py`, `test/test_runner_deactivation.py`
- the wire-contract comment at `webapp/js/constants.js:211`

Happy to leave it as-is if there's a reason I'm missing — but `RunStereoCalibration` suggests the two message families should at least agree on how a robot hands an image to a client.

karmanyaahm:
## `/brain/memory_positions`

### Action items

- [ ] **Round the floats in `store.positions()`** — x/y to 3dp, theta to 4dp, stamp to 1dp. −35% payload, sub-millimetre precision loss, independent of everything else.
- [ ] **Publish on `LATCHED_QOS`** instead of the default volatile profile.
- [ ] **Make the publish conditional** on the payload actually changing — gate on the whole payload, *not* on `store.revision`.

---

### 1 · What's actually on the wire

Built by [`recorder.py:220-230`](https://github.com/innate-inc/innate-os/blob/1223c3817e9f9e25b64b824e758b728b0e43aac3/ros2_ws/src/brain/brain_client/brain_client/memory/recorder.py#L220-L230), a `std_msgs/String` whose `data` is:

```json
{
  "map": "SmallRoom.yaml",
  "fingerprint": "a3f9c21b7e04",
  "cache": "warm",
  "positions": [
    {"id": 1, "x": 4.510457626124076, "y": -1.951365938585802,
     "theta": 0.44450798076186615, "stamp": 1786129128.9938042},
    {"id": 2, "x": -6.692107802361701, "y": -3.7324217122922185,
     "theta": 2.4540238429313175, "stamp": 1786127902.3558671}
  ]
}
```

Size (measured, not estimated):

| Memories | Payload |
|---|---|
| 1 | 206 B |
| 10 | 1.25 KB |
| 50 (the cap) | 5.9 KB |

So worst case ≈ **6 KB/s**.

One thing jumps out: **the floats are unrounded.** [`store.py:99-102`](https://github.com/innate-inc/innate-os/blob/1223c3817e9f9e25b64b824e758b728b0e43aac3/ros2_ws/src/brain/brain_client/brain_client/memory/store.py#L99-L102) passes raw `float` through, while the sibling verdict mirror at [`memory_search.py:383-390`](https://github.com/innate-inc/innate-os/blob/1223c3817e9f9e25b64b824e758b728b0e43aac3/ros2_ws/src/brain/brain_client/brain_client/brain/memory_search.py#L383-L390) rounds to 3dp. Rounding here the same way (x/y 3dp, theta 4dp, stamp 1dp) takes **6068 B → 3935 B, −35%**, at sub-millimetre precision loss. That's free and independent of everything below.

### 2 · Does it need 1 Hz? Can it be latched?

It's currently plain `create_publisher(String, ..., 10)` — VOLATILE, so a page that opens between ticks shows nothing for up to a second. Nothing about the data wants 1 Hz; the rate is just an artifact of living on the recorder's tick.

**Latched is the right call**, and it's the house pattern already — `/brain/agent_status`, `/brain/websocket_status`, and `/brain/memory_search` all use `LATCHED_QOS` ([`brain_client_node.py:50-54`](https://github.com/innate-inc/innate-os/blob/1223c3817e9f9e25b64b824e758b728b0e43aac3/ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py#L50-L54)), so transient-local replay through rosbridge is proven in this stack.

The consumers are already safe for it. [`memoriesPanel.js:120-139`](https://github.com/innate-inc/innate-os/blob/1223c3817e9f9e25b64b824e758b728b0e43aac3/webapp/js/nav/memoriesPanel.js#L120-L139) is purely level-triggered — it recomputes count/chip/reel from whatever arrives, diffs against its own signature, and runs age refresh on an independent 30 s timer. Repeated identical messages are already no-ops there, so sparse messages are equally fine.

Keep the 1 Hz tick as the change detector, make the *publish* conditional:

```python
def _publish_positions(self) -> None:
    payload = {...}
    if payload == self._last_published:
        return
    self._last_published = payload
    self._positions_pub.publish(String(data=json.dumps(payload)))
```

**One thing to get right:** `cache` changes independently of the store. It's `warm|cold|inline|unsupported|off` and flips when a background `warm()` completes or a handle expires by wall clock — neither bumps `store.revision`. So gate on the **whole payload** (as above), not on `revision`. Cheap: `cache_state()` is a snapshot + two comparisons.

**Net effect:** idle robot goes from 6 KB/s forever to one message, and a freshly-opened Nav page paints the layer immediately instead of up to a second later. Strictly better in both directions.

karmanyaahm (approved):
im so excited!

PR #615, opened 2026-08-06
feat(settings): settings 2.0

ugtthis:
Note: I have some general assumptions on the hierarchy of the settings. I don't personally use the robot or have talked to the users that do use the robot to really know what settings make sense to show up further at the top. Every setting kept its original label, description, default, and ROS path. Only page and section intro notes were new or rewritten

lmk if you think there are certain setting that make more sense to be closer to the top, or if there is something that could be better worded

## Compare UI

https://github.com/user-attachments/assets/b4250f45-aad5-458d-a18e-634555c4562d

## Problems

* total of 65 different settings with 14 expandable paths on first glance
* settings hard to scan and some toggles only had one param
* currently hover only highlights the row its over. clicking does nothing
* misaligned inputs, odd/ inconsistent spacing, etc
* redundant info(i.e. explicitly showing default values)

## Solutions

* group settings to be more oriented around a general goal (14 paths  → to now 6)
* have a basic search to bring up closest matched setting
* improved spacing, default slider,  layout, handling of error message from input
* click anywhere on the row to jump to the input box and edit versus clicking only within the input

Note: bottom fixed bar was not touched would be a separate PR

<br>

| Search feature | Comparing UI states | Overview |
| :---: | :---: | :---: |
| <video src="https://github.com/user-attachments/assets/bd31c1e5-5d21-4b4a-88cf-eaa04f62ef24" controls width="100%"></video> | <video src="https://github.com/user-attachments/assets/7d34179b-255e-44d4-a305-ed9e082090c2" controls width="100%"></video> | <video src="https://github.com/user-attachments/assets/10de115f-2d46-4ed9-80ad-bdff7be7c685" controls width="100%"></video> |

PR #628, opened 2026-08-08
fix(webapp): the MARS has no power button — say what actually brings it back

karmanyaahm:
The Shutdown button added in #622 tells the operator, in both the confirm dialog and the
button tooltip, to press the robot's power button to turn it back on. The MARS has no power
button — the only way back up is to unplug the battery and plug it in again. Someone who
powers the robot off from settings and then goes looking for a button has no way to recover
it.

Text only, three lines:

- confirm: "…the only way back up is to unplug the battery and plug it in again."
- tooltip: "Power off the Jetson (unplug and replug the battery to bring it back up)"
- the comment above the disabled-buttons block, which said the same thing

Note for #615: that PR rewrites `webapp/js/settings/main.js` wholesale and carries the old
wording, so whichever of the two merges second needs this carried across.


theo-michel (approved):
Yeah mb

PR #635, opened 2026-08-10
feat(motor sound): 34 acceleration voices, and the lip trill as default

theo-michel:
**Stack** — merge bottom-up:

| | PR | |
|---|---|---|

> Stacked on #653 — base is `fix/startup-sound-wav`, whose GCS downloader this PR reuses for its own clips. Merge #653 first and this retargets to `main` cleanly.
>
> #630 was closed unmerged, so its Mad-mode tune (`tune(motor sound): drop Mad mode into the low register, snarl at redline`) is carried here as the first commit — merging this lands that too.

**Audition board with every voice playable: https://claude.ai/code/artifact/60e85308-a99a-4ae7-8c08-ee0e0f41666f**

Same nine-second drive rendered through each — park, floor it, hold top speed, ease down — straight out of the synth in this PR, so what you hear is what the robot plays.

## The default changes

`motor_sound.voice` now defaults to `lip_trill`: a mouth going brrrrrr.

A synthesised motor is a *claim about what the robot is*, and everybody is equipped to hear that claim as slightly wrong — which is why the whine went through a V8, a retune and a snarl and still only ever got closer to passable. A mouth noise makes no such claim, so there is nothing about it left to be wrong. It is also the safest thing to boot into: entirely procedural, so unlike the sampled voices it cannot be silenced by an asset that failed to install.

The whine is unchanged and one parameter away:

```bash
ros2 param set /motor_sound motor_sound.voice motor
```

## What's here

34 voices behind one parameter, swappable live — the node rebinds the reference and the new voice ramps its gain from zero, so the handover cannot click. They share the motor's *feel* by construction: `Rotor`, `bandpass_kernel` and `Stream` are now public in `motor_synth` and reused, so the same underdamped spring lags the heard speed behind the real one whichever costume is on.

| | |
|---|---|
| **Made of mouths** | `lip_trill` `mars_voice` `mars_song` `kazoo` `whistle` `choir` `throat` |
| **Swiss Army Man** | `jetski` `acappella` |
| **Impossible** | `shepard` `risset` `doppler` |
| **Wrong domain** | `tape` `spoke_card` `geiger` |
| **Alive** | `panting` `heartbeat` |
| **Musical** | `music_box` `harp` `disco` `chiptune` `bagpipe` `gamelan` |
| **Creatures & machines** | `purr` `steam_train` `theremin` `hyperdrive` |
| **Sampled** | `gallop` `quack` `boing` `didgeridoo` `whale` `f1` `crowd` |

Three ideas worth singling out. **`shepard`** is a Shepard tone — it rises forever and never gets higher, which is the only honest answer to a robot that *has* a top speed. **`doppler`** models a real flypast rather than faking one (pitch over one-plus-radial-velocity, loudness over one-over-distance) so the swoop and the level agree. **`f1`** is a real race engine, included deliberately as a control condition: it's the obvious answer, and having it on the board is how you rule it out.

## Shared machinery, not 34 copies

`Metronome` banks the fractional remainder so an event rate stays exact while it changes every block. `PluckBank` renders decaying notes from their absolute age, so where block boundaries fall across a note changes nothing. `_cycle` turns any oscillator into a repeating envelope — that's how the chuffing, sputtering and four-on-the-floor voices keep a rhythm without a pool of events in flight. `LoopVoice`/`TriggerVoice` cover all seven sampled voices between them. `clamp01` and `Stream` (a FIR that carries its tail across block boundaries, which `motor_synth` had hand-rolled three times) moved into `motor_synth` beside `Rotor` and `bandpass_kernel` rather than being copied — the whine renders bit-identical across a drive cycle.

Everything is phase-continuous across blocks and allocates only per block, because `render` runs on the PortAudio thread.

## Things measurement caught that listening would not

- **Formants were mud.** The first vowel voices filtered the finished buzz with a 95-tap FIR, which at 48 kHz has a ~2 kHz transition band — no such filter can resolve a 120 Hz formant. Formants now weight each harmonic as it is summed: sharper, cheaper, and it tracks a pitch sweep with no zipper.
- **`mars_voice` could not say brrrrr, and neither can TTS.** Across ten spellings nothing held steady past 0.28 s and none of it flapped. The clip is now the longest steady `rrr` the robot's real voice *does* hold, and the synth flaps it at 19–34 Hz — which is what a lip trill physically is anyway.
- **`pcm_44100` is stereo-interleaved.** Read as mono, every sampled asset would have played at half rate with the channels combed together.
- **Every sampled trigger fired with a click of its own**, because extraction started wherever the peak was. They now start from silence with a 2 ms attack.
- **Four voices came out far too bright to keep** — raw white noise in a transient dumps most of its energy in the top octave. The Geiger tick was worst at a 5.7 kHz centroid with 60% above 2.5 kHz; it is now 1.6 kHz and 3.8%. The duck was 52% and is now 3.8%.
- **`theremin`'s portamento was a stair**, not a slide: it smoothed the same variable the anti-zipper glide reads, so every block started where it ended.
- **A cricket was cut entirely.** Chirp-rate-as-speed is a lovely idea, but real ones sit at 5 kHz and nothing left it both audible and a cricket.

## Loudness needed three metrics

Choosing a voice must not mean choosing a volume, so each carries a measured `trim`. RMS suits drones; the sparse voices carry crest factors of 9–13 against the motor's 4.2, so matching *their* RMS drove them far enough into the saturator to flatten the clicks into blobs, while matching their peak left them four times too quiet. Those are peak-matched and documented as genuinely quieter — that gap is physics, not a tuning error. `gallop` sits at the middle setting that holds a crest of 6.4; the trim that reaches the motor's RMS flattens it to 4.9, which is the horse gone.

## Assets

The eight clips are **not in git**. They live in `gs://karmanyaah-public/sounds/motor_voices/` (~190 KB total) and `post_update.sh` fetches them at step 11c, the same mechanism #653 introduced for the boot sounds — the two downloads it now has are one `fetch_asset` helper over one list.

They land in **`config/sounds/motor_voices/`**, beside `turnon.wav` and `turnoff.wav` and mirroring the bucket's own layout, because a download does not belong in a source tree. The node resolves them through `$INNATE_OS_ROOT` (`~/innate-os` default), which is how every other node here finds the repo — `app.cpp` in this same package already does it. Nothing is installed by CMake and nothing needs a rebuild: dropping a new clip in that directory is enough.

Missing clips are never fatal — each sampled voice names a synthesised understudy (`f1` → `hyperdrive`, `gallop` → `music_box`), so a workspace that never fetched them still makes a noise.

Two scripts regenerate the clips — `fetch_mars_voice_loop.py` (Cartesia, the robot's own voice) and `fetch_sound_assets.py` (ElevenLabs). Both read credentials from the environment; no key is committed. Publishing is now the second half of regenerating, and both docstrings carry the `gsutil cp`. Loops are cut to whole pitch periods and crossfaded end-to-start, so the runtime wraps them with no click.

Each clip is pinned by sha256, following #653: a fresh download is verified before it is installed and one already on disk is verified rather than trusted, so a truncated or tampered file heals on the next update — and republishing a changed clip means updating its sha, not inventing a new filename.

## Test plan

- [x] `ruff check` / `ruff format --check` clean
- [x] All 34 voices render finite audio across a full drive cycle, both with the clips present and with the directory empty (understudy path)
- [x] All eight objects public and byte-identical over the round trip: uploaded, deleted locally, re-fetched from the URLs `post_update.sh` uses, `cmp` clean
- [x] `MotorSynth` output bit-identical before and after the `Stream` refactor (200 blocks, fixed seed)
- [x] Node imports with ROS sourced; constructing it builds `LipTrill` and the yaml and declared defaults agree
- [x] Levels, spectral centroid and >2.5 kHz energy measured for every voice
- [ ] **Listen on the robot's actual speaker.** It rolls off below ~160 Hz, and `purr`, `jetski` and `didgeridoo` lean lowest, so they lose the most in the move from headphones
- [ ] Confirm the live `voice` swap is click-free on hardware
- [ ] One `innate-update` run on a robot, confirming step 11c lands the clips where the node reads them


karmanyaahm:
is this really necessary? like im sure the code is good but do we need this, its like 2500 lines of code

karmanyaahm (approved):

PR #636, opened 2026-08-10
fix(sim): recover silent ROS stream stalls

axelpey:
## Summary

- Observed two silent stalls after action traffic: an `rws` client kept its WebSocket open but stopped forwarding subscribed ROS topics, and `virtual_mars` stopped running ROS timers/subscription callbacks while its render thread and host world stream stayed alive. `/cmd_vel` remained present on the ROS bus but was no longer applied.
- A new temporary subscription immediately revived the existing `rws` client, and the driver timers were all ready but increasingly overdue. This points to stale executor/wait-set state around `rmw_zenoh` dynamic graph/action churn; the exact lower-level defect is not established.
- Mitigate the driver stall by waking its executor guard condition when the 30 Hz odometry timer is more than one second overdue. Mitigate the browser stall by closing an already-flowing socket after 15 seconds with no frames, allowing the existing reconnect path to replay subscriptions.
- These are intentionally narrow recovery mechanisms, not a claim that this is the preferred root fix; a lower-level `rws` or `rmw_zenoh` change may be better.

## Reproduction without this fix

1. Start the simulator with `./innate-sim up`.
2. In the simulator UI, navigate the robot to a goal pose.
3. Repeat navigation to different goal poses several times.
4. Eventually, the simulator can stop receiving the ROS data stream even though the WebSocket still reports itself as connected. The UI then stops reflecting robot movement or other live state until the stream recovers or the connection is recreated.

The failure is intermittent, so the number of navigation attempts varies.

## Validation

- [x] Python compilation and Ruff pass for the driver change.
- [x] Browser syntax, existing RosClient transport tests (9), and changed-file TypeScript checks pass.
- [x] Launcher/proxy test subset passes (17 tests).
- [x] Reproduced live: the browser socket stayed open with frame delivery frozen; a fresh subscription revived the old client. Separately, `virtual_mars` showed all ROS timers overdue while rendering continued and `/cmd_vel` remained on the bus.
- [ ] I did not launch a second full simulator from the isolated PR worktree because the equivalent patch was already running in the active simulator checkout.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/node.py
+++ b/ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/node.py
@@ -72,6 +72,12 @@ SCAN_HZ = 6.0  # lidar.launch.py throttle
 JOINT_STATE_HZ = 30.0
 ARM_STATE_HZ = 20.0
 HEAD_POSITION_HZ = 10.0
+# rmw_zenoh can leave an executor asleep after dynamic graph/action churn even
+# though its timers are ready. A non-executor guard-condition wake is cheap and
+# specifically repairs that stale wait set; only intervene after a full second
+# of missed 30 Hz odometry callbacks.
+EXECUTOR_STALL_S = 1.0
+EXECUTOR_WATCHDOG_PERIOD_S = 0.5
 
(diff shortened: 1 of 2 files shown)


axelpey:
was fixed separately

PR #637, opened 2026-08-10
feat(sim): add household orders challenge

axelpey:
## Summary

- Add the Household Orders simulator challenge with three visually distinct residents on reachable interior floor.
- Keep orders private to the environment, reveal them only through proximity-, facing-, and request-intent-gated dialogue, and require a normalized exact or explicitly allowlisted complete readback.
- Track the three resident confirmations independently before unlocking simulated DoorDash checkout.
- Add the resident assets and declarations required by this environment.

## PR stack

1. #714 — reusable simulated-NPC runtime, chat/TTS bridge, kinematic characters, labels, and optimized GLB support
2. **This PR** — Household Orders scenario data, resident assets, dialogue policy, and behavior tests
3. #680 — autonomous Household Orders agent and reusable skills
4. #689 and #694 — optional navigation/experiment branches stacked on #680

The generic simulator-character infrastructure was extracted into #714 so this PR contains no agent, skill, localization, lighting, or generic chat implementation.

## Validation

- [x] 17 Household Orders behavior tests
- [x] 5 asset-pipeline tests
- [x] 14 generic runtime/TTS tests in #714
- [x] Ruff check and format validation

## Current limitations

- This setup PR does not include or activate the autonomous agent; see #680.
- Checkout remains local simulation and does not contact DoorDash or spend money.
- The judge gates checkout on all resident confirmations and skill completion, but does not semantically compare submitted checkout strings with its private order facts.


sim/uv.lock
```diff
@@ -533,6 +533,9 @@ dependencies = [
     { name = "numpy", version = "2.4.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version == '3.11.*'" },
     { name = "numpy", version = "2.5.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12'" },
     { name = "pillow" },
+    { name = "scipy", version = "1.15.3", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.11'" },
+    { name = "scipy", version = "1.17.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version == '3.11.*'" },
+    { name = "scipy", version = "1.18.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12'" },
```
DavidDobas: Publish the resident_*_collision_*.obj files to innate-sim-assets like the other authored inputs, and decompose_resident, and avoid scipy dependency

sim/uv.lock
```diff
@@ -1537,6 +1541,193 @@ wheels = [
...
+    { url = "https://files.pythonhosted.org/packages/db/0a/92b1de4a7adc7a15dcf5bddc6e191f6f29ee663b30511ce20467ef9b82e4/scipy-1.15.3-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:263961f658ce2165bbd7b99fa5135195c3a12d9bef045345016b8b50c315cb82", size = 35547617, upload-time = "2025-05-08T16:04:43.546Z" },
+    { url = "https://files.pythonhosted.org/packages/8e/6d/41991e503e51fc1134502694c5fa7a1671501a17ffa12716a4a9151af3df/scipy-1.15.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:9e2abc762b0811e09a0d3258abee2d98e0c703eee49464ce0069590846f31d40", size = 37662964, upload-time = "2025-05-08T16:04:49.431Z" },
+    { url = "https://files.pythonhosted.org/packages/25/e1/3df8f83cb15f3500478c889be8fb18700813b95e9e087328230b98d547ff/scipy-1.15.3-cp310-cp310-musllinux_1_2_aarch64.whl", hash = "sha256:ed7284b21a7a0c8f1b6e5977ac05396c0d008b89e05498c8b7e8f4a1423bba0e", size = 37238749, upload-time = "2025-05-08T16:04:55.215Z" },
```
DavidDobas: Same as comment above

DavidDobas (changes requested):
## Request changes — asset pipeline

The challenge/runtime/dialogue side of this PR is solid and this review doesn't block on it. What I'd like reworked before merge is the **resident asset integration**: roughly 500 lines of build-time asset logic that the repo's own conventions make unnecessary. We already have the precedent for exactly this in `casual_man` — converted **offline, once**, with the converted outputs published to `innate-sim-assets` and seeded as plain `RAW_FILES`. The residents should land the same way.

### Preferred: three Sketchfab scans, the `casual_man` recipe

1. Pick three visually distinct standing CC-BY scans (some ship OBJ alongside GLB; either works as the source).
2. Convert offline, once, same as `casual_man`: normalized MuJoCo OBJ (Y-up cm, feet at origin — and pre-rotate to Z-up so no `base_quat` field is needed), single base-color texture extracted.
3. Optimize the viewer GLB before publishing — the 23 MB / 67 MB-decoded `human.glb` is the documented cause of the drop-time jank `propModels.ts` works around, so let's not triple it:
   ```
   npx @gltf-transform/cli optimize scan.glb resident_x.glb --texture-size 1024 --simplify-error 0.001
   ```
   plus strip normal/roughness maps (this PR already authors residents matte). ~2–4 MB per person on disk, ~5 MB decoded.
4. Publish OBJ + texture + GLB per resident to the `sim-geometry-v1` release, add ~9 entries to `RAW_FILES`, add ATTRIBUTION.md entries.
5. Use `collision="hull"` like the existing human. A standing person the robot only approaches and talks to doesn't need anatomical hulls — and a convex hull avoids the concave leg-slot hazard the band decomposition has to comment its way around. This removes the collision pipeline entirely, going one step further than my inline comments on `uv.lock`.

**Acceptable alternative:** keep the Quaternius characters (a good pick — tiny, distinct, CC0) but run this PR's bake **once, locally**, and publish its *outputs* the same way. The transform is deterministic over frozen CC0 inputs; keeping it executable in CI is authoring disguised as derivation. Park the script in the PR description or a gist for provenance. If the band hulls are kept under this path, do what the `uv.lock` comments already ask: publish the pre-baked `resident_*_collision_*.obj` like the other authored inputs and drop `decompose_resident` + scipy.

### What either path deletes

| File | Remove | Why |
|---|---|---|
| `ci/seed_asset_context.py` | ~270 of the +279 lines: `ResidentSource`, the Google Drive fetcher, `_parse_mtl_colors`, `_palette_png`, `_resident_assets`, `export_resident_glbs` + its ten `sys.exit` validators, `iter_seed_destinations`, the `--export-resident-glbs` CLI mode | Becomes ~9 data-only `RAW_FILES` entries. Also removes a real CI risk: Google Drive serves quota/interstitial HTML with HTTP 200, which fails here as a misleading "checksum mismatch". The seeder's own header says derived geometry is not seeded — this PR makes it derive geometry. |
| `sim/Dockerfile.assets` | The `COPY ci/seed_asset_context.py ../ci/…` build-context escape and the `--export-resident-glbs` stage | Three plain `COPY` lines for the GLBs remain, identical to the `human.glb` line above them. |
| `sim/Dockerfile.assets.dockerignore` | `!ci/seed_asset_context.py` | No longer referenced. |
| `sim/tools/decompose_objects.py` | `decompose_resident`, `convex_slab`, `RESIDENT_BANDS`, the humans-dir source discovery | `collision="hull"` needs no pieces (or, alternative path: pieces are published, not derived). |
| `sim/pyproject.toml` / `sim/uv.lock` | `scipy` (+191 lock lines) | As per the existing inline comments: nothing imports scipy; it's only trimesh's backend for `convex_hull`, called only by `decompose_resident`. |
| `sim/tools/build_viewer_physics.py` | The humans-dir loop | No resident hull soups exist under `hull` collision. |
| `mars_sim_driver/props.py` | `mesh_asset_key` / `_mesh_asset_name` dedup and the `asset_name` threading; `primitive_offset`; `base_quat` (if meshes are pre-rotated offline) | The dedup is provably inert in this PR — the residents have three distinct meshes/scales/GLBs and `mesh_scale` is part of the key, so nothing ever collides (this PR's own `test_household_residents_use_distinct_visual_sources` asserts the sources are distinct). `primitive_offset` only decorates the missing-mesh fallback box. |
| `sim/viewer/src/propModels.ts` | `#sourceLoads` / `sourceKey` / SkeletonUtils clone refactor | Same inertness: three distinct GLBs, and the pre-existing 1:1 prop→GLB model already covers them. |
| Tests | The bake/seed-destination tests in `test_assets_image_inputs.py`, the asset-dedup tests in `test_household_npc.py` | Nothing left to cover. |

### What stays (and is right)

- **`kinematic` / mocap support in `props.py`** — the one genuinely new capability here. A 1.7 m freejoint biped topples; this ~40-line addition is needed under every variant and is cleanly done.
- The three resident sidecars, the challenge package, runtime, and chat bridge — out of scope for this review.
- `Pillow` in `ros2_ws/pip-requirements.txt` — unrelated to residents but fixes a real pre-existing undeclared import; keep it.

Net effect: same three residents on screen, same physics, ~500 fewer lines, no new dependency, no lockfile churn, no third-party download host in CI — and the asset story stays the one the seeder's header already tells.

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,99 @@
...
+        self._shared.clear()
+        self._confirmed.clear()
+
+    @staticmethod
+    def _matches(resident: Resident, text: str) -> bool:
+        normalized = _normalize_speech(text)
+        return all(
+            any(_normalize_speech(alias) in normalized for alias in alternatives)
+            for alternatives in resident.required_facts
+        )
```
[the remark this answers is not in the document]
axelpey: Fixed by replacing substring-only acceptance with exact whole-readback matching plus semantic fact validation. Excluded items now have to appear in an explicit exclusion and are rejected if they also remain as a positive item mention. The regression is covered by test_contradictory_readback_does_not_confirm.

sim/uv.lock
axelpey: Addressed using the preferred hull route, so there are no generated resident collision pieces to publish. The three residents now use distinct offline-converted restore50 Sketchfab CC-BY scans, each published as checksum-pinned OBJ, base-color texture, and optimized GLB inputs (9 RAW_FILES total), with collision="hull". The build-time resident decomposer and SciPy/lockfile churn were removed.

sim/uv.lock
axelpey: Addressed together with the thread above: resident collision decomposition and SciPy were removed, and each resident now uses the published source mesh with collision="hull".

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @staticmethod
+    def _matches(resident: Resident, text: str) -> bool:
+        normalized = _normalize_speech(text)
+        has_every_fact = all(
+            any(_normalize_speech(alias) in normalized for alias in alternatives)
+            for alternatives in resident.required_facts
+        )
+        contradicts = any(_normalize_speech(phrase) in normalized for phrase in resident.contradictions)
+        return has_every_fact and not contradicts
```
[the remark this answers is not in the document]
axelpey: Fixed. Validation now removes only explicit negative forms of excluded items, then rejects any remaining positive occurrence. The regression suite includes the exact no-cheese-but-put-cheese wording.

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def reset(self) -> None:
+        self._shared.clear()
+        self._confirmed.clear()
+
+    @staticmethod
+    def _matches(resident: Resident, text: str) -> bool:
+        normalized = _normalize_speech(text)
+        if _REVERSAL_WORDS.search(normalized):
+            return False
```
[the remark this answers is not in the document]
axelpey: Fixed by removing the broad reversal-word blacklist. Common conversational prefixes such as Actually are stripped, while negation is evaluated locally around the relevant order fact. Correct connective-word and natural-paraphrase readbacks are covered by regression tests.

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,115 @@
...
+
+        for item in resident.excluded_items:
+            item = _normalize_speech(item)
+            remainder = normalized
+            for alternatives in resident.required_facts:
+                for alias in alternatives:
+                    alias = _normalize_speech(alias)
+                    if item in alias:
+                        remainder = remainder.replace(alias, " ")
```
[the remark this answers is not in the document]
axelpey: Fixed. Negation around an exclusion is now checked before that exclusion is removed, so double negations such as not no cheese and do not remove the cheese are rejected. Both forms are in the parametrized regression test.

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,142 @@
...
+            for match in matches:
+                if _NEGATION_PREFIX.search(remainder[: match.start()]):
+                    return False
+            remainder = exclusion.sub(" ", remainder)
+            if re.search(rf"\b{phrase}\b", remainder):
+                return False
+
+        for alternatives in resident.required_facts:
+            if not any(HouseholdOrdersRuntime._has_positive_phrase(normalized, phrase) for phrase in alternatives):
+                return False
```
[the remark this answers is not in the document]
axelpey: Fixed in 6c7a051f. Every alias in a required-fact group is now checked for both positive and negated occurrences: the group needs at least one positive alias and no negated alias. Added the exact ShackBurger / no Shack Burger regression; all 34 household NPC tests pass.

axelpey:
@DavidDobas I addressed the requested asset-pipeline cleanup on this PR:\n\n- Replaced the generated resident geometry with three distinct standing restore50 Sketchfab scans under CC BY 4.0.\n- Converted, normalized, and optimized them offline, then published OBJ + base-color texture + GLB for each resident as 9 checksum-pinned RAW_FILES.\n- Switched all three resident props to collision="hull".\n- Removed the build-time resident downloader/baker/decomposer, SciPy and lockfile churn, resident band decomposition, and the inert dedup/source-cloning machinery.\n- Kept the challenge/runtime/dialogue behavior, Pillow for challenge images, kinematic/mocap prop support, and the viewer support needed to decode and preserve the optimized normalized geometry.\n- Added/updated checks for distinct visual sources, hull collision, normalized assets, and attribution.\n\nThe later runtime-review findings are also fixed, including contradictory alias spellings in 6c7a051f. The focused household NPC suite is 34/34 passing, and the PR checks were green before this final small regression fix. Could you please re-review when convenient?

axelpey:
@DavidDobas I implemented the resident presentation follow-up in 2ecedf08:\n\n- Alex, Blake, and Casey now have browser-only floating nameplates attached to their Three.js prop roots. They do not exist in MuJoCo, so the robot camera/agent perception remains unchanged.\n- Resident dialogue now goes through the brain's existing Cartesia TTS path with voice selection per resident. Alex and Casey use the feminine Katie voice; Blake uses the masculine Jameson voice.\n- The transcript is withheld from /brain/chat_in while the audio plays. The TTS worker acknowledges completion, then the challenge bridge releases the named transcript into normal chat with a completion-time timestamp.\n- The completion remains challenge-run scoped, so an abort/restart cannot leak a delayed utterance into the next run.\n- If TTS is unavailable, it fails open to the text transcript rather than blocking the challenge.\n\nValidation: pre-commit passed, 85 repository tests passed, 37 household-dialogue tests passed, 2 TTS timing tests passed, and the viewer typecheck + production build passed.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -320,6 +440,18 @@ def __init__(
         # fresh run judged against the previous one's clock and props.
         self.world_epoch = 0
         self._run_epoch = 0
+        # Environment replies cross from the physics tick to a rosbridge publisher
+        # thread. Every item carries the run token that created it; starting or
+        # aborting a challenge increments the token, so a delayed reply from an
+        # old scene can never speak inside a new one.
+        self._run_token = 0
+        self._speech_serial = 0
+        self._chat_inputs: deque[tuple[int, dict]] = deque()
```
DavidDobas: **Unbounded outbox.** `_chat_inputs` has no `maxlen` and is only drained by `next_chat_input`. When the optional `websockets` client is missing, `_publish` returns immediately (the documented best-effort case) and `tick` keeps appending every runtime reply forever — plus `_deactivate` bumps the token but leaves stale items queued. A long-lived world server accumulates every NPC line ever generated.

Suggest `deque(maxlen=...)` plus a `self._chat_inputs.clear()` in `_deactivate` (stale items are dead there anyway — the token check already rejects them on pop).

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -548,6 +802,197 @@ def _block(self, challenge: Challenge | None) -> dict:
...
+        except Exception as exc:
+            if timed_out.is_set():
+                raise TimeoutError("challenge chat websocket write timed out") from exc
+            raise
+        finally:
+            finished.set()
+            # A successful send cannot leave a timer behind that later closes
+            # a connection already reused for another reply.
+            watchdog.join()
+        if timed_out.is_set():
```
DavidDobas: **A send that succeeds at the deadline still raises → duplicate resident line.** If the write stalls ~2 s and completes in the same instant the watchdog fires, `connection.send()` returns normally, `finally` sets `finished` — and this trailing `if timed_out.is_set()` raises anyway. `_publish` then closes the socket, sleeps 1 s, and re-sends the same payload (token still current), so the resident says e.g. "That's correct. Thank you." twice and the agent sees a duplicated chat turn.

The frame was delivered; only raise when `send()` itself raised (the `except` branch already handles that case). Alternatively — simpler and removes the race class entirely — drop the watchdog and rely on the token re-check before send: the residual harm on this best-effort path is one stale chat line, which the token check already keeps out of the judge.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
```diff
@@ -260,14 +359,33 @@ def load_challenges(roots: list[Path]) -> dict[str, Challenge]:
     for root in roots:
         if not root.is_dir():
             continue
-        for path in sorted(root.glob("*.py")):
-            if path.name.startswith("_"):
-                continue
+        files = [path for path in root.glob("*.py") if not path.name.startswith("_")]
+        packages = [path / "__init__.py" for path in root.iterdir() if path.is_dir() and not path.name.startswith("_")]
```
DavidDobas: **Every subdirectory is assumed to be a challenge package.** This builds a `__init__.py` path for every non-underscore directory without checking it exists — so a plain data/fixtures folder beside the sidecars makes `exec_module` raise `FileNotFoundError`, printed on every server start as `[challenges] skipping __init__.py: …` — naming neither the directory nor the real cause.

Filter with `if (path / "__init__.py").is_file()`.

DavidDobas (approved):

DavidDobas:
Pushed two fixes after testing the voice flow end to end on macOS:

**`86cdadd4` — don't lose resident replies to rosbridge delivery**
- The reply channel now connects to rws eagerly at world-server boot (but 10 s clear of rws startup, which has a known handshake-hang window — see `sim_rosbridge.launch.py`) instead of lazily on the first reply.
- Dropped the `advertise` op; rws resolves topic types itself, and bare publishes are what the webapp already uses.
- Added a 30 s watchdog on the playback ack: if the TTS request or its ack is lost anywhere between the world server and the brain, the resident's transcript is posted as text anyway (exactly once — ack and watchdog race, first wins). Before this, one lost message made the resident silently ignore you forever.

**`6dcdde19` — keep a speaking reply's sentences, subtitle the residents**
- Replies stream to TTS sentence-by-sentence, and a new reply's `replace_pending` flush used to drop the *rest of the reply currently being spoken* (heard as "Understood! … You're welcome!" with the actual order read-back skipped). Sentences now carry a reply id, and the reply holding the floor keeps its remaining sentences; only replies that never started speaking get flushed.
- Environment speech is `protected` — an agent reply flush can never cancel a resident's queued clip (which also guarantees the bridge its ack).
- Resident transcripts now post when their clip *starts* playing rather than after it finishes, so the text leads the voice slightly, same as the robot's own messages.

**One behavioral note on the challenge design:** residents currently disclose their order on *any* robot speech within 1.5 m — there is no check that the robot actually asked about an order. A stray "Hello there!" near Alex triggers the full Chipotle disclosure, and every subsequent utterance is treated as a read-back attempt. Might be worth gating the disclosure on the utterance mentioning ordering/food, or at least ignoring greetings.

axelpey:
@DavidDobas The resident order disclosure issue is fixed in 4584cd38. Residents now stay silent for unrelated nearby speech and disclose their order only when the robot asks with order intent (for example, “What would you like?” or “What is your DoorDash order?”). This is deterministic and covered by regression tests for both unrelated speech and several natural order-request formulations.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
axelpey: Fixed in #714 (squashed commit 6ed00423). `_send_with_timeout()` now reports a timeout only when `send()` raises after the watchdog interrupts it; a send that returns successfully at the deadline is accepted and is not retried. Added the exact deadline-race regression in `test_chat_write_that_completes_at_deadline_is_not_retried`.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
axelpey: Fixed in the generic prerequisite PR #714 (squashed commit 6ed00423). The NPC outbox is now bounded to 64 entries and `_deactivate()` clears it while holding the engine mutex, so missing rosbridge cannot grow memory indefinitely and stale run items are removed immediately. Covered by `test_deactivate_clears_a_bounded_chat_outbox`.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/challenges.py
axelpey: Fixed in #714 (squashed commit 6ed00423). Challenge package discovery now includes a directory only when its `__init__.py` exists, so adjacent fixtures/data directories are silently ignored. Covered by `test_loader_ignores_plain_data_directories`.

DavidDobas:
@greptileai

sim/challenges/40_household_orders/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+            "Get Blake to confirm the order",
+            EventSeen("resident_order_confirmed", {"resident": "blake"}),
+            parallel_group="residents",
+        ),
+        Goal(
+            "Get Casey to confirm the order",
+            EventSeen("resident_order_confirmed", {"resident": "casey"}),
+            parallel_group="residents",
+        ),
+        Goal("Submit all three orders to DoorDash", SkillDone("place_doordash_order")),
```
theo-michel, line 112: **Real, but a merge-ordering call rather than a code defect.**

This goal can't fire until the agent PR lands: `place_doordash_order` exists nowhere on this branch (only #680 carries it), so anyone who picks the challenge from the panel today confirms all three orders and then watches the run die on the 900 s clock. I documented the gap in the module docstring (victory_lap precedent), but the real decision is yours: is merging this ahead of #680 with a known-unpassable final goal acceptable, or should the two land together?

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+    def update(self, state: WorldState, events: list[dict]) -> RuntimeResult:
+        result = RuntimeResult()
+        for event in events:
+            if event.get("type") != "robot_speech" or not isinstance(event.get("text"), str):
+                continue
+            resident = self._nearest(state)
+            if resident is None or resident.id in self._confirmed:
+                continue
+            if resident.id not in self._shared:
+                if not _asks_for_order(event["text"]):
```
theo-michel, line 176: **Worth fixing before merge.**

Decision point: when this heuristic rejects an utterance the robot gets **nothing back** — indistinguishable from being out of range or facing the wrong way. And the heuristic misses a lot of natural phrasings; all of these return `False`:

- “Hi Alex! I'm here to take your DoorDash order.”
- “I need to collect your DoorDash order.”
- “Hello! What are your orders?” (plural ‘orders’ isn't in `_ORDER_TOPICS`)
- “What can I bring you?” / “What are you hungry for?” / “Any takeout requests?”

The post-disclosure path always nudges with “Not quite…”; this pre-disclosure path fails closed and silent, so an agent can burn the whole time limit with zero signal. Options: broaden the heuristic (match on `order`/`orders` as a topic more liberally), and/or have an un-asked resident reply with a neutral prompt (“Are you here for my order?”) instead of silence. Both change the scenario's difficulty, so your call.

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+                continue
+            if self._matches(resident, event["text"]):
+                self._confirmed.add(resident.id)
+                result.events.append({"type": "resident_order_confirmed", "resident": resident.id})
+                result.replies.append(EnvironmentReply(resident.name, "That's correct. Thank you.", resident.voice_id))
+            else:
+                result.replies.append(
+                    EnvironmentReply(
+                        resident.name,
+                        f"Not quite. {resident.order} Please repeat the complete order back to me.",
```
theo-michel: **Worth fixing before merge.**

Decision point: once an order is shared, **every** utterance within 2 m and the facing cone lands here — there's no intent gate on the readback side. An agent that narrates near Blake (“Okay, heading to the kitchen now”) gets the full order re-read on every sentence. Those replies are spoken with `protected=True` (agent flushes can't cancel them), so a chatty agent monopolizes TTS and can fill the engine's 64-slot chat outbox, silently dropping the oldest replies. Worth considering: only treat an utterance as a readback attempt if it mentions any order fact, or rate-limit corrections per resident.

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+            if not matches:
+                return False
+            for match in matches:
+                if _NEGATION_PREFIX.search(remainder[: match.start()]):
+                    return False
+            remainder = exclusion.sub(" ", remainder)
+            if re.search(rf"\b{phrase}\b", remainder):
+                return False
+
+        return bool(resident.required_facts) and all(
```
theo-michel, line 120: **Real, but rarely hit in practice — an accurate comment may be enough.**

Known-limitation flag, mirroring the checkout caveat in the PR description but for readbacks: this path checks fact **presence** only, so additions and inverted framing still confirm. Verified against the shipped data:

- `“Shake Shack Shack Burger, cheese fries, vanilla shake, no pickles, and also add a hot dog”` → confirms Casey
- `“I cannot get you a Chipotle chicken bowl with brown rice, black beans, mild salsa, no cheese.”` → confirms Alex (`_NEGATION_PREFIX` only looks a few tokens back, so whole-utterance polarity is invisible)

If that looseness is an accepted trade for determinism, maybe say so in the comment above (it currently claims ‘contradictory readbacks’ fail, which holds for exclusions but not for additions/inversions).

theo-michel (commented):
Reviewed the whole PR (engine interplay, asset pins, nav-map placement, matching logic). I pushed 141d2169 with the mechanical fixes: Alex/Casey yaws (both faced the nearest wall, 0.7-0.85 m away — the scans face +y at identity), a docstring note that `place_doordash_order` ships with the agent, the facing-cone comment (the gate reads base yaw, not the head camera), one source of truth for the 2.0 m dialogue radius, a dead `None` guard in `_nearest`, the seeder-test skipif, and comment dedup across the sidecars.

### Triage of the line comments

I filed six and they are not equal in weight, so rather than leave you to sort them:

**Worth fixing before this merges (2):**

- **`runtime.py` — the `_asks_for_order` gate fails silently.** This is the one I would actually push on. The README addition in this PR says the point is "select or build the agent separately, then run it against the challenge" — but the front door only opens for roughly half of natural phrasings, and a miss produces *no reply at all*. An agent developed alongside this heuristic will naturally phrase requests it accepts; an independently built one hits undebuggable silence. That is the challenge not delivering on its own stated purpose.
- **`runtime.py` — the correction path re-reads the full order on every utterance** near a shared resident, over uncancellable `protected=True` TTS. Agents narrate, so this degrades ordinary runs, not just adversarial ones.

**Real but not urgent (2):**

- **`__init__.py` — the checkout goal is unpassable until #680.** A merge-ordering call, not a code defect; if #680 lands right behind this, it never mattered. It only bites if this ships alone in a release.
- **`runtime.py` — the judge accepts additions and inverted framing.** A hole in scoring integrity, and honest verification is the engine's whole pitch — but an agent genuinely attempting a readback phrases positively, so it rarely misfires. Making the comment accurate may be a better trade than rewriting the matcher.

**Take or leave (2):**

- The dead exact-match branch, and the Alex/Casey voice collision. The one piece of the voice comment I would act on regardless is flipping the test that currently *asserts* the collision.

If you only act on two, make them the first two — they are the difference between a challenge that works with any agent and one that works with its sibling PR.

### Verified clean

All six release-asset SHA-256s match `ci/seed_asset_context.py`; the OBJ bounding boxes match the declared `size`/`center_offset` exactly; the GLBs' `EXT_meshopt_compression` is handled by the viewer loader from #714; all three drop points sit on free nav-map floor; ruff and the 36 tests pass on the new head.

One minor thing not worth its own thread: `tests/test_household_npc.py` copies the mujoco stub / `sys.path` bootstrap and the fake-sim double from `tests/test_challenge_runtime.py` (#714) — a shared `tests/conftest.py` somewhere in the stack would let both drop it.

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,173 @@
...
+        self._shared.clear()
+        self._confirmed.clear()
+
+    @staticmethod
+    def _matches(resident: Resident, text: str) -> bool:
+        normalized = _normalize_speech(text)
+        while prefix := next((prefix for prefix in _READBACK_PREFIXES if normalized.startswith(prefix)), None):
+            normalized = normalized[len(prefix) :]
+        accepted = {_normalize_speech(readback) for readback in (resident.order, *resident.accepted_readbacks)}
+        if normalized in accepted:
```
theo-michel, line 97: **Nit / take-or-leave.**

FYI/decision: with the shipped data this branch never decides anything — the fact path below already accepts all six exact/allowlisted strings (verified by deleting this short-circuit and re-running), and it's prefix-insensitive, which also makes the `_READBACK_PREFIXES` stripping loop a no-op. It's only reachable for a hypothetical resident with empty `required_facts`. Either keep it as that documented fallback, or drop `accepted_readbacks` + the prefix loop (~10 lines plus per-resident config that must track the orders forever). The PR description's ‘exact or explicitly allowlisted complete readback’ describes the dead path either way.

sim/challenges/40_household_orders/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+
+from mars_sim_driver.challenges import Challenge, Drop, EventSeen, Goal, SkillDone
+
+from .runtime import HouseholdOrdersRuntime, Resident
+
+# Stable American-English Cartesia voices recommended for conversational
+# agents. Blake's scan is masculine; Alex's and Casey's are feminine.
+MASCULINE_VOICE_ID = "a5136bf9-224c-4d76-b823-52bd5efcffcc"  # Jameson
+FEMININE_VOICE_ID = "f786b574-daa5-4673-aa0c-cbe3e8534c02"  # Katie
```
theo-michel, line 16: **Polish — except the test assertion, which is worth flipping.**

Alex and Casey share this voice, so two of the three characters in a spoken scenario are aurally identical — the `“Speaker: …”` prefix that distinguishes them in chat is stripped from the audio. Needs someone with the Cartesia roster to pick a second stable feminine voice (I didn't want to guess an id). Note `test_resident_assets_and_voices_are_distinct_where_expected` currently **asserts** the collision (`alex.voice_id == casey.voice_id`), so it needs flipping to all-distinct in the same change.

axelpey:
Fixed the Alex orientation regression in `d8c0dfd6`.

The recent review fix changed the challenge yaw for Alex from `0°` to `-90°`, based on the assumption that all normalized resident scans needed that room-facing adjustment. For Alex, this rotated the rendered model 90° sideways relative to the intended orientation. Her yaw is now restored to `0°`; the resident asset itself was not changed.

Validation: `tests/test_household_npc.py` passes (21 tests). The correction has also been propagated through the downstream stacked PR branches.

Separately, the navigation failure observed during manual verification was caused by Nav2/ROS timing degradation (missed controller and behavior-tree deadlines plus stale TF), not by this challenge setup change or an unreachable resident placement.

axelpey:
@greptileai

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,149 @@
...
+
+        for alternatives in resident.required_facts:
+            polarities = [HouseholdOrdersRuntime._phrase_polarities(normalized, phrase) for phrase in alternatives]
+            if not any(positive for positive, _negative in polarities):
+                return False
+            # Alternate spellings describe the same fact. A positive mention of
+            # one spelling must not hide a later contradiction using another
+            # spelling (for example, "ShackBurger ... but no Shack Burger").
+            if any(negative for _positive, negative in polarities):
+                return False
```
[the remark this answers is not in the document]
axelpey: Fixed in `1a75fd1b`. Required-fact negation detection now accepts ordinary intervening modifiers such as `more`, `extra`, `additional`, `any`, `of`, and `the`, so a readback such as `no more brown rice` cannot emit the latched confirmation event. Added the reported phrase as a regression case in `test_incomplete_or_contradictory_readbacks_do_not_confirm`; the focused household suite now passes all 22 tests.

axelpey:
@greptileai

sim/challenges/40_household_orders/runtime.py
```diff
@@ -0,0 +1,176 @@
...
+import re
+from dataclasses import dataclass
+
+from mars_sim_driver.challenges import ChallengeRuntime, EnvironmentReply, RuntimeResult, WorldState
+
+_NON_WORD = re.compile(r"[^a-z0-9]+")
+_READBACK_PREFIXES = ("actually ", "okay ", "so you want ", "you want ", "your order is ", "you said ")
+_NEGATION_PREFIX = re.compile(
+    r"(?:^|\s)(?:not|never|no|without|skip|hold|omit|remove)"
+    r"(?:\s+(?:additional|any|extra|more|of|the))*\s*$"
```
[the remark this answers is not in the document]
axelpey: Fixed structurally in `6acba202`. The closed modifier allowlist has been removed. Negation is now evaluated over the phrase clause, with punctuation and coordinating words preserved as scope boundaries, so `no longer brown rice` and arbitrary intervening modifiers are rejected without treating the next comma-separated item as negated. Required-fact groups also reject a negated occurrence of any alias even when another alias appears positively. Added regressions for `no longer brown rice`, the prior `no more brown rice` case, and contradictory aliases. Full simulator suite: 82 passed.

axelpey:
@greptileai

PR #639, opened 2026-08-11
feat(voice): ElevenLabs batch STT with on-robot Silero VAD — field-validated defaults

theo-michel:
> **Part 1 of 2** — batch speech-to-text end to end.
> **1. innate-os #639 (this PR)** — the batch STT engine, on-robot VAD, and the proxy-only credential path
> 2. innate-inc/innate-cloud#83 — lets the proxy forward the audio upload untouched

## Summary

- Speech-to-text now defaults to **ElevenLabs Scribe batch** (`POST /v1/speech-to-text`, `scribe_v2`) instead of the Scribe realtime session. Batch transcribes far more accurately on this hardware (see the benchmark below — Scribe realtime *hallucinates* on real far-field speech), and realtime's early-partial advantage doesn't exist here: the first partial lands ~2.2 s in regardless of clip length, i.e. *after* the audio ends on a short utterance.
- Batch has no server-side VAD, so endpointing moves onto the robot: **Silero VAD** (`stt_vad_engine: silero`, vendored ONNX, energy fallback) drives an `Endpointer` that opens on speech and closes after `stt_vad_silence_secs` of quiet; `BatchSttSession` ships the clip as WAV from a worker thread.
- The vendor call is a transcriber closure (WAV bytes → text), so the same session also serves a **Gemini** `generateContent` backend.
- `stt_backend` still selects `elevenlabs` / `openai` to return to the realtime sessions without a redeploy.

| `stt_backend` | Path | Model knob |
|---|---|---|
| `elevenlabs_batch` *(new default)* | batch HTTP | `elevenlabs_batch_stt_model` (`scribe_v2`) |
| `gemini` | batch HTTP | `gemini_stt_model` (`gemini-3.6-flash`) |
| `elevenlabs` | realtime WS | `elevenlabs_stt_model` (`scribe_v2_realtime`) |
| `openai` | realtime WS | `openai_realtime_model` |

## Hardware validation (2026-08-12, real robots)

The engine choice and every default below were measured on real audio — 50 human utterances recorded on a MARS mic (hand-annotated ground truth) plus a speaker→mic acoustic corpus replayed across a real air gap at a ladder of SNRs.

**Engine ranking on real human speech** (accuracy = 100 − mean WER, n=50; run-to-run noise is ~±5 pts, so only large gaps are meaningful):

| engine | accuracy | median round trip |
|---|---|---|
| **Scribe batch (this PR's default)** | **~70%** | ~600 ms |
| `gpt-transcribe` | 62% (returns *empty* on 60% of quiet clips) | 380 ms |
| `whisper-1` | 61% | 1.2 s |
| `gpt-4o(-mini)-transcribe` | 57% | 320–390 ms |
| `gemini` batch | accurate but 2–15 s round trips | — |
| **Scribe realtime (the engine this PR retires)** | **34% — invents fluent unrelated text on far-field audio** | 166 ms tail |

**VAD**: silero vs the energy fallback is not close on real audio — at 3 dB SNR silero detects ~10× more speech frames (42% vs 4%) with zero false triggers measured anywhere; the energy threshold is knife-edge on the room floor (0.005 merges everything, 0.01 is right for a ~−46 dBFS room, 0.02 collapses recall). `stt_energy_threshold: 0.01` is hereby **verified for quiet rooms** — but it is the fallback for a reason.

**Defaults changed in this PR after measurement** (commit 2b0c48df):
- `stt_vad_threshold` **0.3 → 0.2** — 100% recall down to ~3 dB SNR, still zero false opens; 0.3 misses quiet speech the mic demonstrably captures.
- `stt_vad_silence_secs` **0.7 → 0.5** — the silence timer runs through speech's own trailing quiet, so 0.5 closes just +66..+183 ms after the true end of speech, while 0.7 adds ~200 ms of dead air to *every* turn. 0.3 was tested and clips final words — don't go lower.

**Known, accepted**: `resample_24k_to_16k` is a 3:2 linear decimator with no stopband — a 9 kHz tone folds to 7 kHz ~27 dB hotter than a polyphase resample would allow. Measured effect on Silero decisions is ≤3.4 pp, so it stays simple; noted here so nobody rediscovers it.

### Two details worth a reviewer's eye

**Pre-roll.** 400 ms of audio before the detection crossing is retained — the crossing lands mid-syllable, so without it the first word is clipped.

**Elapsed time is counted in audio fed, not wall clock.** Ducking pauses the stream while the robot speaks; if the endpointer measured wall clock, that pause would read as silence and close an utterance the speaker is still in the middle of. Pinned by `test_ducking_gap_does_not_count_as_silence`.

## Reaching ElevenLabs

Uses `ELEVENLABS_API_KEY` when the robot has one, and falls back to the proxy passthrough otherwise — which is why `request_stream` grew multipart support. **The proxy fallback is now verified end-to-end from a robot**: `elevenlabs_proxy_transcriber` → deployed service proxy → ElevenLabs returned correct transcripts 3/3 at 660–1823 ms (~warm 700–900 ms; the hop costs ~100–300 ms over the direct path). Requires innate-cloud#83 for the upload to be a first-class path rather than a lucky fallback — deploy that first.

## Validation

- [x] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up`. — n/a.
- [x] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`. — n/a.

- `ruff check` + `ruff format --check` clean.
- **254 `brain_client` tests pass** (endpointer state machine, multipart wire shape against the real client, proxy request shape, session surviving HTTP 500 mid-stream).
- Live on-robot: proxy batch path 3/3 (above); Silero endpointing validated across 4 TTS voices × volume ladder over a real air gap (8/8 utterances, zero false triggers).


DavidDobas:
Pushed three additions on top of the batch-API move (tested live on mars-the-47th):

- **Silero VAD v6.2.1** (vendored ONNX, onnxruntime CPU, ~0.7ms/window on the Orin) replaces the RMS threshold as the batch endpointer's voiced decision, behind a new `stt_vad_engine` setting (`silero` default, `energy` fallback — automatic if the model can't load). Side-by-side on the same stream (1s silence + 3s noise at 5× the energy threshold + 2.8s TTS speech): energy shipped one 6.7s utterance merging noise+speech; Silero shipped 3.7s of speech alone, Scribe returned it verbatim.
- One interface landmine, now guarded by a committed 2.8s speech-clip regression test: the Silero graph silently accepts bare 512-sample windows and scores real speech at 0.003 — it needs 64 samples of previous-window context prepended (576 total), which scores 1.000. No synthetic signal catches this; Silero scores tones/noise near zero either way.
- **Voice-input strip on the brain page**: MicroInput emits `vad_status` at 5Hz on `/input_manager/custom` (engine, live level vs threshold, voiced/utterance/ducking, counters); the panel renders a live meter with threshold tick plus the last transcript, and dims when telemetry stops. Also caught up the settings catalog with this PR's backends (the dropdown couldn't select `elevenlabs_batch`).

Also merged origin/main in (#640's unified brain page is the base for the panel).

ros2_ws/pip-requirements.txt
```diff
@@ -27,6 +27,8 @@ einops
 dotenv
 # onnx: exports the ACT graph when building the TensorRT engine (manipulation/act_trt.py)
 onnx
+# onnxruntime: runs the vendored Silero VAD (brain_client/inputs/vad.py)
+onnxruntime
```
DavidDobas, line 31: Nice

ros2_ws/src/brain/brain_client/brain_client/inputs/batch_stt.py
```diff
@@ -0,0 +1,315 @@
...
+            if utterance is None:
+                return
+            try:
+                text = self._transcriber(pcm_to_wav(utterance, self._sample_rate))
+            except Exception as e:  # noqa: BLE001 — one failed call must not kill the mic
+                self.failure_count += 1
+                self._logger.error(f"❌ Batch transcription failed: {e}")
+                continue
+            if text:
+                self._on_transcript(text)
```
DavidDobas: Would be safer to have inside try

def _transcribe_loop(self) -> None:
    while True:
        utterance = self._utterances.get()
        if utterance is None:
            return

        try:
            text = self._transcriber(
                pcm_to_wav(utterance, self._sample_rate)
            )
            if text:
                self._on_transcript(text)

        except Exception as e:  # noqa: BLE001
            self.failure_count += 1
            self._logger.error(f"❌ Batch transcription failed: {e}")

ros2_ws/src/brain/brain_client/test/test_batch_stt.py
```diff
@@ -0,0 +1,343 @@
...
+    # utterance must stay open no matter how long the gap.
+    ep = make_endpointer(silence_secs=0.3)
+    feed_seconds(ep, LOUD, 0.5)
+    assert ep.feed(LOUD) is None  # still open after an arbitrary pause
+
+
+def test_energy_detector_survives_a_partial_frame():
+    """arecord's pipe is raw (bufsize=0), so a read can return an odd byte count.
+    The reader realigns, but int16 parsing must not explode if one slips through."""
+    assert EnergyDetector(0.01)(b"\x01" * 333) in (True, False)
```
DavidDobas: Check if all tests are useful

DavidDobas (approved):
Just a few small comments

ros2_ws/src/brain/brain_client/test/test_batch_stt.py
theo-michel: Yeah probably not you are right

PR #644, opened 2026-08-11
tune(control): halve heading-hold memory to 0.1 s

DavidDobas:
Makes heading hold's memory (`heading_hold.leak`) 0.1 s by default, down from 0.2 s.

## Why

The hold was counter-steering after a manual turn: turn left, recenter the stick, and the robot swings back right.

The cause is when the hold re-latches. During a turn it releases outright, and it re-latches on the **requested** yaw hitting zero — which is the instant the stick recenters. The robot is still rotating at that point: the published yaw rate ramps down over ~150-300 ms (0.1 s lag plus the 6 rad/s² angular decel), and the chassis and caster coast past even that. So the hold latches a heading, the robot rotates through it, and that residual becomes an error it steers back out.

`leak` is the time constant on which the latched target chases actual yaw. At 0.2 s the target barely moves during the coast-out, so nearly all of it gets corrected away. At 0.1 s the target follows twice as fast and most of it is forgiven instead.

## Tradeoff

Steady-state drift against a persistent disturbance is roughly `disturbance / (1 + gain * leak)`. At the default gain of 5, that goes from halving the drift to cutting it to about two-thirds. Rejection is weaker but still there.

This is a mitigation, not the root fix — the latch timing is what actually produces the counter-turn. Latching on the *smoothed* angular settling would remove it outright; the header notes that was tried and rejected because it left the hold disengaged for the whole ramp-down. Worth revisiting if 0.1 still counter-turns on hard flicks.

## Changes

Three places state the default and all move together:

- `drive_smoother.hpp` — `HEADING_LEAK`, the compiled default `mars_app` declares the parameter with
- `config/settings.yaml.template` — the commented example
- `webapp/js/settings/catalog.js` — the Driving → Heading hold "Memory" knob's default, which is what the UI shows as unset

The knob is live-tunable, so any robot with an explicit `heading_hold.leak` in its settings keeps its own value.


theo-michel (approved):
LGTM

PR #645, opened 2026-08-11
refactor(sim): drop the cloud-agent clone and its launcher plumbing

karmanyaahm:
The brain runs in-process inside `brain_client` now — `scripts/launch_sim_in_tmux.zsh` already ignores `--brain-websocket-uri` as obsolete — so the sim launcher was the last thing still cloning, building, and pointing the robot at an external agent nobody connects to.

## What `./innate-sim setup` does now

The key prompts stay (innate-os itself needs them), reworded to say what is actually true: the agent loop always runs on the robot, and the choice is only *how it reaches Gemini* — your own Gemini key straight to Google, an Innate service key through the proxy (adds voice), or none. No clone, no lock alignment, no `[cloud_agent]` config.

## Removed

- **`setup_wizard.py`** — `ensure_cloud_agent_repo`: the clone, the pinned checkout, the drift prompts.
- **`config.py`** — the `hosted`/`local-image`/`local-source`/`auto`/`none` mode machinery, `BRAIN_WEBSOCKET_URI` resolution, `build_cloud_env`, the cloud-agent log target, `cloud_repo`/`cloud_port`/`cloud_image`. In their place `resolve_brain_backend` just reports which key is set — service key wins, matching `brain/transport.py:pick_transport`, so the dashboard cannot lie about the backend.
- **`runtime.py`** — `start_cloud_agent`, the git lock/status/checkout helpers, the 8766/8767 Foxglove port shuffle (the bridge is back on 8765 unconditionally), the `BRAIN_WEBSOCKET_URI` compose value and launch flag, `capture_agent_logs`, the "Cloud agent" startup check.
- **`dashboard.py`** — the AGENT LOGS pane and its worker thread, the "Cloud mode" / "Local agent" / cloud-repo lines. The `Agent:` field becomes `LLM: innate proxy | gemini key | no key`.
- **elsewhere** — the `cloud-agent` compose service and `local-brain` profile, `sim/cloud-agent.lock`, the `[cloud_agent]` template section, the `.gitignore` entry, `./innate-sim logs cloud-agent`.

## One thing kept deliberately

`remove_legacy_cloud_agent()` runs on `up`, `down`, and `clean`. The old container is no longer a service of this compose project, so `down` cannot see it, and it holds host port 8765 — which the Foxglove bridge now publishes on. It logs a line only when it actually removes something.

## Verification

- `ruff check` + `ruff format --check` clean on `sim/launcher/`
- 38 repo tests pass (`tests/`, including `test_launcher_imports` over every launcher module)
- `docker compose -f sim/docker-compose.dev.yml config -q` validates
- exercised by hand: `./innate-sim setup </dev/null`, `status`, `logs --help`, plus the dashboard render and `dashboard_runtime` paths

Robot-side `BRAIN_WEBSOCKET_URI` mentions (`scripts/update/post_update.sh`, `docs/PARAMETERS.md`) are left alone — this change is scoped to the sim.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -5,7 +5,7 @@
 # simulator settings in sim/config.toml.
 
 # --- Secrets ---
-# INNATE_SERVICE_KEY (hosted brain mode) is written to your local .env by
+# INNATE_SERVICE_KEY (Innate proxy access) is written to your local .env by
 # `./innate setup` / `./innate-sim setup` and seeded to /etc/innate.env on update —
 # it is never kept as a line in this template.
 #
--- a/.gitignore
+++ b/.gitignore
@@ -7,10 +7,6 @@ log/
 scripts/build/
 ros2_ws/src/brain/manipulation/ckpts/*
 
-# Cloud-agent source lives OUTSIDE this repo (cloned next to it by `innate-sim setup`).
-# Ignore any stray in-repo copy so it never gets committed.
-sim/innate-cloud-agent/
-
 # Python artifacts
 __pycache__/
 *.py[cod]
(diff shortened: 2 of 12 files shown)


DavidDobas (approved):
LGTM

PR #647, opened 2026-08-12
fix(sim): warn when the daemon is a Docker 29 that cannot mount images

karmanyaahm:
Closes INN-789.

## The short version

INN-789 is real but its conclusion was wrong. This is not "Docker 29" — it is
**29.0.0 – 29.1.3**, an upstream bug ([moby#51687](https://github.com/moby/moby/issues/51687))
that was fixed in **29.1.4** back in January. So the launcher gets a diagnosis,
not the proposed short-alias workaround.

## What actually breaks

Docker 29.0.0 ([PR #50268](https://github.com/moby/moby/pull/50268)) began naming an image
mount's layer directory after the **hex** of the whole `<digest>,src=<ref>,dst=<target>`
spec — twice its length, against `NAME_MAX = 255`, so the spec must be ≤ 127 chars.

Fixed overhead is 74 (64-char digest + `,src=` + `,dst=`), leaving ~53 for ref + target.
Our asset ref is 111 chars (`ghcr.io/innate-inc/innate-os-sim-assets:` + `inputs-` + 64 hex)
and the spec runs 218, so `./innate-sim up` dies at container create with `file name too long`.

Fixed upstream by [#51827](https://github.com/moby/moby/pull/51827) /
[#51829](https://github.com/moby/moby/pull/51829) (hash the spec instead of embedding it),
milestone 29.1.4.

## Why a warning and not the workaround

The issue proposed making the launcher tag a short local alias and mount that. It works
mechanically, but it is permanent complexity — an explicit `docker pull` the launcher does
not do today, plus tag management and offline-path changes — to route around a bug that no
supported Docker has. There is also no partial fix available: the two viewer targets leave
~53 chars for digest + ref, and even the `:main` default is over budget. Shortening is not
a tuning knob.

So the preflight names the bug and the version that fixes it, and gets out of the way.
Warning rather than hard stop: every other verb works on these daemons, and a stale patch
release is not the launcher's call to make. On Linux the remedy mentions that a distro
package can be stuck on a broken patch with nothing newer to upgrade to — Ubuntu's
`docker.io` in particular — so the escape is switching to Docker's own repo.

Deliberately not folded into `_require_min_version`: that only parses major.minor, and a
major.minor check would warn every 29.1.x user forever.

```
[warn] Docker Engine 29.1.3 cannot mount the sim viewer's assets.
      29.0.0-29.1.3 fail every `type: image` mount with 'file name too long'
      (moby#51687), so `./innate-sim up` will die at container create. No launcher-side
      workaround exists -- the mount spec is over budget even for the shortest ref.
      Update Docker Engine to 29.1.4 or newer. Ubuntu's docker.io can sit on a
      broken patch with nothing newer to upgrade to -- if `apt` offers none, switch to
      Docker's own repo: `curl -fsSL https://get.docker.com | sudo sh`
      Details: https://github.com/moby/moby/issues/51687
```

## Verification

Real daemons in dind, the 111-char asset ref vs. a short alias, both image stores:

| Docker | overlay2 | containerd snapshotter |
|---|---|---|
| 28.4.0 | PASS | PASS |
| 29.0.0 | **FAIL** | — |
| 29.1.3 | **FAIL** | **FAIL** |
| 29.1.4 | PASS | PASS |
| 29.7.1 | PASS | PASS |

The short alias passed in every cell, including both failing ones — the fix mechanism was
sound, it is just unnecessary now.

Note this corrects INN-789's "same storage driver and snapshotter" reasoning: 29.1.3 fails
on **both** stores and 29.1.4 passes on both, so the image store never discriminated
anything. Only the version does. It also inverts "every dev breaks as they upgrade" —
upgrading past 29.1.3 is the fix.

`tests/test_broken_image_mount_versions.py` pins both ends of the window and the
unparseable-version silence. `pytest tests/` (46 passed) and `ruff check` / `ruff format
--check` are clean.

### DIFF
--- a/.github/workflows/launcher-tests.yml
+++ b/.github/workflows/launcher-tests.yml
@@ -1,0 +1,27 @@
+name: Launcher Tests
+
+# tests/ guards host-side, no-ROS invariants (launcher preflights, image-tag
+# hashing, workflow path filters); nothing else in CI runs pytest, so without
+# this job the suite only runs when a developer remembers to.
+on:
+  push:
+    branches: [main, develop]
+  pull_request:
+    branches: [main, develop]
+
+jobs:
+  pytest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          # 3.12 for stdlib tomllib; the launcher itself still supports 3.10+tomli.
+          python-version: "3.12"
+
+      - name: Run the host-side test suite
+        run: |
+          pip install pytest pyyaml
+          python -m pytest tests/ -q
--- a/sim/launcher/runtime.py
+++ b/sim/launcher/runtime.py
@@ -292,6 +292,51 @@ def ensure_docker_available(*, command_hint: str = CLI_SIM, require_compose: boo
         command_hint,
         COMPOSE_INSTALL_URL,
     )
+    _warn_broken_image_mounts(result.stdout, command_hint)
+
+
+# Docker 29.0.0 started naming an image mount's layer directory after the hex of
+# the whole `<digest>,src=<ref>,dst=<target>` spec -- twice its length, against
+# NAME_MAX=255. Our refs are content-addressed (`inputs-` + 64 hex), so the spec
+# is 218 chars and `up` dies at container create with "file name too long".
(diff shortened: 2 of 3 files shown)


theo-michel (approved):
LGTM

PR #653, opened 2026-08-12
fix(audio): ship the boot sounds as 48 kHz PCM, not MP3

theo-michel:
**Stack** — merge bottom-up:

| | PR | |
|---|---|---|

## Problem

Follow-up to #652. The startup and shutdown sounds play in the two windows most likely to starve a player — the cold-boot import storm and the service-teardown storm. #652 stopped the player from being *preempted* (SCHED_FIFO); this removes the work it has to do *while* it's running.

Decoding MP3 is CPU the player spends inside that window, and the assets weren't at the mixer's rate either (`turnon` 44.1 kHz, `turnoff` 24 kHz), so ALSA also resampled them up to dmix's 48 kHz.

## Fix

Both sounds converted to **48 kHz / 16-bit / stereo PCM WAV** — the rate dmix already runs at, so playback is a straight read off disk into the mixer: no decode, no resample.

The WAVs (~1.4 MB combined) are **not in git**: they live in `gs://karmanyaah-public/sounds/` and `post_update.sh` downloads them into `config/sounds/` (step 11c, same skip-if-present curl pattern as the skill assets in 11b). A failed download logs a warning and means a silent boot — never a failed update. The MP3s are deleted; nothing else referenced them.

| file | before (git) | after (GCS) |
|---|---|---|
| `turnon` | 138 KB mp3, 44.1 kHz | 1.3 MB wav, 48 kHz |
| `turnoff` | 19 KB mp3, 24 kHz | 171 KB wav, 48 kHz |

Known limitation (shared with 11b): skip-if-present means a future sound *change* needs a new filename or a manual delete on robots that already downloaded it.

## Verification

- Both objects uploaded and public: `curl -I https://storage.googleapis.com/karmanyaah-public/sounds/turnon.wav` → 200 (same for `turnoff.wav`).
- Levels unchanged by conversion: `turnon` peak -20.7 dB before and after, `turnoff` -7.4 → -7.2 dB. Durations shorten ~30 ms — MP3 decoder delay/padding going away, not audio lost.
- **Not yet verified on a robot.** Wants one `innate-update` run (confirms the download lands), then a cold boot and a poweroff for the pop.

## Rollout

Robots pick this up on the next `post_update.sh` run — it both downloads the WAVs and reinstalls the shutdown unit.


DavidDobas (approved):
Great

PR #656, opened 2026-08-13
Remove the No Prompt agent; demo_agent becomes the default

theo-michel:
## Summary

Removes the idle **No Prompt** agent (`EmptyDirective` / `empty_directive`) from the OS — no traces left:

- `workspace/innate_agents/empty_directive.py` deleted.
- `initialize_agents` now prefers `demo_agent` as the boot default (same structure as before: falls back to the first loaded agent when it's absent).
- The simulator-mode auto-activation no longer special-cases the idle id: with no idle agent left, sim mode auto-activates whatever default directive is selected.
- Tests updated: `test_empty_directive_is_default` → `test_demo_agent_is_default`.

Docs counterpart: https://github.com/innate-inc/docs/pull/27 — removes the No Prompt card and presents `demo_agent` as the default in the introduction.

## Behavior changes to be aware of

- **Boot default**: robots now boot with `demo_agent` selected (previously the idle No Prompt agent). Nothing runs until the brain is activated — `is_brain_active` still gates that — so on hardware the only visible change is the pre-selected agent in the app.
- **Simulator**: sim mode previously skipped auto-activation when the default was `empty_directive` (i.e. always, on a stock workspace). It now auto-activates the demo agent at boot, so `./innate-sim up` with a brain configured starts the agent loop immediately.

## Verification

- `ruff check` + `ruff format --check` clean.
- Full brain_client suite: **227 passed** (run in `innate-os-test:fullbuild` with `mars_msgs`/`brain_messages` rebuilt as an overlay — the image's baked messages are stale).
- `basedpyright brain_client/`: error set identical to `origin/main` (all pre-existing, from the stale local `ros2_ws/install`); zero new diagnostics.


DavidDobas (approved):
Great

PR #657, opened 2026-08-13
feat(settings): pick the driving sound from the settings page

karmanyaahm:
**Stack** — merge bottom-up:

| | PR | |
|---|---|---|

> Stacked on #635 — base is `feat/accel-voices`, which is itself stacked on #653. This diff is only the settings-page wiring; merge #635 first and this retargets cleanly.

## What

The Voice page becomes **Sound**, and gains a **Motor sound** picker above the TTS voice. Choosing one writes `motor_sound.voice` to `config/settings.yaml` and pushes it to `/motor_sound` with `set_parameters`, so it takes effect on the next drive without a restart — #635's node rebuilds its voice from a parameter update and ramps the new voice's gain from zero.

## The list comes from the robot

The catalog does **not** name the 35 voices. The node declares them as a read-only string-array parameter built from the registry:

```py
("motor_sound.voice_options", ["motor", *VOICES],
 ParameterDescriptor(read_only=True, description="Accepted values for motor_sound.voice")),
```

and the page reads it with `get_parameters` over the rosbridge socket it already holds for the volume slider. Add a voice in `accel_voices.py` and it is in the UI with no webapp change at all.

Nothing new is served for this: no HTTP endpoint, no proxy code, and no import of `mars_control` into the webapp process — so the list cannot go stale against a workspace that was not rebuilt, because it comes from the node that is actually running. Labels are derived from the values (`spoke_card` → "Spoke Card"), which is what keeps the webapp from having to know what a spoke_card is. The cost of that is the hand-written names and the per-voice blurbs, which a `ParameterDescriptor` has nowhere to put.

**With rosbridge down or the node not up**, the picker keeps only its "Custom…" entry and shows the saved value in the free-text field, still editable and saveable. That is deliberate: a static fallback list would reintroduce exactly the drift this design exists to kill. The node accepts an off-list name too — it logs a warning and falls back to the motor whine (`motor_sound.py:140`).

## Machinery

`optionsFrom: {node, param}` is one new knob property and ~30 lines in `main.js`: a per-entry `setOptions` that mutates the captured options array in place (so the "is this a stock option" test stays honest) and re-renders, deferred behind `onStateChange` because on a cold load the socket is still opening when the page builds. The re-render is load-bearing — a saved override like `purr` arrives from `/settings.json` before the list does, and without it the row would sit in the Custom text field showing a name that is on the list.

## Test plan

- [x] `pytest tests/ -q` — 51 pass, including a new drift guard: the catalog keeps `options: []` + `optionsUrl` (no static list creeping back), and its default matches the node's declared default and is a real voice
- [x] Endpoint served and fetched over aiohttp: 200, 35 options, `Cache-Control: no-cache`
- [x] **Rendered in Chrome** against a local harness: 36 entries (35 + Custom…), `motor` first, blurbs as tooltips, no console errors
- [x] A saved override of `purr` — delivered before the parameter read lands — ends up selected with the Custom field hidden, which is the case the re-render exists for
- [x] `ruff check` / `ruff format --check` clean on the node and the test; `catalog.js` and `main.js` import cleanly under node
- [ ] Click it on a robot: pick `gallop`, drive, confirm it applies without a restart and survives a reboot


tests/test_motor_voice_setting.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright (c) 2026 Innate Inc
+"""Drift guard for the settings page's motor-sound picker.
+
+The picker's choices come from the robot -- the node declares them as a read-only
+parameter built from ``VOICES``, and the page reads that over rosbridge -- which is
+the whole point. What is left to drift is the three constants around it: the
+catalog's default, the node's declared default, and whether that name is a voice at
+all. Pure text parsing, no numpy and no ROS, because CI has neither.
+"""
+
```
DavidDobas: Consider whether these tests are essential

DavidDobas (approved):
Good

PR #660, opened 2026-08-14
feat(sim): report anonymous simulator usage

DavidDobas:
The reporting side of sim usage tracking. Robot behaviour is unchanged — `logger_node` is untouched.

## Why

The sim already ships the `innate_logger` package and colcon already builds it. It is simply never launched. But that node is also the wrong node for the sim: it reports battery and diagnostics that do not exist here, it *requires* a service key most sim users do not have (`raise RuntimeError` — a crashed tmux pane), and it blocks on auth at startup. That last one is correct for a robot, whose cold boot will eventually get DNS and a valid clock. A sim user on a plane is not a transient.

## What

**`sim_logger_node`** — a separate node, deliberately not shared code with `logger_node`.

- **Nothing blocks the executor.** Callbacks only enqueue; all auth, HTTP and backoff live on the uploader thread. (The existing node does a synchronous 5 s `urlopen` from a timer callback, which is tolerable on a robot and not here.)
- **Offline is an ordinary state.** Backs off to 5 minutes instead of retrying every heartbeat, and logs at debug — an unconnected sim spamming "telemetry unreachable" every 20 s reads to a user as though the simulator is broken.
- **Events spool to disk; heartbeats do not.** A stale "alive 3 days ago" says nothing that a gap in the series does not already, and would crowd out the events worth keeping. The spool survives container restarts.
- Works keyed (`AuthProvider`) or anonymous (mints against the install id).

**Identity and version come from the host launcher** via the generated `.env`, because none of it is knowable inside the container:

- It has no `.git` — compose mounts subdirectories, so `git rev-parse` in there returns `"unknown"` for *every* sim user today.
- Docker randomises the container MAC per run, so a MAC read in there would inflate the device count rather than establish it.

So `sim/launcher/identity.py` hashes the **OS machine id** on the host instead (`IOPlatformUUID` / `/etc/machine-id`): stable across re-clones, and — unlike a MAC's ~48-bit space — not brute-forceable back out of the digest. Raw identifiers never leave the machine. `install_id` (per checkout) and `device_hash` (per machine) together answer both "how many installs" and "how many actual people".

**Real-time factor** stands where battery does for a robot: observed `/odom` rate against the driver's nominal 30 Hz. It is the signal that says the sim is unusable on someone's machine — which, given the known WSL latency pain, is the number I would look at first.

**Challenge outcomes** reach the node over rosbridge (`/sim/telemetry_event`) rather than a shared spool file. The engine lives in the host-side world server, the node lives in the container, and `flock` across that filesystem boundary is not dependable. The world server already speaks rosbridge, so this reuses machinery that exists.

## Privacy

Announced on first run, documented in `.env.template`, disabled by `INNATE_TELEMETRY=0`. No prompts, no code, no raw hardware identifiers.

## Testing

Both risky paths exercised against a real HTTP server (`SimTelemetryUploader`, `SimAuth`):

- offline → 5 events spooled to disk, heartbeat correctly dropped, nothing delivered
- reconnect → spool drains, heartbeat delivered
- restart mid-outage → spooled event survives and is delivered on the next connection
- anonymous auth → mints once, caches, re-mints after `invalidate()`, sends the right `install_id`

`identity.py` verified on this machine (macOS): resolves a stable device hash, platform, commit and `git describe`, and honours the opt-out. Python/zsh syntax clean.

**Not yet run:** the node itself inside a live sim. Wants a `./innate-sim up` before merge to confirm the launch line, that `/odom` counting gives a sane RTF, and that the challenge bridge round-trips.

## Ships with

- innate-inc/innate-cloud#85 — tables and endpoints (**merge and deploy first**)
- innate-inc/dashboard `feat/sim-usage-view` — the Sim view

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -18,6 +18,17 @@
 # GEMINI_MODEL overrides the brain's model (default gemini-3.6-flash).
 # GEMINI_MODEL=
 
+# --- Usage telemetry ---
+# The simulator reports that it is running — an anonymous install id, the sim
+# version, your OS and CPU/RAM, how well the sim keeps up, and which features
+# you used (drove it, ran a skill, ran the agent, finished a challenge). It
+# never includes your prompts, your code, or a raw hardware identifier: the
+# machine id is salted and hashed before it leaves your computer. Without a
+# service key it is fully anonymous.
+#
+# Turn it off completely:
+# INNATE_TELEMETRY=0
+
 # --- Cloud endpoints (all read from the environment; set by default) ---
 # TELEMETRY_URL=https://logs.svc.innate.bot
 # UNINAVID_WS_URL=wss://uninavid-v1.svc.innate.bot
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ Innate OS is currently based on ROS 2, the reference framework for robotics oper
 - **[mars_nav](ros2_ws/src/mars_bot/mars_nav)** - Nav2-based navigation, SLAM mapping, and the mode manager that switches between `mapfree`, `mapping`, and `navigation`.
 - **[brain_client](ros2_ws/src/brain/brain_client)** - bridge to the Innate cloud brain, websocket client, skills action server, and user input manager.
 - **[manipulation](ros2_ws/src/brain/manipulation)** - records and replays manipulation demonstrations and runs learned or scripted manipulation policies.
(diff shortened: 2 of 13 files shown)


axelpey:
privacy breaching, forget about it

PR #665, opened 2026-08-18
feat(sim): talk to the robot in sim — the browser is its microphone

DavidDobas:
## Summary

- You can talk to the real robot, but not to the one in the sim. This adds a mic toggle to the Agent page in sim, and makes the browser the robot's microphone — the exact mirror of `/tts/audio`, where the browser is already its speaker.
- The sim was mute for two reasons, not one: the container has no capture device (`arecord` isn't even installed, so `MicroInput` died on `FileNotFoundError`), and `input_manager.launch.py` was never in the sim's tmux set, so there were no input devices at all.
- The Agent page publishes base64 PCM16 mono at 24 kHz on `/mic/audio`; `RosPcmStreamer` hands it to `MicroInput` in place of `ArecordStreamer`. Same queue-of-chunks surface, so the device drives either without knowing which it holds.

### Everything downstream is the robot's code

Silero VAD, the endpointer, the ElevenLabs/Gemini transcribers, TTS ducking, `chat_in`, and the VAD telemetry the Brain page's meter reads — all unchanged and shared.

There is **one** sim conditional in the diff, in the webapp composition root, next to how the battery readout and profiling panel are already gated:

```js
createAgentPanel(root, ros, agentState, { onView: setView, mic: !!config.simControls })
```

The panel never learns what a sim is; it receives `mic: true` as a capability. Nothing on the robot side branches on sim at all — the one new branch there asks about hardware, not environment:

```python
device = self._detect_audio_device()
if device is None and self.node is not None:
    mic = RosPcmStreamer(self.node, self.logger)   # no capture device — take PCM off the topic
```

A robot whose USB microphone died takes the same path.

### Choices worth reviewing

**Streaming, not push-to-talk.** A design where the browser records one blob and posts it straight to the vendor would have been ~150 fewer lines in the browser. Rejected: it's a second transcription path that bypasses `MicroInput` entirely, so sim voice would drift from the robot's, the Brain page's VAD meter would stay dead in sim, and Chrome's webm/opus isn't a documented Gemini audio input.

**An AudioWorklet, not the main thread.** `getUserMedia` yields float samples at the sound card's rate in 128-sample quanta; the VAD wants int16 on an exact 24 kHz grid. That conversion runs on the audio thread because the Agent page renders the sim with Three.js, and audio captured on the main thread drops out whenever a frame runs long. `micWorklet.js` is excluded from `tsconfig` — it runs in the AudioWorklet global scope, whose globals `lib.dom` doesn't declare.

**Advertise before publishing.** `/mic/audio`'s only subscriber appears when the agent starts, so a bare publish has no type to resolve against the graph. `RosClient.advertise` states it up front and replays on reconnect. If `rws` ignores the op we're no worse off than today.

**Ducking is doubled in sim.** On the robot, `/tts/is_playing` stops the mic while it speaks. In sim the *browser* plays the clip, so that flag drops before playback starts — mistimed. A browser-side gate pauses publishing while a clip plays. Extra sim-only behavior, but it exists to make the robot's behavior come out the same: don't transcribe yourself.

**Two quiet divergences closed.** `onnxruntime` joins the sim image — without it sim falls back to energy VAD while the robot runs Silero. And the input manager keeps its `ProxyClient` when no service key is set, since it still carries the STT settings and a Gemini-key-only sim can transcribe on a direct key; that path previously returned `None` and silently disabled the microphone.

## Validation

- [ ] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up`.
- [ ] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`. — no asset changes

Ran:

- `pre-commit run --all-files -c .config/pre-commit-config.yaml` — all 6 hooks pass.
- `brain_client` suite in a ROS container — 227 passed.
- The server chain against the real code, with a stub node: no capture device → `RosPcmStreamer` → subscribes `/mic/audio` → base64 round-trips → VAD reports voiced → endpointer opens and closes an utterance. Bad payloads dropped, subscription torn down on close.
- New `webapp/tests/micWorklet.test.js` — 6 passing, covering 24/48/44.1 kHz resampling with render-quantum seam continuity, clamping, chunk shape (`node tests/micWorklet.test.js`).
- `tsc --noEmit` at the exact pre-change baseline error count — no new type errors.

**Not run, and worth knowing before merge:** `./innate-sim up` from this checkout, and therefore the browser half in a real browser. There's no `getUserMedia` in a headless test, and the `innate-dev` container currently running here mounts a *different* checkout (`innate-os-sim`), whose stack predates `batch_stt` — so it could neither exercise nor validate this code. It did, however, reproduce the bug exactly: `Found 0 audio devices` → `arecord` → `FileNotFoundError`.

Two things to watch on the first real run: whether `rws` accepts the `advertise` op, and whether echo cancellation plus the TTS gate is enough to stop the robot hearing itself.


DavidDobas:
Pushed a fix for a bug this surfaced — **not sim-specific, and pre-existing on `main`**.

`micro` has not loaded on Python 3.10 (what the robot runs) since micro_input.py started importing `Transcriber = Callable[[bytes], str]` for its runtime annotations. On 3.10 a subscripted generic still satisfies `inspect.isclass`, so the alias reached `issubclass`, which rejects it — and the `TypeError` escaped `_find_classes` and voided the entire file:

```
ERROR input_manager_node: Error loading InputDevice from .../micro_input.py: issubclass() arg 1 must be a class
INFO  input_manager_node: Loaded 1 input devices: ['arm_vitals']
WARN  input_manager_node: No input device named 'micro' — the microphone toggle will not gate anything
```

Confirmed against `main` in a 3.10 container, not just this branch. It is invisible on 3.11+, where `isinstance(list[int], type)` is False.

Fixed in the loader rather than in micro_input.py: it scans user files under `workspace/`, so tolerating whatever ordinary Python leaves in a module namespace is its job — otherwise any authored device file that imports a type alias silently loses all its devices.

`test_input_loading.py` pins it, holding the same guarantee `test_agent_loading.py` states for agents: nothing silently vanishes. Verified it fails without the fix with exactly the log line above, and passes with it.

Suite is 238 passing; `pre-commit` clean.

DavidDobas:
Third commit, and like the loader one it is **pre-existing, not introduced here** — found while testing voice.

`speak_text_async` documents that utterances "play in order, one at a time". On the robot that holds, because `speak_text` blocks until `aplay` finishes the clip. In the sim it cannot: `_synthesize_to_topic` is finished the moment the clip is published, so the robot half can only serialize *synthesis*. A two-sentence reply reaches the browser a few hundred ms apart, and ttsAudio played each on arrival — so the robot talked over itself. A joke and its punchline came out simultaneously.

The brain-side log shows it plainly:

```
TTS start (16 chars): 'Yes, I hear you!'
TTS streamed to browser (16 chars, 138KB)
TTS start (19 chars): 'How can I help you'   <- 0.5s later, previous clip still playing
```

The queue belongs on the playback side, since that is the only side that knows when a clip has actually finished. Clips now play in arrival order, each starting when the last ends — restoring what the robot's speaker does. Bounded at four, because a hidden or muted tab must not bank a monologue and deliver it late; that is the same reasoning behind `replace_pending` dropping superseded speech rather than playing the backlog.

It lands in this PR rather than a separate one because it is in a file this PR already touches, and because a robot talking over itself is also a robot transcribing itself — it degrades the microphone feature directly.

Two related findings from the same session, both pre-existing and **not** code changes, noted here for whoever picks this up:

- The one-tab-plays election (`innate-tts-speaker` Web Lock) is per-origin and needs a secure context. The sim serves the same app on both `:443` and `:80`, so tabs on `https://localhost` and `http://localhost` each elect themselves and both play. On a plain-http LAN address `navigator.locks` does not exist at all and the tab always plays. Not fixable across origins — browsers isolate them by design.
- That gate fails silently. A stale background tab swallows all audio with no log anywhere, which is a rough thing to debug. Worth a one-time log line.

DavidDobas:
Greptile's failed-open finding was correct — fixed in `fix(inputs): a failed microphone open must not leave the device active`.

`InputDeviceManager._apply` clears the active flag only when `on_open` raises, and says so in its own comment:

> Roll back the flag: a failed open left as "active" would make every later open attempt a no-op (stuck half-open until restart).

`MicroInput.on_open` swallowed every failure instead, so a transcriber that could not connect — an expired key, or a proxy blip on the very activation that opens the mic — left the device marked active with a running capture source and no STT session behind it. Later `active_inputs` messages see it as already open and do nothing. One correction to the report: it is not stuck until *restart*, because deactivating the directive still calls `on_close`; it is stuck until the directive is toggled. The symptom is the same either way — you speak and nothing ever appears.

Failures now propagate, after `on_close` releases the source and socket so the retry starts clean. A missing proxy client raises for the same reason: a device that cannot work should not sit in the roster claiming otherwise.

Not hypothetical — an auth misconfiguration during testing put the stack in exactly this state.

Also in this push: the format check (a `ruff format` nit in the new test, which my local `pre-commit --all-files` skipped because the file was still untracked), and a trim of the tests. The worklet resampler tests are gone — nothing in CI runs `webapp/tests`, and a resampling bug is audible the first time you speak to the robot. What remains is the one loader regression, which CI does run and which guards a failure that shipped silently for weeks.

workspace/inputs/micro_input.py
```diff
@@ -133,38 +139,46 @@ def set_tts_playing(self, is_playing: bool):
...
-            # Connect via proxy
             self._connect_via_proxy()
-
         except Exception as e:
             self.logger.error(f"❌ Failed to start microphone: {e}")
-            import traceback
+            self.on_close()  # release the device/socket so the retry starts clean
+            raise
+
+    def _open_source(self):
```
ugtthis: can we make this method name more self documenting?

workspace/inputs/micro_input.py
```diff
@@ -133,38 +139,46 @@ def set_tts_playing(self, is_playing: bool):
...
+        """The started audio source: the machine's ALSA capture device, or a
```
ugtthis: are all these doc string lines really needed?

workspace/inputs/micro_input.py
```diff
@@ -658,6 +677,45 @@ def _on_transcript(self, text: str) -> None:
 # ========== Audio Streaming Helpers ==========
 
 
+class RosPcmStreamer:
+    """Streams microphone audio a client pushes over ROS on MIC_AUDIO_TOPIC.
```
ugtthis: if we need this comment maybe this implementation isnt as self documenting?

workspace/inputs/micro_input.py
```diff
@@ -133,31 +138,14 @@ def set_tts_playing(self, is_playing: bool):
...
-                device=detected_device if detected_device != "default" else "default",
-                sample_rate=DEFAULT_SAMPLE_RATE,
-                channels=DEFAULT_CHANNELS,
-            )
-
+            self.mic = self._open_source()
             self.logger.info(f"🎙️ Microphone started (rate: {DEFAULT_SAMPLE_RATE}, channels: {DEFAULT_CHANNELS})")
-
-            # Connect via proxy
             self._connect_via_proxy()
```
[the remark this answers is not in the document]
DavidDobas: Correct, and fixed in 6bb4c9cf. `_apply` clears the active flag only when `on_open` raises — its own comment says so — and MicroInput swallowed everything, so a failed connect left the device active with a running capture source and no STT session. Failures now propagate after `on_close` releases the source and socket.

One correction: it is not stuck until restart. Deactivating the directive still calls `on_close`, so it recovers on a directive toggle. The symptom is the same though — you speak and nothing appears.

Not hypothetical, an auth misconfiguration put the stack in exactly this state while testing.

workspace/inputs/micro_input.py
DavidDobas: Renamed to `_start_audio_source` in fe3e9ced — it starts the source and returns it, which the old name said nothing about.

workspace/inputs/micro_input.py
DavidDobas: No. Dropped it in fe3e9ced. With the better name the first line was restating the signature; the only line worth keeping was why the fallback keys on the `arecord` binary rather than on it listing no cards, and that has moved down onto the branch it explains.

That distinction is load-bearing, so it stays somewhere: a robot whose microphone did not enumerate this boot still has a working ALSA `default`, and keying on the shared `None` would have sent it to listen on `/mic/audio` instead — deaf, waiting for audio no browser is sending.

workspace/inputs/micro_input.py
DavidDobas: Fair — it said the same thing three ways. Cut to one line in fe3e9ced:

```python
class RosPcmStreamer:
    """ArecordStreamer's surface, fed by MIC_AUDIO_TOPIC instead of a capture device."""
```

If you want the interchangeability enforced rather than described, the next step is a small `PcmSource` Protocol (`queue`, `start`, `stop`) that both streamers satisfy and `_start_audio_source` returns — same pattern as `VoicedDetector` in batch_stt.py. Happy to add it, left out for now as prototype-stage weight.

ugtthis (approved):

PR #666, opened 2026-08-18
feat(sim): show the face of Mars as the default start view and default to agent mode

ugtthis:
Problem: 
- the default angle of Mars is seeing his back
- we introduce unnecessary friction just to see the face

Solution:
 - show a better intimate view of the face
 - this will also a prereq for the new onboarding where the first interaction a person will have is talking to Mars in sim
 - this also defaults to agent mode VS telop
 
 
 New default cam view:
new-angle


ugtthis:
opening a new PR

PR #667, opened 2026-08-18
feat(sim): show the face of Mars as the default start view and default to agent mode 

ugtthis:
Problem: 
- the default angle of Mars is seeing his back
- we introduce unnecessary friction just to see the face
- first interaction is telop when the surprising/alive interactions are found in agent 

Solution:
 - show a better intimate view of the face as the defualt view when sim loads up
 - this will also a prereq for the new onboarding where the first interaction a person will have is talking to Mars in sim
 - default to agent page to verbally talk with the robot
 
 
#  New default cam view:
new-angle


DavidDobas (approved):
LGTM

PR #668, opened 2026-08-18
fix(sim): mount config/ read-write so the Settings page can save

theo-michel:
## Problem

Saving from the webapp **Settings** page in the sim always fails:

```
Save failed: settings update failed: [Errno 30] Read-only file system: '/root/innate-os/config/.settings.wzd7c8f4.tmp'
```

`settings_store.apply_changes` writes `config/settings.yaml` atomically (tempfile in `config/` + `os.replace`), but the sim compose file mounts `../config` into the container **read-only** — a leftover from the original simulator import (#323), which predates the Settings page entirely.

## Fix

Mount `config/` `:rw`, like `ros2_ws/`, `workspace/`, and `scripts/` already are.

- `config/settings.yaml` is gitignored, so saves don't dirty the checkout.
- Persisting settings on the host matches existing behavior: `./innate-sim clean` deliberately keeps `config`, and the sim README already documents `config/` as a host-shared directory you can edit from either side.
- Everything else under `config/` only *reads* `settings.yaml` (bringup/nav launch files), so nothing depended on the `:ro`.

## Verification

On a running sim (`./innate-sim down && ./innate-sim up` to pick up the new mount — the launcher reuses a running container, so `up` alone doesn't recreate it):

```
$ docker inspect innate-dev ... | grep config
/Users/.../innate-os/config -> /root/innate-os/config rw=true

$ curl -sk -X POST https://localhost/settings.json -H 'Content-Type: application/json' -d '{"sets":[],"clears":[]}'
{"ok": true, "message": "saved"}
```

and `config/settings.yaml` on the host is rewritten (previously: the Errno 30 above).


DavidDobas (approved):
Good

PR #669, opened 2026-08-18
feat(brain): move the local brain and Gemini STT to 3.7 Flash

theo-michel:
Answers the question "are we on 3.7 yet?" — we weren't. The brain agent loop and the Gemini STT backend ran `gemini-3.6-flash`, and the `innate.gemini` skills helper was still on `gemini-3.5-flash`. All three now run [`gemini-3.7-flash`](https://ai.google.dev/gemini-api/docs/models/gemini-3.7-flash) (stable since August 2026).

**The upgrade is not free, and the interesting part is not latency.** 3.7 Flash dropped the `minimal` thinking level we ran at, and — with our current tools — it answers a person who speaks to it far less reliably than 3.6 does. Both problems are fixed here, but the second one costs something; see the tradeoff below.

## How this was measured

The harness drives the shipped code — `build_system_prompt`, tool declarations built from the 23-skill workspace roster, `GeminiContext`'s body assembly and SSE parsing — with two 640x480 sim camera frames (head + wrist), ~7.9k prompt tokens, streaming, against the real API. Only the model id and thinking level vary between conditions; conditions are shuffled per trial so drift and ordering can't favour one. Probes cover 5 user utterances x 3 history variants (none / model-called-wait / model-spoke).

## Finding 1 — 3.7 goes silent when the user speaks

Counting how often the robot answers a person talking to it (n=60 per condition; a miss = it called `wait` and said nothing — verified from raw parts, no text hidden behind `thought: true`):

| condition | answers the user |
|---|---|
| 3.6 @ `minimal` (before) | **58/60 (97%)** |
| 3.7 @ `low` | 34/60 (57%) |
| 3.7 @ default | 34/60 (57%) |
| 3.7 @ `high` | 13/30 (43%) |

**The thinking level is not the cause** — every level fails. The attractor is the `wait` tool: offered a no-op, 3.7 takes it instead of replying. Drop `wait` from a turn carrying user speech and 3.7 answers **29/30 at `low`, 30/30 at default**, with 3.6 unaffected (30/30 either way).

`build_tools` already did exactly this while a skill is running, for exactly this reason ("offered any no-op tool, the model calls it and goes silent") — this PR extends the rule to idle turns. With that in place the thinking level goes back to `low`, 3.7's fastest.

**The tradeoff, measured, not assumed:** `wait` is also how the robot ignores stray STT fragments. Dropping it on spoken turns takes noise discipline from 88% to 74% (n=24; the extra chatter is almost entirely the robot answering its own speech echoed back). Ignoring a real person two turns in five is worse than occasionally answering an echo — but if the team disagrees, the lever is one `if` in `build_tools`.

## Finding 2 — speed

Median / p90 of the full turn; *spoken reaction* is time to the first spoken token, which is when the robot starts talking:

| condition | idle turn | spoken reaction | thought tokens |
|---|---|---|---|
| 3.6 @ `minimal` (before) | 1.59s / 2.00s | **1.59s** / 1.81s | 0 |
| 3.6 @ `low` | 1.54s / 1.64s | 1.70s / 2.00s | 0 |
| **3.7 @ `low` (shipped)** | 1.93s / 4.03s | **2.46s** / 4.12s | 36 |
| 3.7 @ default | 1.93s / 2.11s | 2.47s / 2.90s | 108 |

3.7 is a modest latency regression, not a win: roughly +0.3s idle and +0.85s on a spoken reaction, because `minimal` is exactly what it removed. That is the price of the upgrade; nothing in the config buys it back.

The Gemini STT path keeps an explicit `low` (it can't use `minimal` either): 2.39s / 3.20s / 5.03s for short/mid/long utterances vs 2.51s / 3.16s / 5.50s on 3.6 at `minimal` — no regression.

## The Innate proxy needs no change

`GeminiAdapter` is model-agnostic by construction — the model lives in the caller's path or body, never in the adapter. Verified against the current `origin/main` proxy with a fake upstream (real adapter, real `services.yaml`):

| check | result |
|---|---|
| `main.py` endpoint validation accepts `v1beta/models/gemini-3.7-flash:{generate,streamGenerate}Content` | pass |
| native `generateContent` → upstream path untouched, key in `x-goog-api-key` | pass |
| native `streamGenerateContent` → `alt=sse` added, SSE relayed, `thinkingConfig` forwarded intact | pass |
| OpenAI-compat `/v1/chat/completions` → rewritten to `/v1beta/openai/chat/completions`, `model` kept as `gemini-3.7-flash` | pass |

No Live API usage anywhere in the repo (3.7 doesn't support it), so nothing else to adapt.

## Deliberately left behind

- `workspace/innate_skills/chess/` pins its own cascade (`gemini-3.1-pro-preview` → `gemini-3.6-flash`) and drives it with `thinking_budget`, which Gemini 3 docs no longer describe. Retuning that needs a board in front of the robot.
- `webapp/debug/reaction-latency.html` is a record of a measurement taken on 3.6 — left as-is rather than rewritten.
- `innate-cloud-agent` (the hosted brain) is still on `gemini-3.5-flash` / `gemini-3.1-flash-lite`. Separate repo, separate deploy.

## Verified

- `ruff check` + `ruff format --check` on `brain_client` — clean; `pre-commit -c .config/pre-commit-config.yaml` on the changed files — all hooks pass
- `pytest test/` in a ROS container — 237 passed (one new test covers the wait-tool rule)
- `basedpyright brain_client/` — 15 errors, all pre-existing in files this PR doesn't touch
- All probes ran against the real Gemini API over the direct path, so proxy latency isn't in the numbers. The harness is not checked in — say the word and it goes under `tools/` so the next model bump is one command.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -15,7 +15,7 @@
 # service key), talking to Google directly.
 # GEMINI_API_KEY=
 
-# GEMINI_MODEL overrides the brain's model (default gemini-3.6-flash).
+# GEMINI_MODEL overrides the brain's model (default gemini-3.7-flash).
 # GEMINI_MODEL=
 
 # --- Cloud endpoints (all read from the environment; set by default) ---
--- a/config/settings.yaml.template
+++ b/config/settings.yaml.template
@@ -206,7 +206,7 @@
 #     scan_stale_after_sec: 10.0    # seconds without a lidar scan before flagging stale
 #     send_arm_camera_image: true   # also send the arm wrist camera image to the model
 #     log_everything: true          # log every turn's full input
-#     gemini_model: "gemini-3.6-flash"
+#     gemini_model: "gemini-3.7-flash"
 #     idle_turn_interval: 3.0       # seconds between looks when no skill is running
 #     supervision_turn_interval: 5.0  # seconds between looks while a skill runs
 
@@ -237,7 +237,7 @@
 #     stt_realtime_vad_threshold: 0.3       # realtime pair: vendor VAD sensitivity
 #     stt_realtime_vad_silence_secs: 0.7    # realtime pair: vendor silence that ends a turn
 #     elevenlabs_batch_stt_model: "scribe_v2"
-#     gemini_stt_model: "gemini-3.6-flash"
+#     gemini_stt_model: "gemini-3.7-flash"
 #     elevenlabs_stt_model: "scribe_v2_realtime"
 #     openai_realtime_model: "gpt-4o-realtime-preview"
 #     openai_transcribe_model: "gpt-4o-mini-transcribe"
--- a/ros2_ws/src/brain/brain_client/brain_client/brain/tools.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/brain/tools.py
@@ -93,11 +93,12 @@ def build_tools(
     its dispatch map from the same pairs, so the declared names and the
     dispatched names can never diverge.
 
-    While a skill runs, the robot's only action is stopping it, so the
-    declarations collapse. The shape depends on whether the user just spoke:
-    offered any no-op tool, the model calls it and goes silent, so a turn
-    carrying a user message gets stop_current_skill ALONE — plain text becomes
-    the reply channel, and the description steers stop away from questions.
+    A turn carrying user speech never gets a no-op tool: offered one, the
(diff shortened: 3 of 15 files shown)


theo-michel:
Will close cause we saw regressions

PR #672, opened 2026-08-18
perf(webapp): 3–5× faster cold loads, one /map fetch per session

theo-michel:
## What

The default page took ~1.9 s to mount on robot WiFi and moved **1.8 MB** of uncompressed source per cold load — then re-downloaded the ~472 KB latched `/map` grid on **every page switch**. Now it's ready in ~650 ms on WiFi, moves **237 KB**, and fetches the map once per session.

| median, cold load | before | after |
|---|---|---|
| LAN (8 ms, 25 Mbps) ready | 860 ms | **333 ms** |
| Robot WiFi (40 ms, 8 Mbps, 2× CPU) ready | 1865 ms | **649 ms** |
| Weak link (120 ms, 3 Mbps, 4× CPU) ready | 4370 ms | **1443 ms** |
| bytes on the wire | 1817 KB | **237 KB** |
| `/map` tax per page switch | 472 KB | **0** |

Measured with headless Chrome over CDP (network + CPU throttling), against the real `proxy/https_server.py` plus a rosbridge stand-in so the pages actually connect and subscribe.

## Server — `proxy/https_server.py`

- **Gzip text assets on the way out**, cached in memory per file identity with its own ETag/304 path. Closes the static half of `TODO(INN-674)`; the dynamic-JSON half (joints/logs) is still open and noted in the docstring. `app.css` 154 KB → 32 KB; the app total ~4×.
- **`public/vendor/**` served `public, max-age=31536000, immutable`** — three.js r160 (655 KB) is fetched once per browser instead of revalidating every `/armsdk` visit. Everything else stays `no-cache` (the zero-build "deploy" is a file edit).

## Client

- **`index.html` preloads all 27 first-paint modules.** The import graph was discovered one level at a time — five serial round trips before the page module was even requested. `node tests/preload.test.js` (new) recomputes the real graph and fails on drift in either direction.
- **The page import starts in parallel with the sim config gate** instead of after it (one round trip off every load).
- **Route warm-up: strictly after first mount, one route at a time, cockpit neighbours only** (`agent`/`teleop`/`nav`). It used to fire while the first page was still waiting on the network — all twelve route graphs competing for the same six HTTP/1.1 connections — and billed every load ~140 KB for pages most sessions never open. An unwarmed route is one dynamic import away.
- **three.js and the map widget are dynamic imports** — they only load when the Arm SDK view / map tile actually shows.
- **`/map` rides one session-lived subscription** with the last grid cached at module level in `mapWidget.js`. rws replays the full latched grid to every new subscriber, so each page switch re-downloaded an unchanged map (Agent → Nav → Agent = 3 × 472 KB before; one replay now, and the returning page paints from cache instantly). SLAM updates and map switches still flow — the standing subscription *is* the invalidation.
- **One shared agent state** for the shell's "agent is running" pill and the Agent panel. Two instances doubled every `get_available_directives` against the brain's single-threaded executor — the exact query `refresh()` already defends with retry backoff.
- **Boot splash**: removed immediately under `prefers-reduced-motion` (its fade is `transition: none` there, so it lingered opaque over a ready page until the fallback timer); fade cut 0.4 s → 0.2 s now that the mount is faster than the old fade was.

Docs: README gains a "Load speed without a bundler" section stating the four rules that keep the unbundled app fast.

## Verification

- proxy pytest: 20 passed (old `test_no_compression` replaced with tests pinning gzip + its ETag/304 + skip cases + vendor immutable caching); ruff clean
- all 7 webapp node tests pass (incl. the new `preload.test.js`, which was also fault-injected to confirm it catches drift)
- `tsc --noEmit` output byte-identical to baseline
- all 11 routes probed in a real browser (mount + no uncaught errors), including the sim-gated redirect path and the shell pill on non-Agent pages
- map cache probed live: three page mounts → one `/map` replay, teardown of every other subscription still exact

## Left for a robot session (can't validate against a fake rosbridge)

- `compression: "cbor"` on the OccupancyGrid subscribe — if rws honours it, the 472 KB JSON grid becomes tens of KB
- debouncing the double `/webrtc/start` at page load (needs real SDP/media)


webapp/proxy/https_server.py
```diff
@@ -197,15 +201,80 @@ def _safe_resolve(path: Path) -> "Path | None":
...
+    mtime=0 keeps the output byte-identical across calls, so a cache miss after a
+    restart cannot change what a revalidating browser is holding."""
+    global _gzip_cache_bytes
+    key = (str(path), stat.st_mtime_ns, stat.st_size)
+    hit = _GZIP_CACHE.get(key)
+    if hit is not None:
+        return hit
+    body = gzip.compress(path.read_bytes(), compresslevel=6, mtime=0)
+    _GZIP_CACHE[key] = body
+    _gzip_cache_bytes += len(body)
```
theo-michel: **Thread-safety: the gzip cache is mutated by concurrent thread-pool workers with no lock.** `_gzipped` runs under `asyncio.to_thread` (`_serve_static`, line 276), so a cold load — `index.html` preloads ~27 modules, all cache misses at once — runs many `_gzipped` calls concurrently on executor worker threads.

- `_gzip_cache_bytes += len(body)` is a non-atomic read-modify-write, so concurrent misses lose updates and the counter drifts from the real cached size.
- The eviction loop `_GZIP_CACHE.pop(next(iter(_GZIP_CACHE)))` isn't atomic either: once the cache exceeds `GZIP_CACHE_BUDGET` (24 MB — reachable in a sim/dev checkout, where `.obj/.stl/.urdf` up to 4 MB each are in `COMPRESSIBLE`), two workers can read the same first key and the second `.pop()` raises `KeyError` (or `next(iter(...))` hits *dictionary changed size during iteration*) → a 500 on that asset.

Guard the read/insert/evict with a module-level `threading.Lock`.

webapp/js/teleop/cameraSwitch.js
```diff
@@ -138,23 +139,27 @@ export function createCameraSwitch(parent, session, ros, opts = {}) {
...
-          saveMapZoom();
-        },
-        // Memories pulse in live while you drive — the tour builds the
-        // robot's spatial memory, and this is where you watch it happen.
-        layers: { memories: true },
+    if (mapOn && !mapHost) {
+      const host = document.createElement("div");
+      host.className = "cam-map-host";
+      mapHost = host;
+      void import("../map/mapWidget.js").then((m) => {
```
theo-michel: **The map's dynamic `import()` has no `.catch`, so a failed fetch blanks the map for the whole session.** `mapHost` is set before the import; if `import('../map/mapWidget.js')` rejects (a transient blip on the robot WiFi this PR targets), the `.then` body never runs — `mapWidget` stays null and `mapHost` stays non-null. `ensureMap`'s create branch requires `mapOn && !mapHost` (now false) and the teardown branch requires `!mapOn` (dead — `reconcile()` pins `mapOn = true`), so the map is never retried: a blank tile until a full page reload, plus an unhandled rejection.

The router guards its dynamic imports with `.catch(() => {})`; mirror that and reset `mapHost = null` on failure so the next `ensureMap()` can retry.

webapp/proxy/https_server.py
```diff
@@ -197,15 +201,80 @@ def _safe_resolve(path: Path) -> "Path | None":
...
+    immutable vendor libraries.
+
+    A gzipped body is a different representation of the same file, so it carries
+    its own ETag — a client that stops sending Accept-Encoding revalidates against
+    the identity ETag and gets the plain file back, never a mislabelled body."""
+    # public/vendor holds pinned libraries (three.js r160 etc.) — updated by
+    # replacing the file, never edited — so browsers may cache them for good.
+    # Everything else is no-cache: the zero-build app's "deploy" is a file edit.
+    vendored = path.is_relative_to(ROOT / "public" / "vendor")
+    cache = "public, max-age=31536000, immutable" if vendored else "no-cache"
```
theo-michel, line 302: **`immutable` on unversioned vendor URLs means an in-place library bump is invisible to returning clients for a year.** The comment says vendor libs are "updated by replacing the file," but `Cache-Control: immutable` tells the browser never to revalidate. `viz.js` statically imports the unversioned `public/vendor/three.module.min.js`; when someone bumps three.js in place, `viz.js` (no-cache) ships new code but every already-visited browser keeps the stale three.js for up to a year → new viz.js runs against old three.js, unrecoverable short of a hard cache clear.

Consider a content-hash/version in the vendor URL (or a shorter `max-age` without `immutable`) so a replace is actually picked up. Newly introduced — everything was `no-cache` before this PR.

webapp/js/router.js
```diff
@@ -84,7 +84,15 @@ const configPromise = getConfig();
...
+  // Start the page's module graph downloading before the sim gate resolves: the
+  // gate only ever redirects a non-sim section, so on every other load waiting
+  // for /config.json first cost the page a round trip it never needed.
+  let pending = route.load();
+  if ((await configPromise)?.simControls && !SIM_SECTIONS.has(route.key)) {
+    void pending.catch(() => {}); // the speculative load lost the gate; don't strand its rejection
+    route = ROUTES[0];
+    pending = route.load();
+  }
   if (seq !== navSeq) return; // superseded while awaiting config
```
theo-michel: **A render superseded during `await configPromise` orphans the speculative `pending` with no `.catch`.** The redirect branch above deliberately guards its discarded load (`void pending.catch(() => {})`), but reaching this `if (seq !== navSeq) return` drops `pending` (the original load, or the reassigned `ROUTES[0]` load) unhandled. If that module import later rejects (flaky fetch during a rapid nav), it surfaces as an unhandled promise rejection.

Add `void pending.catch(() => {})` before this early return.

webapp/js/map/mapWidget.js
```diff
@@ -29,6 +29,33 @@ import {
...
+let mapFeedStarted = false;
+
+function ensureMapFeed() {
+  if (mapFeedStarted) return;
+  mapFeedStarted = true;
+  ros.subscribe(
+    MAP_TOPIC,
+    (msg) => {
+      lastMapMsg = msg;
+      for (const cb of mapFeed) cb(msg);
```
theo-michel: **The shared feed dispatch lost rosClient's per-handler isolation.** With the old per-widget `ros.subscribe(MAP_TOPIC, onMap, ...)`, rosClient wrapped each handler in its own try/catch (`#handleFrame`), so one widget's `onMap` throwing never affected others. Now all widgets ride a single rosClient handler whose `for (const cb of mapFeed) cb(msg)` is unguarded — a throw from any `onMap` aborts the loop (later subscribers miss that map update) and escapes to rosClient's catch.

Bounded today since normally only one map widget is mounted, but cheap to restore: wrap `cb(msg)` in try/catch.

webapp/proxy/https_server.py
```diff
@@ -197,15 +201,80 @@ def _safe_resolve(path: Path) -> "Path | None":
...
+    headers = {"Content-Type": _content_type(path), "Cache-Control": cache}
+    stat = _gzip_candidate(path, request)
+    if stat is None:
+        return web.FileResponse(path, headers=headers)
```
theo-michel, line 310: **The identity `FileResponse` branch omits the `Vary: Accept-Encoding`** the gzip branch sets. The two representations of one URL then disagree on `Vary`, so an intermediary cache can't safely negotiate them — and a vendor identity response is served `immutable` with no `Vary`. Low impact behind a direct HTTPS/private cache, but set `Vary: Accept-Encoding` on this branch too for consistency (`headers` is already built above).

webapp/js/map/mapWidget.js
```diff
@@ -29,6 +29,33 @@ import {
...
+// The topic is latched, and rws replays the full grid — hundreds of KB of JSON
+// for an apartment — to every new subscriber, so re-subscribing on each page
+// switch re-downloaded an unchanged map. A newly mounted widget paints the
+// cached message synchronously; map changes (SLAM updates, a map switch) keep
+// flowing to the cache, and outside those the topic is silent, so the standing
+// subscription costs nothing.
+/** @type {any} */
+let lastMapMsg = null;
+/** @type {Set<(msg: any) => void>} */
+const mapFeed = new Set();
```
theo-michel: **Cleanup (reuse):** the module-level `mapFeed` Set + `ensureMapFeed` fan-out re-implements rosClient's ref-counted subscribe (it already keeps a per-topic handler Set and loops it, `rosClient.js:281/487`). The genuinely-new need is only a session-lived latch: one permanent `ros.subscribe(MAP_TOPIC, m => { lastMapMsg = m })` keeps the upstream ref-count > 0 — so per-widget subscribe/unsubscribe never tears it down or triggers an rws re-latch — and each widget then uses rosClient's own subscribe/unsub for `onMap` plus the synchronous `lastMapMsg` replay it already does. No second handler registry to maintain.

webapp/js/router.js
```diff
@@ -195,16 +210,23 @@ if (connectTarget) {
...
+void render(routeFor(location.pathname)).then(prefetchRoutes);
+
+// Only the cockpit neighbours: warming all twelve routes pulled ~140 KB (and a
+// long tail of requests) off the robot's uplink on every load, mostly for pages
+// the session never opens. An unwarmed route is one dynamic import away (~50 ms
+// on a LAN) — cold-open cost, not correctness. Strictly after the first page has
+// mounted, and one route at a time: the browser counts the main thread as idle
+// while the first page still waits on the network, so an earlier idle callback
+// put the warm-up on the same six connections as the page the user asked for.
+const WARM_KEYS = new Set(["agent", "teleop", "nav"]);
```
theo-michel: **Cleanup (altitude):** `WARM_KEYS` hardcodes route-key strings disconnected from `ROUTES`. Rename a key and warming silently breaks, with no drift guard — unlike the sibling modulepreload list, which `preload.test.js` checks. A `warm: true` flag on the Route objects co-locates the intent with the routes and can't drift out of sync.

webapp/js/teleop/agentState.js
```diff
@@ -26,10 +27,25 @@ import {
...
+ */
+export function sharedAgentState() {
+  return (_shared ??= createAgentState(ros));
+}
+
 /**
  * @param {import("../rosClient.js").RosClient} rosClient
  */
-export function createAgentState(rosClient) {
+function createAgentState(rosClient) {
```
theo-michel: **Cleanup (simplification):** now that `createAgentState` is private and the module imports `ros` directly (line 10), the `rosClient` parameter is vestigial — the sole caller `sharedAgentState()` passes that same `ros`. Drop the parameter and reference `ros` directly.

theo-michel:
All 12 review threads are addressed in #677 (targets this PR's branch — merging it folds the fixes in here). Thread-by-thread:

**Server — `proxy/https_server.py`**
- *Gzip rejection is ignored* (greptile) → Accept-Encoding is now parsed: `gzip;q=0` rejects, `*` accepts, an explicit gzip token outranks the wildcard. Test added.
- *Gzip cache thread-safety* → the counter and eviction loop now run under a module-level `threading.Lock`; a concurrent duplicate compression is inserted/counted once.
- *Identity branch omits `Vary`* → `Vary: Accept-Encoding` is set for every compressible suffix on both branches.
- *Immutable pins stale vendors* (greptile + follow-up) → `immutable` now requires the filename to name its version; the three vendored files are renamed (`three.module.min.r160.js` etc.) and imports updated, so a bump is a new URL fetched together with the new app code, while an unversioned vendor file falls back to `no-cache`. Test covers both cases.

**Client**
- *Map cache crosses robot connections* (greptile) → the cached grid records which robot it came from and is dropped when a connection targets a different address, so the previous robot's map can never paint.
- *mapFeed re-implements rosClient's subscribe* + *lost per-handler isolation* → the registry is gone: one permanent latch handler pins the `/map` ref-count above zero, and widgets use plain `ros.subscribe`/unsub — per-widget mounts never re-trigger the rws latch replay, and handlers are back inside rosClient's per-handler try/catch.
- *Map dynamic import has no `.catch`* → a rejected import clears `mapHost` so the next `ensureMap()` retries instead of stranding a blank tile.
- *Orphaned `pending` on superseded render* → `void pending.catch(() => {})` before the early return.
- *`WARM_KEYS` disconnected from `ROUTES`* → replaced by a `warm: true` flag on the Route objects.
- *Vestigial `rosClient` parameter* → dropped; `createAgentState()` uses the module's `ros` import.

Validation: proxy pytest 20 passed + ruff clean, all 7 node tests pass (preload drift check included), and `tsc --noEmit` diagnostics are identical to this PR's baseline modulo line shifts and the vendor renames.

---

webapp/js/map/mapWidget.js
```diff
@@ -29,6 +29,46 @@ import {
...
+  );
+  // The cache is one robot's grid: after connecting to a different robot it
+  // must not paint the old map while the new one latches (or never does, when
+  // that robot has no map publisher yet).
+  ros.onStateChange((_state, ip) => {
+    if (ip !== lastMapIp) {
+      lastMapMsg = null;
+      lastMapIp = null;
+    }
+  });
```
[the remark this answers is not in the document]
theo-michel: Fixed in 38401fc: each widget now watches the connection — a switch to a different target drops the rasterized grid plus every frame-anchored leftover (AMCL anchor, trail, memory popups, via the extracted `dropFrameState`) and paints "waiting for /map" until the new robot's grid lands.

---

webapp/js/map/mapWidget.js
```diff
@@ -1722,7 +1749,9 @@ export function createMap(root, opts = {}) {
   render();
 
   syncLayerSubs();
-  const unsubMap = ros.subscribe(MAP_TOPIC, onMap, 250, "nav_msgs/msg/OccupancyGrid");
+  ensureMapFeed();
+  mapFeed.add(onMap);
+  if (lastMapMsg) onMap(lastMapMsg);
```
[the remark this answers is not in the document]
theo-michel: Fixed in e973b67 + 38401fc: the module cache records which robot it came from and is cleared when a connection targets a different address, and a mounted widget now also blanks its rendered grid on the switch instead of keeping the old robot's map painted.

---

webapp/proxy/https_server.py
theo-michel: Fixed in e973b67 via the version-in-URL route: `immutable` is granted only when the vendor filename names its version, the three files are renamed (`three.module.min.r160.js` etc.) with imports updated, and an unversioned vendor file falls back to `no-cache` — so a bump is a new URL and can never strand returning browsers on stale bytes.

---

webapp/proxy/https_server.py
theo-michel: Fixed in e973b67: `Vary: Accept-Encoding` is now set for every compressible suffix in the shared headers, so the identity `FileResponse` and the gzip response agree; test asserts it on the identity branch.

---

webapp/js/map/mapWidget.js
```diff
@@ -1721,8 +1761,41 @@ export function createMap(root, opts = {}) {
...
+    amclPose = null;
+    odomAtAmcl = null;
+    trail.length = 0;
+    // The memory layer swaps itself when the positions payload names the new
+    // map; the search verdict and popup belong to the old frame — drop now.
+    closeMemPopup();
+    hideMemSearchCard();
+    memSearch = null;
+    pose = composedPose();
+  }
```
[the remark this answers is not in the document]
theo-michel: Fixed in 0e09971: the robot-switch handler now clears every message-derived layer — `costGrid`, `localGrid`, `scanMsg`, plan/goal, and the memory state including the known-set (so the new robot's first payload adopts silently instead of pulsing its whole store). A same-robot map switch keeps `mapChanged()`'s existing behavior, since there the grid replays and the memory feed swaps itself.

---

webapp/proxy/https_server.py
```diff
@@ -197,15 +203,114 @@ def _safe_resolve(path: Path) -> "Path | None":
...
+        # URL must agree on Vary or an intermediary can't negotiate them.
+        headers["Vary"] = "Accept-Encoding"
+    stat = _gzip_candidate(path, request)
+    if stat is None:
+        return web.FileResponse(path, headers=headers)
+    etag = f'"{stat.st_mtime_ns:x}-{stat.st_size:x}-gz"'
+    headers |= {"ETag": etag}
+    if _matches_etag(request.headers.get("If-None-Match", ""), etag):
+        return web.Response(status=304, headers=headers)
+    body = await asyncio.to_thread(_gzipped, path, stat)
```
karmanyaahm, line 315: Is there a reason this is reason compression is reimplemented instead of using compress=true from the framework?

karmanyaahm (approved):
After the comment is fixed it lgtm

webapp/proxy/https_server.py
theo-michel: Yes, deliberate — three reasons aiohttp's own compression (`enable_compression()` / a compress middleware) doesn't fit here:

1. **It deflates per response, with no cache.** Every reload would recompress the same 655 KB three.js / 154 KB app.css for every client, on the Jetson the stack shares with SLAM and vision. This path compresses once per file edit and serves cached bytes (keyed by mtime+size, bounded to 24 MB).
2. **It runs zlib on the event loop** — the same loop relaying `/ws` (teleop drive commands), so large-body compression adds jitter exactly where it hurts. Here the one-time compression happens off-loop in `asyncio.to_thread`.
3. **Conditional requests break.** With on-the-fly compression the gzip body has no representation-specific validator, so revalidation either misses or serves the wrong 304. The cache gives the gzip representation its own deterministic ETag (`mtime=0` keeps bytes stable across restarts) with a bodyless 304 + `Vary`, while Range/sendfile stay on the untouched `FileResponse` path.

The framework's cached alternative — precompressed `.gz` siblings that `FileResponse` serves natively — was the original TODO(INN-674) plan, but it needs a build step to generate the siblings and this app is zero-build (deploy = file edit), so they'd silently go stale. Added a sentence saying this at the `COMPRESSIBLE` block in 65d64ea so the next reader doesn't have to ask.

---

PR #675, opened 2026-08-18
feat(agents): add Mars, a friendly intro agent

DavidDobas:
Adds `IntroAgent` ("Mars"), a friendly assistant agent meant for first-run demos and intros.

- Skills: navigation, pick/open/close gripper, memory search, head emotions, and the recorded Wave.
- Mic input enabled, plus gaze so it tracks the person it is talking to.
- Prompt keeps replies short, pairs every utterance with a head emotion, and honors "stop" immediately.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/agents/initializer.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/agents/initializer.py
@@ -109,12 +109,12 @@ def initialize_agents(
     if broken:
         logger.warning(f"{len(broken)} agents failed to load: {list(broken)}")
 
-    # Set default agent (fallback to first available if demo_agent not found)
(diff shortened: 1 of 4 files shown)


ugtthis:
**context:** it seems like the demo agent wants to navigate too much and if user talks to demo agent while its navigating(since in the code it says it is "curious") mars will not respond to user since mars is focused on navigating. That makes this an odd experience for first time users. mars _should_ respond to user regardless if it is navigating or not and ideally if it is doing something it can confirm with user if the user wants mars to continue the previous task it was doing or something like that after it responds. 

doing this intro agent will make it less confusing for first time users talking to mars until we fix why mars cant multi task

ugtthis (approved):

PR #677, opened 2026-08-19
fix(webapp): address PR #672 review — gzip negotiation, vendor pinning, map cache

theo-michel:
## Summary

Addresses all 12 review threads on #672 (targets its head branch, so merging this folds the fixes into that PR), and **merges the latest `main` in**, resolving #672's conflicts — once this lands on the branch, #672 is mergeable again.

- **Gzip negotiation** (`proxy/https_server.py`): Accept-Encoding is parsed instead of substring-matched — `gzip;q=0` is now a rejection, a bare `*` is acceptance, and an explicit gzip token outranks the wildcard.
- **Gzip cache thread-safety**: `_gzipped` runs on concurrent executor threads (`asyncio.to_thread`); the byte counter and the oldest-first eviction loop now run under a `threading.Lock`, closing the lost-update drift and the double-`pop` `KeyError` once the cache exceeds its budget.
- **`Vary: Accept-Encoding` on the identity branch too**, so both representations of a compressible URL agree and an intermediary can negotiate them.
- **Vendor caching can no longer pin stale bytes**: `immutable` is granted only when the vendor filename names its version — the three vendored files are renamed (`three.module.min.r160.js`, `OrbitControls.r160.js`, `STLLoader.r160.js`) and their imports updated, so a library bump is a new URL fetched together with the new app code; an unversioned vendor file falls back to `no-cache`.
- **Map cache** (`mapWidget.js`): the parallel `mapFeed` registry is gone — one permanent latch handler pins rosClient's ref-count for `/map` so per-widget subscribe/unsubscribe never tears down the rws subscription, widgets ride rosClient's own per-handler try/catch again, and the cached grid is dropped when connecting to a *different* robot so the previous robot's map can't paint there.
- **Map import failure** (`cameraSwitch.js`): a rejected `import("../map/mapWidget.js")` clears `mapHost` so the next `ensureMap()` retries, instead of stranding a blank map tile for the session plus an unhandled rejection.
- **Router**: a render superseded while awaiting the sim-config gate now `.catch`es its speculative module load before returning; `WARM_KEYS` is replaced by a `warm` flag on the `Route` objects so renaming a key can't silently detach warming.
- **`agentState.js`**: `createAgentState`'s vestigial `rosClient` parameter dropped; the module's own `ros` import is used directly.
- **Merge of `origin/main`**: one conflict, in `agentPanel.js`'s JSDoc — kept main's new push-to-talk `opts`/returns shape and applied the PR's one-line `createAgentState` → `sharedAgentState` type change. Main's `agentMicControl.js` joined the first-paint graph, so it's added to `index.html`'s modulepreload list (caught by `preload.test.js`, exactly the drift it exists for).

## Validation

  - proxy pytest: 20 passed (vendor test now covers versioned→immutable and unversioned→no-cache; gzip test covers `q=0` rejection, `*` acceptance, and identity-branch `Vary`); `ruff check` + `ruff format --check` clean
  - all 7 webapp node tests pass on the merged tree, including `preload.test.js` (28 first-paint modules, no drift)
  - `tsc --noEmit` diagnostics on the merged tree are identical to `origin/main`'s baseline modulo line-number shifts and the vendor renames (verified by normalized diff)
- [ ] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up`. — N/A, no simulator changes of mine (main's sim changes ride along via the merge)
- [ ] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`. — N/A

https://claude.ai/code/session_01BDEbxUzegvyURwgHAHnvom


theo-michel:
Pushed directly onto `webapp-loading-speed-683836` per the author's request (fast-forward, `32baae9..e973b67`), so #672 now carries these fixes and the main merge itself. Closing this PR as redundant.

---

PR #678, opened 2026-08-19
Compact and group Agent activity logs

axelpey:
## Summary

image

- Keep the Agent activity stream permanently compact: hide completed reasoning and unmatched legacy output while retaining live processing, chat, and every skill-call row.
- Group a skill call's inputs and output or error into one accessible disclosure. The collapsed row shows the complete wrapping skill name, status, and a clear chevron without ellipsizing the name.
- Include successful code-skill output in terminal status events and persist authoritative task rows so grouped calls survive history reloads, while reconciling the legacy `skill_output` echo for compatibility.

## Validation

- [x] Ran all seven webapp Node test files: 45 assertions passed.
- [x] Ran `test_activity_protocol.py`: 4 tests passed.
- [x] Ran JavaScript syntax checks, changed-file TypeScript filtering, Python compile checks, and `git diff --check`.
- [x] Browser-checked the real 360 px panel and 320–390 px phone layouts: no ellipsis or horizontal overflow; click, Enter, and Space disclosure controls work; Inputs plus Output/Error are preserved; compact visibility and bottom pinning hold; no console errors.
- [x] Simulator behavior, config, and assets are unchanged, so simulator startup and asset-pack checks are not applicable.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/transport/activity.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/transport/activity.py
@@ -1,11 +1,43 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (c) 2026 Innate Inc
-"""Small, dependency-free helpers for Agent activity events."""
+"""Small, ROS-free helpers for Agent activity events."""
 
 from __future__ import annotations
 
 import time
 from typing import Any
+
+from typing_extensions import NotRequired, TypedDict
+
+from brain_client.common.enums import SkillResult
+
+
+class TaskStatusPayload(TypedDict):
+    """The /brain/skill_status_update wire format (JSON on a String topic)."""
+
+    primitive_name: str
+    primitive_id: str | None
+    skill_name: str
+    skill_id: str | None
+    status: str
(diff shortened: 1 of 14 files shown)


ros2_ws/src/brain/brain_client/brain_client/transport/activity.py
theo-michel, line 1: Probably remove those tests they are not that usefull long term

theo-michel (approved):
LGTM, maybe just remove the tests that are not that usefull

ugtthis:
Im closing this one in favor of the agent page refresh Im doing. Chat ux/ui below. Still in the works but similair idea. Will post PR once its closer to being done
- simple mode will pin your chat to top and show the next 3-4 msgs from mars including minor details and what mars said
- detailed mode is how the original chat currently is but by default it will be on simple mode
- thing i will add to simple mode is IF any of the skills fail then it should show on simple mode so bugs are more obvious


https://github.com/user-attachments/assets/915eb48f-aeea-4bf8-8169-0ab3820e63d6

PR #679, opened 2026-08-19
feat(webapp): Expression Studio — semantic pose axes + blind VLM judges

theo-michel:
## Summary

- New **`/expression` page (Expression Studio)** — the first-step tooling for giving the MARS agent emotional body language: a representation of "all the values", a live 3D preview, and blind VLM evaluation.
- **Two-layer pose representation** (`expression/model.js`, pure + dependency-free): an actuator vector (6 arm joints, head pitch in servo degrees, base advance/sidestep/turn) and a semantic basis matching the source design's core vocabulary, split into **shape** (the body's configuration) and **stance** (base placement vs the person, the runtime driving layer's future domain):
  - shape: `approach` (retract ↔ lean in — arm+head mass, feet planted), `expand` (contract ↔ expand), `rise` (low ↔ tall; gaze belongs to `attend`, so tall-but-downcast is expressible), `attend` (downcast ↔ raised), `askew` (the asymmetry axis — MARS's head can't roll, so wrist roll + arm cant carry the cocked-head quality)
  - stance: `advance` (withdraw ↔ advance), `orient` (turned away ↔ squared on — neutral stands casually oblique at 15°, so squaring on is an engagement signal), `sidestep`
  - A pose is `neutral + Σ weight·delta + manual offsets`, clamped to actuator limits. Emotions ship as preset weight vectors — e.g. fearful retreats and shrinks *while the head stays on the threat*; curious carries the head-cock cant.
- **Editable basis**: the axes are data, not code. A basis editor panel (grouped under shape/stance eyebrows) renames axes/end labels, edits endpoint deltas as JSON, previews extremes, and adds/removes/restores axes — plus the puppet workflow: sculpt an extreme with the sliders + fine-tune, then **capture** it as that axis's −1/+1 endpoint (the slider takes over holding the pose at ±1). Edits persist per browser and export with the library; saved/imported poses carry their exact actuator vectors so they keep their look across basis changes, and importing a file authored on a different basis offers to adopt it.
- **Full-robot 3D preview** (`expression/viz.js`): same URDF/meshes and parser as the Arm SDK page, but the whole robot rides a movable root (base offset/turn ease like joints), the head joint animates, and a ghosted person-scale observer figure stands where the expression is aimed — the ~1.4 m face-height gap is why `attend` runs to +70°.
- **Blind VLM judges** (`expression/judge.js`): renders three fixed world-frame viewpoints (person's eyes / three-quarter / profile) offscreen and asks a Gemini judge panel — never told the target — to distribute probability over a fixed 10-label vocabulary, plus cues and an impression. Pairwise **A/B vs a pinned pose** names the target but randomizes presentation order per judge. Uses the brain's dev path (`generativelanguage.googleapis.com` + API key, kept in localStorage).
- **Studio page** (`expression/main.js`): axis sliders with preset chips, per-actuator fine-tune offsets, a pose library with thumbnails + JSON export/import (weights **and** exact actuator vectors, ready for the later PCA / runtime-controller steps), score bars + judge log, and **apply-to-robot** over the existing Arm SDK stream topic + `/mars/head/set_position` (base offset stays preview-only; real driving belongs to the runtime controller).
- Rail/router registration (also visible in sim — preview and judging are fully local), README entry, rail-fingerprint test updated.

## Validation

  - `node tests/railLayout.test.js` (updated fingerprints) and the rest of the webapp node tests pass.
  - Headless Playwright against a local static server: page mounts and keeps working with no robot connection; every preset renders a distinct, readable pose in all three judge views (with the realigned basis: the oblique neutral, curious's cant, suspicious/fearful's turned-away-while-watching all verified visually); axis→chip sync, fine-tune offsets, library save/load/thumbnail, export payload shape, and the judge + A/B flows verified against a stubbed Gemini endpoint.
  - Basis editor verified the same way: capture holds the sculpted pose exactly and lands the slider at ±1 (with group eyebrows present); endpoint JSON validation; custom axes insert inside the shape group; persistence across reload; a saved pose loads back visually identical after an axis is redefined; export → import into a fresh profile adopts the file's basis and reproduces the pose; `orient +1` squares the base to exactly 0°.
  - `tsc` adds no errors in `model.js` / `judge.js` / `main.js`; `expression/viz.js` carries only the same untyped-vendored-three.js error classes as the existing `armsdk/viz.js` (the check is documented as optional and the repo baseline is not clean).
  - Python/ROS untouched, so the brain_client ruff/pytest/basedpyright gates don't apply.
- [ ] If I changed simulator behavior, config, or assets, I checked that the simulator starts with `./innate-sim up`. — n/a: no sim behavior, config, or asset changes (webapp-only).
- [ ] If I changed simulator asset files or asset references, I published the asset pack and committed the updated `sim/assets.lock.json`. — n/a.

https://claude.ai/code/session_012Cgr5b4y2dDztAEf6PAWKS

### DIFF
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -10,6 +10,11 @@ same modules:
   global costmap / traveled-path overlays, telemetry panels (pose, velocity,
   lidar, nav state, per-topic receive rates), and live strip charts (commanded
   vs measured velocity, nearest obstacle).
+- **Expression** (`/expression`) — the Expression Studio: author MARS body
+  language on semantic pose axes (approach, turn, rise, expand, reach,
+  attend), preview it on the full-robot 3D model, apply it to the live arm +
+  head, and score poses with blind VLM judges that are never told the target
+  emotion.
 - **Collect** (`collect/`) — record episodes (learned skills) and one-shot
   recorded movements; reuses the teleop cockpit with a recording HUD.
 - **Datasets** (`datasets/`) — browse a skill's episodes and replay them
@@ -68,7 +73,7 @@ js/
   shell.js              icon rail on every page
   railLayout.js         the rail's grouped roster (pure — tests/railLayout.test.js)
   teleop/               teleop modules (joystick, keyboard, head tilt, TTS, arm)
-  nav/ collect/ datasets/ training/ logging/      per-page modules
+  nav/ collect/ datasets/ training/ logging/ expression/      per-page modules
 ```
 
 ## Robot interface (rosbridge `ws://<robot>:9090`, rws)
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -14,7 +14,8 @@ same modules:
   language on semantic pose axes (approach, turn, rise, expand, reach,
   attend), preview it on the full-robot 3D model, apply it to the live arm +
   head, and score poses with blind VLM judges that are never told the target
-  emotion.
+  emotion. The axis basis itself is editable in-studio: sculpt a pose, then
+  capture it as an axis's ±1 extreme; edits persist and export with poses.
 - **Collect** (`collect/`) — record episodes (learned skills) and one-shot
   recorded movements; reuses the teleop cockpit with a recording HUD.
 - **Datasets** (`datasets/`) — browse a skill's episodes and replay them
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -11,11 +11,12 @@ same modules:
   lidar, nav state, per-topic receive rates), and live strip charts (commanded
   vs measured velocity, nearest obstacle).
 - **Expression** (`/expression`) — the Expression Studio: author MARS body
-  language on semantic pose axes (approach, turn, rise, expand, reach,
-  attend), preview it on the full-robot 3D model, apply it to the live arm +
-  head, and score poses with blind VLM judges that are never told the target
-  emotion. The axis basis itself is editable in-studio: sculpt a pose, then
-  capture it as an axis's ±1 extreme; edits persist and export with poses.
+  language on semantic pose axes split into shape (approach, expand, rise,
+  attend, askew) and stance (advance, orient, sidestep), preview it on the
+  full-robot 3D model, apply it to the live arm + head, and score poses with
+  blind VLM judges that are never told the target emotion. The axis basis
+  itself is editable in-studio: sculpt a pose, then capture it as an axis's
+  ±1 extreme; edits persist and export with poses.
 - **Collect** (`collect/`) — record episodes (learned skills) and one-shot
   recorded movements; reuses the teleop cockpit with a recording HUD.
 - **Datasets** (`datasets/`) — browse a skill's episodes and replay them
--- a/webapp/js/expression/judge.js
+++ b/webapp/js/expression/judge.js
@@ -1,0 +1,233 @@
+// @ts-check
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 Innate Inc
+// Blind VLM judges for the Expression Studio. The absolute judge is never
+// told the target emotion — it distributes probability over a fixed label
+// vocabulary from the renders alone, so a pose scores on what an uninvolved
+// observer would read, not on priming. The pairwise judge does name the
+// target ("which looks more X?") because comparisons are what VLMs rate
+// reliably; A/B order is coin-flipped per call to kill position bias.
+//
+// Calls Gemini's generateContent directly (the brain's dev path —
+// generativelanguage.googleapis.com with an API key), straight from the
+// browser: this is bench tooling, the key stays in localStorage.
+
+const BASE_URL = "https://generativelanguage.googleapis.com";
+const DEFAULT_MODEL = "gemini-3.6-flash"; // the brain's default model
+const CONFIG_KEY = "innate.expression.judge";
+const API_KEY_KEY = "innate.geminiApiKey";
+
+// The perceptual vocabulary — a superset of the preset emotions plus the
+// readings poses collapse into when they fail (confused, aggressive,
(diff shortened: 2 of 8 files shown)


theo-michel:
## Code review — Expression Studio (`xhigh`)

_Automated multi-agent review of this branch. Note: the requested target `720` doesn't resolve to any PR in this repo (highest is #702); this is the current branch, so I reviewed **PR #679**. Inline comments weren't available in this run, so findings are consolidated below with `file:line` anchors._

Scope: the 8 changed files (4 new `webapp/js/expression/*.js` + rail/router/README/test). Cross-file API surface (imports, ROS topics/actions, `HEAD_*` constants, the URDF `joint_head`/mesh prefix, `destroy()` disposal, the A/B un-flip logic, normalize, and export/import round-trip) was checked and is **correct**.

### Correctness / security

1. **`main.js:988` (+`:1029`) — `crypto.randomUUID()` is unguarded; Save & Import break on the plain-HTTP robot origin.** It's secure-context-only. On `http://<robot-ip>` (the origin `clipboard.js` says the webapp normally uses) it's `undefined`, so Save throws silently and Import reports "import failed" for every file. The rest of the app already guards it — `webrtcSession.js`/`agentPanel.js` use `crypto.randomUUID?.() ?? <fallback>`. Do the same here.

2. **`main.js:850` — XSS: VLM `impression`/`cues` are put into `innerHTML` unescaped** (every other free-text field uses `escapeHtml`). A model reply like `impression:""` executes in a page holding a live `ros` client — it can drive the arm/head and read the Gemini key from `localStorage`. Use `textContent`/`escapeHtml`.

3. **`main.js:680` — `lastHeadSent` is advanced even when the publish is a no-op (socket down), so the head is never re-sent after reconnect.** Drive head=30 while disconnected → publish is dropped but `lastHeadSent=30` → after reconnect, re-sending the same pose is deduped away and the head never reaches 30. The arm self-heals (stream re-publishes every cycle); only the head sticks. Reset the latch on disconnect, or advance it only after a real publish.

4. **`main.js:995` — a saved pose can carry another pose's blind score.** `lastVerdict` is set on Evaluate but never invalidated on edit, so Evaluate A → tweak to B → Save stores A's score on B (and the log prints A's number). This silently mislabels the exported dataset. Clear `lastVerdict` on any weight/offset change.

5. **`main.js:879` — A/B comparison uses `Promise.all` with no per-call catch,** so one flaky `judgePair` (timeout/429/empty response) discards all successful verdicts and reports "compare error". `judgePanel` deliberately tolerates per-judge failures; the pairwise path should too.

6. **`viz.js:290 — `snapshot()` has no `try/finally`.** It sets `scene.background`/hides the home ring, then lazily creates a *second* WebGLRenderer and calls `render`/`toDataURL`. If any of that throws (e.g. WebGL context cap), the restore never runs and the live preview is left with an opaque gray background + hidden ring for the rest of the session.

7. **`viz.js:27` — the person's-eyes judge view fov (26°) is too narrow to cover the ±0.3 m base roam the comment promises.** At ~0.95 m it frames ~±0.22 m, so extreme `sidestep` poses get cropped for the blind judge — biasing exactly the stance axis this view exists to show.

8. **`judge.js:52` — `judges` is clamped but not floored.** A numeric input still accepts `3.5`; `Array.from({length:3.5})` floors to 3 runs while the status shows "/3.5". Add `Math.round` inside the clamp.

9. **`main.js:899` — `loadLibrary()` validates only `Array.isArray`, not per-entry shape.** A corrupt entry (non-numeric `judge.p`) makes `renderLibrary`'s `.toFixed(2)` throw and blanks the *entire* pose list. Sanitize entries or guard the `toFixed`.

10. **`model.js:286` — `synthesize()`'s offset loop iterates the object's own keys, not `ACTUATORS`.** A stray key (reachable via the unsanitized `loadLibrary` above) becomes `NaN` in the returned pose, and `clampPose` only scrubs known actuators — so a bogus field leaks into the exported `actuators`. Restrict the loop to `ACTUATORS`.

11. **`viz.js:292 — `stepTargets(1)` "settle" permanently fast-forwards the shared animation without restoring,** contradicting its docstring ("restored after"). The on-screen preview snaps to the settled pose on every Evaluate/Refresh.

### Cleanup

12. **`viz.js` (reuse) — ~140 lines forked near-verbatim from `armsdk/viz.js`** (`rpyQuat`, `vec3`, URDF parse guard, renderer/scene/lights/grid setup, the link/joint tree build, the STL load+dispose loop, the frame/resize/ResizeObserver loop). Extract a shared `createUrdfViz`; expression's movable root / head joint / observer figure / snapshots become options.

13. **`main.js:51` (reuse) — ~80 lines of CSS (`.armsdk`→`.exs`), the arm constants (`ARM_ACTION`/`ARM_ACTION_TYPE`/`STREAM_TOPIC`/`ARM_STATE_TOPIC`), `deg`, and the `pumpRobot` deadman loop all duplicate `armsdk/main.js`.** Move the constants into `constants.js` (which already owns `ARM_STATUS_TOPIC`/`HEAD_*`) and share the pump as `streamJointsUntilSettled`.

14. **`main.js:394` (efficiency) — `groupEyebrow` rebuilds `new Set(basis.map(...))` per axis per render → O(n²).** Compute `multiGroup` once per render and pass it in.

15. **`main.js:358` (efficiency) — `markActivePreset` runs on every `applyPose`** (every slider frame / live-drive step), re-querying the DOM and doing O(presets×axes) comparisons just to toggle a class. Cache the buttons and recompute only when weights/offsets change.

_Also worth a quick pass (not listed above): Save synthesizes the pose 3× (`main.js` save handler) and `renderLibrary()` runs twice on load._

theo-michel:
Closing as stale — Expression Studio never went past draft. The branch `claude/mars-expression-studio-5fiqi6` keeps the work if it is wanted later.

PR #681, opened 2026-08-19
fix(sim): harden localization, navigation maps, and lighting

axelpey:
## Summary
- recover simulator localization after pose resets and sustained AMCL drift
- keep generated navigation maps within authored apartment floor while preserving lidar-consistent obstacle geometry
- make MuJoCo apartment lighting directional so robot-camera visibility matches the viewer
- reject manual web-map goals outside known-free occupancy cells

This is the platform-hardening base for the Household Orders challenge and agent PRs. It contains no resident assets, challenge runtime, agent, or skill code.

## Validation
- `uv run --project sim python sim/sandbox/test_navigation_grid.py`
- `node webapp/tests/mapGoal.test.js`
- downstream challenge suite: 30 passed (`test_household_npc.py`, `test_assets_image_inputs.py`, `test_sim_world_lighting.py`)

The ROS-dependent localization unit test remains covered by CI because the host environment does not provide `rclpy`.


sim/sandbox/test_navigation_grid.py
```diff
@@ -0,0 +1,79 @@
...
+import sys
+from pathlib import Path
+
+import numpy as np
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "tools"))
+
+import _driver_pkg  # noqa: F401
+from export_nav_map import DEFAULT_NAVIGATION_ENVIRONMENT, _load_navigation_environment
+from mars_sim_driver.core import _navigation_grid
```
DavidDobas: Are these tests needed?

DavidDobas (commented):
Reviewed the two commits against `main`. The four fixes are each defensible — the lighting bug, the `free_thresh` 205-reloads-as-free bug, and the exterior-floor clipping are real and worth having.

The concern is cost-to-value. Roughly **400 of the 850 added lines** are test scaffolding or JSON plumbing that can come out without losing a fix: a 2-line XML lighting fix ships with a 29-line test that never runs, and a boolean mask ships with a new top-level directory plus four build-config edits.

Inline comments below, roughly in priority order. Two are correctness risks (the reset heuristic's 200 ms window, and the unbounded pending-pose handshake); the rest are placement, DRY, and dead-code cleanups.

One verification note: the assets image currently in use predates this branch, so the new export path isn't exercised by a running sim without an assets rebuild.

ros2_ws/src/mars_bot/mars_sim_driver/test/test_grid_localizer_sim.py
```diff
@@ -0,0 +1,209 @@
+import math
```
DavidDobas: **This file never executes.** `.github/workflows/launcher-tests.yml:3` states it directly: *"nothing else in CI runs pytest"* — no job runs `colcon test` or pytest over `ros2_ws/`. The PR description says this test *"remains covered by CI because the host environment does not provide `rclpy`"*, but there is no job that would run it.

Beyond not running, the approach is fragile: lines 29-43 test by calling unbound methods against a `SimpleNamespace` that hand-mirrors all six private fields of the node. That mock has to be updated in lockstep with every state change, and it still can't catch a wiring error. It's also the only file in this package's `test/`, so it introduces a pytest surface the package didn't previously have.

CLAUDE.md is explicit here — *"delete the ones you wrote before pushing the PR. A new test file ships only when the user asked for tests."*

Suggest deleting (−209 lines). If you want the drift state machine pinned, that's a reason to add a job that actually runs it, not to keep an unrun file.

tests/test_sim_world_lighting.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-License-Identifier: Apache-2.0
```
DavidDobas: **This always skips.** It lives in the one directory CI does run (`launcher-tests.yml` → `pytest tests/ -q`), but that job installs only `pytest pyyaml`, so `pytest.importorskip("mujoco")` skips the module every time.

Even if it ran, it asserts `light_type == [DIRECTIONAL, DIRECTIONAL]` and `castshadow == [1, 0]` — a direct restatement of the two-line XML change in `world.py`, so it can't fail for any reason other than someone reverting that line.

Suggest deleting (−29 lines).

sim/sandbox/test_navigation_grid.py
```diff
@@ -0,0 +1,79 @@
+"""Fast invariant checks for navigation-grid fusion and environment metadata.
```
DavidDobas: Manual-run script, not wired to CI. `test_navigation_grid_fusion` exercises trivial boolean masking, and `test_apartment_navigation_metadata` only asserts that the JSON's fields are non-empty — which `_validate_navigation_grid` already enforces far more strictly at export time.

Worth noting the tail wagging the dog: the `ValueError` shape guard in `core.py:107-111` exists only so lines 47-52 here can assert it. Per CLAUDE.md — *"Never wrap code that cannot raise"* — deleting this test lets that guard go too, since the only caller passes arrays derived from a single `base_grid.shape`.

Suggest deleting (−79 lines, plus the guard).

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
```diff
@@ -4,57 +4,203 @@
...
 
     def _on_odom(self, msg):
+        previous = self._last_odom
         self._last_odom = msg
+        # ROS stamps remain wall time across a MuJoCo reset. Ground-truth odom
+        # itself teleports to the challenge spawn, which is the reliable reset
+        # signal available to this ROS-side sim adapter.
+        if previous is not None:
+            sample_gap_s = (_stamp_ns(msg) - _stamp_ns(previous)) / 1_000_000_000
+            if not 0.0 <= sample_gap_s <= ODOM_RESET_MAX_SAMPLE_GAP_S:
```
DavidDobas: **Correctness risk, and this reinvents a signal that already exists.**

Two explicit reset signals are already in the tree: `/virtual_mars/reset` (`node.py:170`) and `challenges.world_epoch`, which is already surfaced through `world_server.py:146`. Inferring a reset from an odom position jump is strictly weaker than subscribing to either.

The failure mode is the exact case this exists for. `/odom` is nominally 30 Hz (`node.py:70`), but the guard here silently skips detection whenever a sample gap exceeds 200 ms — and a MuJoCo reset plus prop respawn is precisely the event most likely to stall the publisher past that. `node.py:399` also suspends `/odom` outright when the world server is unreachable, and a loaded or WSL host drops below 5 Hz routinely. In all of those cases the reset goes undetected and AMCL is never reseeded.

Switching to the explicit signal also deletes `ODOM_RESET_POSITION_JUMP_M`, `ODOM_RESET_YAW_JUMP_RAD`, `ODOM_RESET_MAX_SAMPLE_GAP_S`, `_stamp_ns`, and this whole comparison block.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
```diff
@@ -4,57 +4,203 @@
...
+            self._begin_localization()
 
-    def _on_amcl_pose(self, _msg):
-        if self._retry_timer is not None:
-            self._retry_timer.cancel()
-            self._retry_timer = None
-            self.get_logger().info("AMCL localized; stopping /initialpose retries")
+    def _on_amcl_pose(self, msg):
+        pose = msg.pose.pose
+        if self._pending_pose is not None:
```
DavidDobas: **The pending handshake has no bound, and it gates the drift guard.**

This branch returns before any drift logic, so while `_pending_pose` is set, drift detection is fully disabled. The only thing that clears it is a positional match — there is no timeout and no retry ceiling.

The match is not guaranteed. `_publish_pose` re-anchors `_pending_pose` to *current* odom on every 2 s retry, while AMCL's `/amcl_pose` reflects its estimate one update latency ago. Driving faster than ~0.35 m per AMCL latency means the two never land inside `POSITION_MATCH_TOLERANCE_M`, so the retry timer never cancels and the drift guard stays off indefinitely. In practice it usually converges, but nothing here enforces that — and the previous code cancelled on any `/amcl_pose`, so this is a new way to get stuck.

Suggest either time-bounding the pending state (fall back to the old cancel-on-any-pose after N seconds), or comparing AMCL's pose against live odom rather than the snapshot taken at publish time.

sim/tools/export_nav_map.py
```diff
@@ -6,35 +6,261 @@
...
+    exterior_boundary_half_width_m: float
+
+
+def _point(value, label: str) -> tuple[float, float]:
+    if not isinstance(value, list) or len(value) != 2 or not all(isinstance(item, int | float) for item in value):
+        raise ValueError(f"{label} must be an [x, y] number pair")
+    return float(value[0]), float(value[1])
+
+
+def _load_navigation_environment(path: Path = DEFAULT_NAVIGATION_ENVIRONMENT) -> NavigationEnvironment:
```
DavidDobas, line 49: **`sim/environments/` costs four build-config touchpoints for one 26-line file, and nothing selects an environment.**

The commit that moved this data out of the tool bought: a 26-line `navigation.json`, ~60 lines of hand-rolled `isinstance` validation (lines 44-100), a Dockerfile `COPY`, a `.dockerignore` entry, a CI path glob, and a launcher pathspec — and it lost type safety in the process.

The multi-environment payoff isn't there yet: this function has a hardcoded default path and no caller passes an argument. There's also no selector anywhere in the tree.

It's a third naming scheme, too — source assets are `sim/assets/apartment_*`, installed bundles are `assets/apartments/<name>/`, and this adds `sim/environments/apartment/`.

The same data as a frozen dataclass literal in this module is typed for free, needs no parser, and touches zero build config. If multi-environment support is genuinely imminent, the better move is to wire it to a real selector — otherwise this is plumbing for a caller that doesn't exist.

sim/tools/export_nav_map.py
```diff
@@ -6,35 +6,261 @@
...
+        return None
+    return row, col
+
+
+def _cell_value(grid: np.ndarray, ox: float, oy: float, x: float, y: float) -> int | None:
+    index = _cell_index(grid, ox, oy, x, y)
+    return None if index is None else int(grid[index])
+
+
+def _free_component(grid: np.ndarray, start: tuple[int, int]) -> set[tuple[int, int]]:
```
DavidDobas, line 78: This file went 40 → 269 lines and now carries three concerns: the export itself, geometric editing of the grid, and validation.

`_free_component`, `_distance_to_segment`, `_apply_indoor_boundary`, `_validate_navigation_grid`, and `_validate_pgm_roundtrip` are all pure — no I/O, no MuJoCo. CLAUDE.md puts exactly that category in a `utils.py` beside the caller (*"One concern per module"*, *"Pure functions — no ROS, no I/O, no self-state — live in a `utils.py` beside their caller"*).

Splitting them out leaves `main()` readable end to end: load environment → build grid → clip → validate → encode → validate encoding → write.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/core.py
```diff
@@ -519,13 +546,22 @@ def occupancy_grid(self, resolution: float = 0.05) -> tuple[np.ndarray, float, f
 
         grid = np.full(n, -1, dtype=np.int8)
         origins = np.column_stack([gx.ravel(), gy.ravel(), np.full(n, wall_top)])
+        # The infinite ground plane exists so the robot has something to
```
DavidDobas: **Two of these three filters are dead on the only reachable path, and the comment argues for the wrong one.**

The ground plane is filtered three ways here: the `_lidar_groups` mask, `bodyexclude=0`, and `geom_bodyid[hit] == apt`. But the function now early-returns when `_lidar_groups is None`, so `visual_rooms` is always truthy by this point — which means `collision_group = COLLISION_GROUP` (`world.py:223`) and the ground plane is authored at `group=3` (`world.py:288`), while `_lidar_groups` selects only `VISUAL_GROUP=1`. The mask alone already excludes it; `bodyexclude=0` and the `apt` check add nothing for the ground plane.

The `apt` check isn't purely dead — props are also emitted into `VISUAL_GROUP` (`world.py:221-222`), so it does reject them. But props are parked off-map at export time, so the comment's *"future static scenery"* framing describes a hypothetical while the real (current) justification goes unmentioned.

Also `floor_groups = self._lidar_groups` on the next line is a pure alias — `self._lidar_groups` reads fine inline.

Suggest keeping the single filter you actually mean, dropping the alias, and collapsing this to one line stating why the ground plane isn't apartment floor. As written it's the narrating comment CLAUDE.md rules out.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
```diff
@@ -4,57 +4,203 @@
...
 from nav_msgs.msg import Odometry
 from rclpy.lifecycle import LifecycleNode, TransitionCallbackReturn
 from rclpy.qos import DurabilityPolicy, QoSProfile
 from std_srvs.srv import Trigger
 
+POSITION_MATCH_TOLERANCE_M = 0.35
+YAW_MATCH_TOLERANCE_RAD = math.radians(20.0)
+ODOM_RESET_POSITION_JUMP_M = 0.5
+ODOM_RESET_YAW_JUMP_RAD = math.radians(30.0)
+ODOM_RESET_MAX_SAMPLE_GAP_S = 0.2
```
DavidDobas: **12 module constants and 6 instance fields for one state machine.**

The severe and moderate tiers differ only in `(position_threshold, yaw_threshold, min_samples, dwell)` — that's one table of two rows, not eight flat constants, and `_on_amcl_pose` currently picks between them with a pair of ternaries.

On the state side, `_drift_bad_since`, `_drift_bad_samples`, `_drift_armed`, `_drift_healthy_since`, and `_last_auto_reseed_at` are one lifecycle spread across five fields, and the same four-line reset block appears verbatim in three places (`on_activate`, `on_deactivate`, and the reset branch of `_on_odom`). CLAUDE.md prefers a dataclass for process-internal state like this — folding them into one with a single `reset()` removes the triplication and makes the armed/rearm/cooldown sequence readable in one place.

Worth asking whether the moderate tier earns its keep at all: in sim, odom *is* ground truth, so the two-tier design is guarding against fighting AMCL rather than against uncertainty.

DavidDobas (approved):
Approving — the four fixes are sound and I don't want the cleanup blocking them.

Everything in my [detailed review](https://github.com/innate-inc/innate-os/pull/681#pullrequestreview-4988303267) is addressable in follow-up **except** two items I'd want resolved before merge, since both are live correctness risks rather than style:

- **The 200 ms odom-gap window** silently disables reset detection exactly when a reset is most likely to stall the publisher past it. Subscribing to `/virtual_mars/reset` or `challenges.world_epoch` is both simpler and strictly more reliable.
- **The unbounded `_pending_pose` handshake** gates the drift guard with no timeout, so a persistent mismatch leaves it off indefinitely.

The rest — the three unrun test files (−317 lines), `sim/environments/` plumbing, the dead ground-plane filters, and the drift-constant table — is cost-reduction that can land separately.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
```diff
@@ -64,6 +210,8 @@ def _publish_pose(self) -> bool:
         msg.pose.pose = self._last_odom.pose.pose
         msg.pose.covariance[0] = msg.pose.covariance[7] = 0.01
         msg.pose.covariance[35] = 0.01
+        pose = msg.pose.pose
+        self._pending_pose = (pose.position.x, pose.position.y, _yaw(pose))
```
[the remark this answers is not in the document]
axelpey: Fixed in 8a6362b7. The moving `_pending_pose` target is gone; the first fresh `/amcl_pose` now stops retries, restoring the original bounded handshake behavior.

ros2_ws/src/mars_bot/mars_sim_driver/test/test_grid_localizer_sim.py
axelpey: Deleted in 8a6362b7.

tests/test_sim_world_lighting.py
axelpey: Deleted in 8a6362b7.

sim/sandbox/test_navigation_grid.py
axelpey: Deleted in 8a6362b7, together with the test-only shape guard.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
axelpey: Fixed in 8a6362b7 with an explicit reset generation rather than the odom heuristic. Every `VirtualMars.reset()` advances `world_epoch`; the world server carries it to the ROS driver, which publishes a latched `/virtual_mars/world_epoch`. The localizer waits for the first odom sample from the new generation before reseeding, covering challenge starts, manual resets, and physics recovery.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
axelpey: Fixed in 8a6362b7. `_pending_pose` and its early-return gate were removed; any fresh AMCL response cancels the retry timer.

sim/tools/export_nav_map.py
axelpey: Keeping the coordinates environment-owned intentionally: this stack is expected to support additional environments, and apartment geometry should not return to the generic exporter. I did remove the hand-rolled schema validation, reducing the loader from 53 lines to 24; malformed checked-in data still fails naturally during the asset build.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/core.py
axelpey: Simplified in 8a6362b7: removed the redundant world-body exclusion, the alias, and the narrating comment. The visual-group mask excludes the ground plane; the apartment-body check excludes props sharing that group.

sim/tools/export_nav_map.py
axelpey: Agreed as a follow-up. I am leaving the pure-function module split out of this correction so the review fix does not introduce another file move; the exporter is now smaller after removing the schema plumbing.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
axelpey: Partially simplified in 8a6362b7: the three inferred-reset constants are gone and lifecycle reset state now goes through one helper. I am leaving the two drift tiers and state representation unchanged here; consolidating those is a separate behavior/refactor decision rather than part of the two correctness fixes.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
```diff
@@ -4,57 +4,190 @@
...
+        pose = msg.pose.pose
         if self._retry_timer is not None:
-            self._retry_timer.cancel()
-            self._retry_timer = None
+            self._clear_drift_observation()
+            if not self._drift_armed:
+                self._drift_healthy_since = time.monotonic()
+            self._cancel_retry()
             self.get_logger().info("AMCL localized; stopping /initialpose retries")
+            return
```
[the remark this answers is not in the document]
axelpey: Fixed in c152bcc7. Startup/reset localization still accepts the first fresh response, but automatic drift recovery now compares each AMCL pose against live odometry and keeps retrying until both position and yaw are within the healthy tolerances. This avoids both the stale-pose cancellation and the earlier moving-snapshot target.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/grid_localizer_sim.py
```diff
@@ -4,57 +4,196 @@
...
+        pose = msg.pose.pose
         if self._retry_timer is not None:
-            self._retry_timer.cancel()
-            self._retry_timer = None
+            if not self._drift_armed:
+                if self._last_odom is None:
+                    return
+                position_error, yaw_error = _pose_error(pose, self._last_odom.pose.pose)
+                if position_error > POSITION_MATCH_TOLERANCE_M or yaw_error > YAW_MATCH_TOLERANCE_RAD:
+                    return
```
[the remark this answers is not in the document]
axelpey: Fixed in 550234ab. Automatic recovery now requires both: (1) the AMCL message header timestamp is at or after the current localization attempt began, and (2) its pose matches live odometry. Retries keep a stable freshness boundary even while republishing current ground truth, and a successful recovery immediately rearms drift monitoring.

PR #683, opened 2026-08-21
perf(sim): stop re-downloading dependencies, and give each checkout its own stack

DavidDobas:
Seven things people kept hitting while working on the sim. They have separate causes but one shared shape: every one only fires when somebody has *edited* something. CI builds from a clean checkout and end users run published images, so neither ever executes these paths — which is why they sat undetected, and why the first person doing sim development meets all seven at once.

Numbers below are measured on this branch against the live registry, not estimated.

---

## "Docker downloads gigabytes every time I pull main"

**What happens.** You `git pull`, run `./innate-sim up`, and wait through a multi-gigabyte pull for a change that touched a few source files.

**Why.** Docker skips layers you already have. Ours were never the same twice. `sim/Dockerfile` said `FROM ros:humble-ros-base` — a *tag*, and upstream rebuilds that image roughly weekly. A moved base makes every layer above it different, so nothing could be reused. The two published `innate-os-sim-ros` images from Aug 15 and Aug 18:

| | |
|---|---|
| non-empty layers each | 26 |
| shared between them | **1** |
| layers shared from the bottom | **0** — not even the ROS base |
| wire cost of the newer given the older | **1437 MB** |

Only ~65 MB of that was the actual commit.

**Fix.** Pin both bases by digest: the upstream ROS base to its multi-arch index digest, and the ROS image's own base to the digest of the deps image just pushed rather than a mutable `buildcache` tag. Both links have to be immutable or neither is. CI also reuses a published deps image when its inputs are unchanged, so identical content stops being rebuilt into different bytes.

**→ 1437 MB → 64.6 MB on the wire.** Verified after the fact on the branch's own published images: the ROS prebuilt's 28 layers contain **all 23** of the deps image's, so a developer holding one pulls the other for **0 bytes**. On disk, holding both costs 6.72 GB rather than 12.9 GB.

---

## "I edited one file and now it wants half an hour"

**What happens.** You change something in `ros2_ws/src`, run `up`, and get *"No matching prebuilt image… Building it locally instead"* — 25–30 minutes of apt and torch.

**Why.** The prebuilt image bakes a colcon workspace, so its tag hashes all of `ros2_ws/src`. Edit one file and no published tag matches — correct, and by design. But underneath it sits a dependency-only image with nothing to do with your source, which CI has built and published as `innate-os-sim-deps` since the very first publish workflow. **Nothing has ever pulled it.** The launcher knew only `innate-os-sim-ros`, so it rebuilt from scratch what was already in the registry.

That is the ordinary state of a tree being worked on, and the **only** state a fork with no CI can ever be in.

**Fix.** Publish the deps image at its own content address (`deps-<hash>` over the files `sim/Dockerfile` actually reads) and have the launcher pull it, adopted under the local build's name — the same address, so the offline resolution, the prune and the next run's fast path keep working untouched.

---

## "I have two clones and switching between them rebuilds everything"

**What happens.** A branch clone beside your main one. `up` in the second either silently serves the *first* one's code, or rebuilds the whole ROS workspace — and going back rebuilds it again. Every switch, both directions.

**Why.** One container name and one set of named volumes for the whole machine. If the first clone's container was running, the launcher reused it, bind mounts and all, pointing at the wrong checkout. If it wasn't, compose recreated it and `validate_sim_ros_install.zsh` found a foreign repo id in the volume and wiped `build/install/log`.

**Fix.** Project and container names carry a hash of the checkout path, so each clone keeps its own container, network and compiled workspace. A checkout change no longer triggers a wipe — the paths inside the container are identical either way, and `colcon_build_with_retry` already cleans on the stale-symlink failure when one actually happens.

Ports are still shared (9090 and 9999 aren't overridable), so only one stack runs at a time — but a second `up` now stops with the **name and path of the checkout holding them**, read off that container's own bind mount, instead of failing inside compose.

**Verified on a real machine:** two fully-built workspaces coexisting, 290.4 MB and 21 packages each, each volume stamped with its own checkout id. That is structurally impossible on `main`.

---

## "I changed one file and it rebuilt the whole workspace"

**What happens.** A clean-tree `up`, then one edited Python file, and colcon rebuilds everything — 23 packages, C++ from scratch, north of 20 minutes. The edit *after* that takes 20 seconds.

**Why.** The clean-tree run seeds `install/` from the prebuilt image, and that seed is what breaks the next build: it drops the image's `install/` — built at `/opt/innate-os-prebuilt/ros2_ws`, and **without** `--symlink-install` where the in-container build uses it — on top of a `build/` tree from a local build. colcon cannot build incrementally against that. Reverting the edit re-arms it, because the local install marker goes stale and the next `up` seeds again.

**Fix.** Seed only into a workspace with nothing to build on. Keyed on `build/` rather than on `install/` matching, because an install that matches right now says nothing about whether a rebuild can be incremental. First runs are unaffected — an empty `build/` still seeds, which is the case the prebuilt exists for.

**→ measured: the clean-tree round trip that used to reseed now finishes in 16s with a 2.82s colcon pass.**

---

## "Docker has eaten my disk"

**What happens.** `docker system df` climbing without bound: 19.31 GB of images, 13.07 GB reclaimable, including two ~6.7 GB ROS images sharing 32 bytes and three ~332 MB asset images nothing has referenced since asset distribution moved to Releases.

**Why.** ghcr retention was designed twice — the original weekly cleanup in #391, then #525's age/branch-aware policy sized for every-branch publishing. The client side never got a counterpart. The prune derived its repo from `resolve_local_os_image`, so it only ever swept images this machine **built**. Every pulled tag was kept forever.

**Fix.** The sweep covers the pulled families too, keeps one predecessor so bouncing between branches doesn't re-download, runs on every `up` rather than only when the container is recreated, and never runs under `--offline`.

The ordering matters: this is only cheap *because* the bases are now pinned. On unshared images, dropping the old one frees 5 GB and costs 5 GB to get back. Pinned, it is ~65 MB either way.

---

## "I edited the driver and the sim won't start at all"

**What happens.** `up` dies before starting anything:

```
assets  ✗ world geometry
Error: No published sim asset image for this checkout (…): HTTP 404
```

…with the advice *"Push the branch so CI publishes it."*

**Why.** `mars_sim_driver` and `mars_sim` are hashed into the asset image for one narrow reason: `export_nav_map.py` instantiates `VirtualMars`, so the nav map is a function of the laser's height. That is enough to rename the whole image. Editing `node.py` — a pure RPC client that touches no geometry — renames a multi-hour image, and geometry is the one layer with no local build path to fall back to. For a fork this is terminal, not slow: no CI, so "push the branch" is not advice they can take.

**Fix.** A second, narrower hash answers the question `up` actually needs — *is the geometry already on disk still the geometry this checkout describes?* Unchanged means it is, so it is reused with a log line instead of a refusal. The image's identity is untouched: a driver edit really can move the nav map, and the tag must still say so.

The narrow list is carved out of `ASSETS_IMAGE_PATHSPECS` by **exclusion**, so a pathspec added to the asset image later joins the geometry hash by default. That is the safe direction to be wrong in: one wrongly excluded costs a needless refusal, one wrongly included would reuse geometry that no longer matches.

---

## "It has said the same thing for four minutes — is it working or hung?"

**What happens.** The `os` step shows `Starting the Innate OS container` and an elapsed clock, and nothing else, for as long as it takes. `scripts/install-sim.sh` has drawn a live log tail under its status line all along.

**Why.** `LiveStep` never grew the same region — `_animate` wrote one line. `run_logged_with_heartbeat` computed the latest log line and then used it only *outside* a step. And the label was wrong anyway: that one step covers starting a container, a colcon build that can run for minutes, and launching the nodes.

**Fix.** Two things. `LiveStep` draws up to three cleaned log rows under its status line, written in one go and stepped back over — the technique `draw_step_frame` uses, for the same reason (clearing and drawing separately leaves the region blank long enough to read as flicker). And the step retitles itself as the work moves between phases. Steps with a progress formatter keep their single line: a formatter means there is something better to show, and for the pulls that is the aggregate bar which replaced exactly this layer chatter on purpose.

---

# How this fits what came before

Worth reading the trail, because none of this reverses a decision that was made — it finishes several that were half-made.

| PR | decided | what it did not cover |
|---|---|---|
| **#390** (May 11) | `Dockerfile.ros-prebuilt` + `seed_prebuilt_install`, **opt-in**. Only stated rationale is a config comment: *"skip the first local colcon build when source matches."* | The seed was designed for a first run. Nothing considered a returning developer with a warm build tree. |
| **#391** (May 13) | Flipped the default to `auto` and content-addressed the tags: *"avoids asking developers to copy a commit SHA while also avoiding a moving `:main` image mismatch."* | Hit the mutable-base problem **during that PR** — *"the ROS image still rebuilding because its base image tag changed per run"* — and fixed it for CI's build time with a stable `buildcache` tag, never for consumers' pulls. Its cost analysis is entirely registry-side (Cloud Build minutes, GHCR storage "currently free"). |
| **#524** (Jul 13) | Multi-arch, on a clear trade: *"~25–30 min local Docker build (plus ~10 GB of build cache) that amd64 users skip with a ~5 min pull."* Also moved publishing from ~50/month to **400–600/month**. | Compared pull *time* against build *time*, never per-pull bytes — and 10×'d the frequency in the same change. |
| **#525** (Jul 14) | Age/branch-aware GHCR retention sized for that new volume. | Sized GitHub's disk. Nobody sized the user's. |
| **#526** (Jul 14) | Hash alignment: *"silently broken for everyone, on every platform… The 'normal first-run build' everyone has been paying is this bug."* | Means the pull path only went live in mid-July — simultaneously with multi-arch and 10× publishing. #391's assumptions were never re-examined under those conditions. |
| **#661** (Aug 18) | One-command install made it load-bearing: `sim-stable` only advances to commits whose images are published. | |
| **4c71ddca** (May 27) | The repo-id marker, *"Avoid repeated ROS volume cleanup for same checkout."* | Solved the same checkout re-cleaning. Two checkouts sharing one volume set was never in scope. |

So:

- **The digest pin is #391's own fix, extended to the side it never reached.** They diagnosed a moving base and stabilised it for the builder; this stabilises it for everyone downstream.
- **The client-side prune is #525, written for the other disk.** Same policy shape, same reasons.
- **The deps image needs no new infrastructure** — CI has published it since #391. Only the consumer was missing.
- **The seed guard keeps #390's intent and narrows its blast radius.** A prebuilt still accelerates a first run; it just stops clobbering a workspace that has a build tree of its own.
- **The repo-id marker from 4c71ddca stays.** It simply no longer triggers a wipe, because per-checkout volumes make the hazard it guarded against impossible.
- **The geometry hash follows the direction already taken twice**: `1e2bd5ca` moved the geometry from a pinned tarball to a content-addressed layer fetched over plain HTTPS, and `777d0692` moved the viewer assets off `type: image` mounts onto disk keyed to a layer digest. This adds the second, narrower key those markers needed to tell a rename from a real change.

The one place this argues *against* a prior decision is `image = "local"`, which resolves to an empty image string and so silently disables the whole prebuilt-and-deps path in favour of a mutable `:latest`. Left alone here and documented in the template, since changing it is a semantics change worth its own discussion.

---

# Where this lands

| | before | after |
|---|---|---|
| standard user, per update | 1437 MB, kept forever | ~65 MB, predecessor pruned |
| developer, per commit | 1437 MB, or a 25–30 min build | 0 |
| second checkout | wrong code, or a full rebuild | own warm workspace; clear refusal on port clash |
| one-file edit after a clean `up` | >20 min, 23 packages | 16–20s, incremental |
| fork editing the driver | `up` dies | warning, keeps running |
| images on disk | 19.31 GB, 13.07 GB reclaimable | 6.72 GB |
| a long step | a clock | phase name + live log tail |

# Tested

End to end on a real machine, on this branch:

- **The deps adoption**, on a dirty tree: the ROS tag 404s, `deps-053cc128…` is pulled, no local Docker build. The retag holds — `innate-os-sim-deps:deps-…` and `innate-os-sim-clean-innate:inputs-…` share one image ID, so two names cost one copy.
- **The publish path** ran on the branch: `deps-<hash>` is published and properly multi-arch (amd64 + arm64).
- **Layer sharing**, against the live registry: all 23 deps layers inside the ROS prebuilt's 28; 1437 MB → 64.6 MB.
- **Per-checkout stacks**: two fully-built workspaces coexisting, each volume stamped with its own checkout id.
- **The seed guard**: a fixture covering fresh / current / stale / reverted states picks the seed only when `build/` is empty, and the real round trip that used to reseed now takes 16s.
- **The step frame**, under a real pty: phase label follows the work, tail rows render in the gutter, formatter-driven steps stay single-line, an empty tail renders as one line.
- ruff, the shell and yaml parsers, the jq retention policy, and the existing suite.

# Out of scope

Seeding the prebuilt image's colcon **build tree** alongside its `install/`. With the seed guard above this now only affects a genuinely fresh workspace — the *first* local build after a first run, which has no build tree to be incremental against. It needs the image built at the workspace's final path and its `--symlink-install` flags reconciled with the local build's, so it wants a CI round trip of its own.

# Upgrade note

The first `up` after this removes the shared `innate-dev` container (it holds 443/80/9090) and builds this checkout's workspace once. The old `innate-os_*` volumes are left alone and named in the log, so they can be reclaimed deliberately rather than silently.

### DIFF
--- a/.github/workflows/publish-sim-images.yml
+++ b/.github/workflows/publish-sim-images.yml
@@ -61,6 +61,7 @@ jobs:
     outputs:
       image_tag: ${{ steps.image_tags.outputs.image_tag }}
       image_inputs_hash: ${{ steps.image_tags.outputs.image_inputs_hash }}
+      deps_inputs_hash: ${{ steps.image_tags.outputs.deps_inputs_hash }}
       assets_inputs_hash: ${{ steps.image_tags.outputs.assets_inputs_hash }}
       assets_published: ${{ steps.image_tags.outputs.assets_published }}
       push_main_tags: ${{ steps.image_tags.outputs.push_main_tags }}
@@ -82,6 +83,11 @@ jobs:
           # Shared with cleanup-sim-images.yml, which pins main's
           # inputs-<hash> versions against retention.
           image_inputs_hash="$(python3 sim/launcher/config.py sim-image-hash)"
+          # The deps image's own address: sim/Dockerfile plus the apt lists,
+          # with no ros2_ws/src in it, so this tag stands still while the ROS
+          # image's moves per commit. That standing-still is the point -- it is
+          # what the launcher pulls instead of a 25-30 min local build.
+          deps_inputs_hash="$(python3 sim/launcher/config.py deps-image-hash)"
           assets_inputs_hash="$(python3 sim/launcher/config.py assets-image-hash)"
           short_sha="${GITHUB_SHA::12}"
           # Every build is content/commit-addressed (sha-, inputs-); only
@@ -103,6 +109,7 @@ jobs:
 
           echo "image_tag=${image_tag}" >> "$GITHUB_OUTPUT"
           echo "image_inputs_hash=${image_inputs_hash}" >> "$GITHUB_OUTPUT"
+          echo "deps_inputs_hash=${deps_inputs_hash}" >> "$GITHUB_OUTPUT"
           echo "assets_inputs_hash=${assets_inputs_hash}" >> "$GITHUB_OUTPUT"
           echo "assets_published=${assets_published}" >> "$GITHUB_OUTPUT"
           echo "push_main_tags=${push_main_tags}" >> "$GITHUB_OUTPUT"
@@ -139,7 +146,7 @@ jobs:
             --gcs-source-staging-dir="${{ vars.GCP_CLOUD_BUILD_SOURCE_STAGING_DIR }}" \
             --service-account="projects/${{ vars.GCP_PROJECT_ID }}/serviceAccounts/${{ vars.GCP_CLOUD_BUILD_SERVICE_ACCOUNT }}" \
             --config=ci/cloudbuild.sim-images.yaml \
-            --substitutions="_IMAGE_PREFIX=${{ vars.SIM_IMAGE_PREFIX }},_IMAGE_TAG=${{ needs.tags.outputs.image_tag }},_IMAGE_INPUTS_HASH=${{ needs.tags.outputs.image_inputs_hash }},_PUSH_MAIN_TAGS=${{ needs.tags.outputs.push_main_tags }},_GHCR_USERNAME_SECRET=${{ vars.GHCR_USERNAME_SECRET }},_GHCR_TOKEN_SECRET=${{ vars.GHCR_TOKEN_SECRET }},_COMMIT_SHA=${GITHUB_SHA},_SHORT_SHA=${{ needs.tags.outputs.short_sha }}"
+            --substitutions="_IMAGE_PREFIX=${{ vars.SIM_IMAGE_PREFIX }},_IMAGE_TAG=${{ needs.tags.outputs.image_tag }},_IMAGE_INPUTS_HASH=${{ needs.tags.outputs.image_inputs_hash }},_DEPS_INPUTS_HASH=${{ needs.tags.outputs.deps_inputs_hash }},_PUSH_MAIN_TAGS=${{ needs.tags.outputs.push_main_tags }},_GHCR_USERNAME_SECRET=${{ vars.GHCR_USERNAME_SECRET }},_GHCR_TOKEN_SECRET=${{ vars.GHCR_TOKEN_SECRET }},_COMMIT_SHA=${GITHUB_SHA},_SHORT_SHA=${{ needs.tags.outputs.short_sha }}"
 
   arm64:
     needs: tags
@@ -168,6 +175,7 @@ jobs:
           IMAGE_PREFIX: ${{ vars.SIM_IMAGE_PREFIX }}
           IMAGE_TAG: ${{ needs.tags.outputs.image_tag }}
           IMAGE_INPUTS_HASH: ${{ needs.tags.outputs.image_inputs_hash }}
+          DEPS_INPUTS_HASH: ${{ needs.tags.outputs.deps_inputs_hash }}
           PUSH_MAIN_TAGS: ${{ needs.tags.outputs.push_main_tags }}
           COMMIT_SHA: ${{ github.sha }}
           SHORT_SHA: ${{ needs.tags.outputs.short_sha }}
@@ -197,6 +205,7 @@ jobs:
           IMAGE_PREFIX: ${{ vars.SIM_IMAGE_PREFIX }}
           IMAGE_TAG: ${{ needs.tags.outputs.image_tag }}
           IMAGE_INPUTS_HASH: ${{ needs.tags.outputs.image_inputs_hash }}
+          DEPS_INPUTS_HASH: ${{ needs.tags.outputs.deps_inputs_hash }}
           PUSH_MAIN_TAGS: ${{ needs.tags.outputs.push_main_tags }}
           SHORT_SHA: ${{ needs.tags.outputs.short_sha }}
         run: bash ci/stitch_sim_manifests.sh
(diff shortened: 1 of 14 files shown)


axelpey (approved):
LGTM

PR #687, opened 2026-08-21
UI: agent page refresh

ugtthis:
- Visual refresh of the Agent page and controls.
- New right-side Agent panel with improved Start, Stop, Reset, and agent selection.
- Reworked chat with Compact/Detailed modes, scrollback, and grouped skill calls.
- Updated microphone UI and hold-to-talk behavior.
- Added simulator scene controls, placement tools, and 4:3 layout.
- Refreshed camera switching, telemetry, challenges, icons, colors, and spacing.

# Compare UI

https://github.com/user-attachments/assets/1005fb95-062e-4629-b95d-fd40072bb3bc


webapp/js/agent/agentPanel.js
```diff
@@ -345,15 +681,17 @@ export function createAgentPanel(root, rosClient, agentState, opts) {
...
     setThoughtsStatus(true);
-    snapIfAtBottom(wasAtBottom);
+    settleStreamAfterAppend(wasAtBottom);
   }
 
   /** @param {string} kind @param {string} text @param {number} ts @param {string} [label] */
   function addMessage(kind, text, ts, label) {
     const wasAtBottom = atBottom();
     finalizeThoughts();
+    if (kind === "user" || kind === "robot") skillStreak = null;
```
DavidDobas: `skillStreak` is only cleared for `user`/`robot` messages, but thought blocks go through `addThought` and never reset it.

Since `promoteSkillStreak` hoists the group to the **first** run's position and moves later runs into it, a sequence like `skill A -> thought -> skill A` renders the second run *above* the thought. The transcript ends up out of order whenever the agent thinks between two runs of the same skill.

Clearing the streak when a thought block is finalized should be enough.

webapp/js/agent/agentMicControl.js
```diff
@@ -49,167 +53,286 @@ export function createAgentMicControl(root, callbacks) {
...
-  function onPointerRelease() {
-    holdSources.delete("pointer");
-    syncHoldState();
+  function onMicPointerCancel() {
+    cancelPendingPointerHold();
+    activeHoldSources.delete("pointer");
+    syncPushToTalk();
+  }
+
+  function spacebarCanControlMic() {
```
DavidDobas, line 231: `composerInput.value.length !== 0` short-circuits before the focus check, so **any** text sitting in the composer disables spacebar push-to-talk page-wide — even when the composer isn't focused.

Leave a half-typed draft and walk away, and the spacebar silently stops working while the hint still reads "hold down your spacebar". The draft-guard only needs to apply when the composer actually has focus:

```js
if (document.activeElement === composerInput) return composerInput.value.length === 0;
return !isTypingContext();
```

DavidDobas (commented):
Nice refresh — the diff is clean (no debug leftovers, no dead code, no dangling refs to the deleted preview modules, and `tsc --noEmit` passes for both `webapp` and `sim/viewer`). Two user-visible bugs I'd fix before merge, inline below.

DavidDobas (approved):
Approving — the diff is clean and the visual work is solid.

The two bugs I flagged above are real but small, and one of them (`skillStreak` not resetting on thought blocks) sits exactly in the code being extracted, so I'll carry both fixes in the follow-up refactor PR that splits `agentPanel.js` into `skillFormat.js` / `chatStream.js` / `directiveControls.js` rather than churn this diff further.

Worth squash-merging so the branch name stays out of main.

webapp/js/agent/agentPanel.js
DavidDobas: Fixed in `refactor/agent-panel-modules` (f3bbcfa6), which also carries the `agentPanel.js` split — this code now lives in `chatStream.js`.

A thought block opening ends the streak, so promotion only ever groups runs that are actually contiguous. Runs separated by thinking no longer collapse into a group, which is the right trade: contiguity is the invariant the hoisting depends on.

Leaving this thread open until that PR lands.

webapp/js/agent/agentMicControl.js
DavidDobas: Fixed in `refactor/agent-panel-modules` (f3bbcfa6).

The focus check now comes first, so the draft-guard only applies while the composer actually has focus:

```js
if (!(composerInput instanceof HTMLTextAreaElement)) return false;
if (document.activeElement === composerInput) return composerInput.value.length === 0;
return !isTypingContext();
```

An unsent draft no longer disables push-to-talk page-wide.

Leaving this thread open until that PR lands.

DavidDobas:
Pushed some follow-up work onto this branch, plus `main` merged in — summarising so it's clear what's new versus the original redesign.

### Refactor

`agentPanel.js` had grown to 1291 lines while every sibling in `js/agent/` sits between 5 and 12KB. Split into three modules:

| Module | Lines | Role |
|---|---|---|
| `agentPanel.js` | **324** (was 1291) | shell, header, composer, ROS wiring |
| `chatStream.js` | 607 | the transcript and all its state |
| `directiveControls.js` | 288 | agent picker, Start/Stop, Reset |
| `skillFormat.js` | 190 | pure formatting helpers |

The transcript deliberately stays whole rather than splitting further — scroll pinning, the open thought group, live skill runs and the streak are mutually entangled (one message can touch all four), so threading them between modules would add more coupling than it removes.

Three things are not pure movement:

- The active-skill chip belongs to the control panel, not the transcript, so `chatStream` reports the running skill out via `onActiveSkill` and tracks the name itself instead of reading it back off `textContent`.
- The live `chat_out` handler and `replayEntry` had identical routing chains; they share `routeChatOut` now so the two paths can't drift.
- History is `replay(entries)` rather than `loadHistory()` — the service call stays with the rest of the ROS wiring, which also makes the transcript testable from a fixture.

### Fixes to the review comments above

Both threads are addressed: the skill-streak ordering bug and the spacebar draft-guard.

### Map on the sim stage

Promoting the map to the big stage showed nothing but a sliver above the frame. The map host is inserted as the cockpit's first child so overlays paint over it by DOM order, and the frame — opaque, added by this PR — was painting straight over it. It now takes the frame's box, inset by the frame's own 1px border so the hairline and corners stay the frame's.

Coming back from the big stage then left the thumbnail black. `.cam-tile.live` declared `grid-template-rows` but no columns, so its column was content-sized: the canvas returns still carrying its big-stage width (measured 1245×113 inside a 113px tile), inflates the track, `width: 100%` resolves to the inflated track, and re-measuring confirms it — a stable wrong size that `overflow: hidden` merely clips. Pinned with `minmax(0, 1fr)`. Not map-specific: any oversized tile child could have triggered it.

### Layout and defaults

- The 4:3 feed derived its height from the dock's leftover width with nothing checking it against the stage, so on a wide, short screen (1728×870 → a 1314×986 frame in 834px of room) the bottom edge and both corners fell outside. The frame is now sized from available height and the dock takes what's left, so the chat gets wider instead of the feed being clipped. Below the crossover the dock width still binds — no change for narrower or taller screens.
- The floating controls disagreed: camera strip and scene setup at 14px inside the frame, Inspect Brain flush in the corner, and the challenge dock on a hardcoded `left: 80px` that predated the frame having a gutter (and overlapped scene setup). They share `--agent-overlay-inset` now.
- The Agent page opens on the sim's top view every visit. `reconcile()` already preferred it, but only when there was no valid saved choice — gated on `simControls`, since real robots have no orbit camera.

### Checks

`main` is merged in (`mergeable`). `tsc --noEmit` at main's exact baseline with no new errors in any touched file; 8/8 tests pass, including the `preload` module-graph test, which needed the three new modules added to `index.html`'s modulepreload list.

Worth squash-merging so the branch name stays out of main.

PR #692, opened 2026-08-22
feat(teleop): keyboard shortcut to use text to speech

ugtthis:
- "/" takes you to input
- "ctrl" takes you to the joystick
- easier way to drive and quickly switch to text to speech feature rather than using your mouse

*credit to axel for pointing out this flaw while he was messing around with mars

# UI

https://github.com/user-attachments/assets/aca00a85-ee63-4dc2-8cf4-de95165637c4


ugtthis:
I pushed from my fork. Refer to https://github.com/innate-inc/innate-os/pull/693

PR #694, opened 2026-08-23
feat(agent): react faster to residents during search

axelpey:
## Summary
- let each agent configure idle and running-skill visual observation cadence
- preempt search navigation when a resident becomes visible
- leave already-handled residents before re-interrupting them
- exclude handled resident locations from frontier utility
- score frontier information against route cost and discourage immediate backtracking
- hold global search after a new resident has `NOTE_MISSING` until the interaction is saved or the guard safely expires

## Why
The original 20-run baseline had a valid maximum of 873.4 seconds. Slow traces showed four generic sources of wasted time: delayed visual reconsideration while navigating, revisiting handled people, route scoring that rewarded long travel, and abandoning a newly found resident after a missed reply.

## Six-round validation
Each round used 20 fresh `e2-standard-8` GCP Batch VMs in parallel, one trial per VM, zero retries, a 900-second challenge limit, and retained video, brain, navigation, telemetry, and result artifacts.

| Round | Main change | Valid | Median | P95 | Max valid |
| --- | --- | ---: | ---: | ---: | ---: |
| R1 | departure guard | 17 | 383.8s | 860.3s | 860.3s |
| R2 | handled-person frontier exclusion | 19 | 283.8s | 528.8s | 528.8s |
| R3 | appearance cache | 19 | 298.1s | 771.1s | 771.1s |
| R4 | route-aware information gain | 18 | 304.8s | 587.9s | 587.9s |
| R5 | frontier anti-backtracking | 17 | 290.0s | 428.7s | **428.7s** |
| R6 | interaction guard | **20/20** | 337.8s | 476.7s | 595.7s |

R5's three non-successes were manually confirmed infrastructure failures (skills-server crash after collecting all orders, permanent localization loss, and unresponsive planner), not agent mistakes. R6 was the only completely clean 20/20 campaign. The ineffective R3 appearance cache was removed before R4.

## Checks
- simulator image workflow: amd64, arm64, manifest, and Greptile checks green
- 11 focused departure/interaction guard tests pass in the simulator container
- 29 orchestrator/manifest/worker offline tests pass before every campaign
- Python compile, formatting checks, exact source-tree attestation, and immutable image/source hashes verified before submission

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/agents/loader.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/agents/loader.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 import base64
 from pathlib import Path
 
-from brain_client.agents.types import Agent, TurnIntervals
+from brain_client.agents.types import Agent, DepartureGuard, TurnIntervals
 from brain_client.common.dynamic_loader import class_name_to_snake_case, evict_modules_under
 from brain_client.common.script_paths import (
     classify_source,
@@ -93,6 +93,12 @@ def build_agent_instances(
             if not isinstance(intervals, TurnIntervals):
                 raise TypeError(
                     f"{type(agent).__name__}.get_turn_intervals() must return TurnIntervals, got {intervals!r}"
+                )
+            departure_guard = agent.get_departure_guard()
+            if departure_guard is not None and not isinstance(departure_guard, DepartureGuard):
+                raise TypeError(
+                    f"{type(agent).__name__}.get_departure_guard() must return DepartureGuard or None, "
+                    f"got {departure_guard!r}"
                 )
             agent.uses_gaze()
             skill_ids = agent.skill_ids()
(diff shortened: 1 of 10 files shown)


axelpey:
20-run GCP comparison on exact commit `23887ece` (campaign `pr694-fast-observation-23887ece-20260823-20`):

- 17/17 behaviorally valid trials passed. Three runs were infrastructure-invalid: r09 brain reset unavailable before trial, r19 simulator never sustained LIVE, and r03 had no localized pose because grid_localizer/AMCL failed.
- Valid pass timing: min 295.5s, median 347.9s, max 873.4s, p90 660.0s. Prior valid baseline: min 248.9s, median 379.0s, max 497.3s, p90 484.8s.
- Median improved by 31.1s (8.2%), but max and p90 regressed badly.
- Tail cause: repeated interruptions for already-known residents. r06 made 8 KNOWN_PERSON checks and 4 IDENTITY_UNAVAILABLE checks; r11 likewise made 8 KNOWN_PERSON and 6 IDENTITY_UNAVAILABLE checks. Prompt context did not reliably enforce the same-view/known-person guard.

Conclusion: faster supervision is promising, but this prompt-only interruption policy should not merge as-is. The next version needs a deterministic known-person cooldown/dedup gate before allowing another navigation interruption.

theo-michel:
Please don't merge this PR it's modifying the brain for the challenge in particular

PR #697, opened 2026-08-24
feat(brain): fold the arm to rest when an agent starts

DavidDobas:
Starting an agent hands the robot to the model, but not in any known shape: an arm left extended by teleop or by the previous session stays extended, and the agent's first navigation skill drags it into the nearest doorframe.

Activation now folds it, the way Mad speed mode braces it — one `/mars/arm/goto_js_v2` call from the node (`fold_arm_for_mad_mode` in `mars_control/app.cpp`), not the `arm_rest_position` skill. The skill would occupy the single execution slot the agent's own first turn needs, and it blocks; this is fire-and-forget, so the loop starts thinking while the arm moves.

Two details carried over from the mechanisms it borrows from:

- **j6 carries the standing grip target**, read from the latched `/mars/arm/command_state`, rather than the rest pose's own value. j6 is current-based position control, so commanding it above that target zeroes the preload and drops a held object — the same reason `mars_control` subscribes to that topic for the Mad fold.
- **The goto is re-commanded once on completion**, as `Manipulation.rest()` does: the fold shifts a held load, and the servos need the second command to settle under it.

The pose itself is `Manipulation.REST`, imported rather than re-typed, so the auto-fold cannot drift from the skill.

The fold is skipped when a skill already owns the arm (a manual run from the webapp), and an arm service that is not up is a warning, not a failure — activation proceeds either way. It hangs off `activate_brain` after `brain.start()` succeeds, so a rolled-back activation does not move the arm. Deactivation is untouched.

## Verification

`ruff check` and `ruff format --check` are clean.

**Not verified:** the pytest suite and `basedpyright` need a ROS environment, which this machine does not have (no `rclpy`, and Docker was not running for the sim container) — both want a run there. Run anyway, basedpyright reports nothing on the new file beyond the unresolved-ROS-import errors every module in the package shows without ROS on the path. The fold has not been watched on hardware or in the sim.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/core/lifecycle.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/core/lifecycle.py
@@ -27,6 +27,7 @@ class BrainLifecycle:
         runner,
         gaze,
         chat,
+        rest_pose,
         active_inputs_pub,
         stop_robot,
         publish_status,
@@ -40,6 +41,7 @@ class BrainLifecycle:
         self._runner = runner
         self._gaze = gaze
         self._chat = chat
+        self._rest_pose = rest_pose
         self._active_inputs_pub = active_inputs_pub
         self._stop_robot = stop_robot
         self._publish_status = publish_status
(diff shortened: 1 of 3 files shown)


DavidDobas:
Superseded by #699 — the branch was renamed to `chore/agent-auto-rest-pose`.

PR #700, opened 2026-08-24
feat(brain): give the agent the wall clock, settable per robot

DavidDobas:
## Why

The agent's only temporal signal was uptime since activation:

```
[t+123s] pose: x=1.20m y=0.30m heading=45° | battery: 87%
```

So it can't tell morning from midnight, can't answer "what time is it", and can't situate the **absolute** timestamps memory search already hands it — `_frame_label` prints `recorded 2026-08-24 14:30` and verdicts say `recorded 3h ago`, with no "now" to compare against.

## What

The status line now leads with the local date and time:

```
[Mon 2026-08-24 14:32 PDT | t+123s] pose: x=1.20m y=0.30m heading=45° | battery: 87%
```

Plus a `timezone` parameter, exposed in **Settings → Brain client → Clock**, applied **live** (no restart).

## Design notes

**Ambient, not a `get_current_time` skill.** While a skill runs, `build_tools` collapses the tool list to `stop_current_skill` alone — a clock skill would be uncallable during navigation, exactly when elapsed time matters. It would also cost a full turn round-trip, against ~10 tokens on a line that already changes every turn (`t+Ns`), so there is no prompt-cache cost either.

**Uptime stays.** It is what separates "just woke" from "been on all day", and it is the cheap relative delta across history turns.

**Minute precision.** At a 3 s idle interval, seconds churn every turn and read as something to act on. The weekday is there so a directive can say "weekday mornings".

**An unsynced clock prints nothing.** Year < 2025 (no RTC battery, no network yet at boot) drops the clock and falls back to plain `[t+5s]`; the prompt tells the model to say it does not know the time rather than reading 1970 aloud. One more prompt rule keeps it from narrating the time every turn, mirroring the existing battery rule.

**Timezone is set per robot, not derived from the browser.** The robot OS ships on `Etc/UTC`:

```
Time zone: Etc/UTC (UTC, +0000)
System clock synchronized: yes
```

The clock is accurate, only the zone is wrong — so the agent has to be told where the robot physically stands. Deliberately *not* taken from `Intl.DateTimeFormat()`: that reports the **operator's** zone, which is wrong the moment someone teleoperates from another continent. The Settings dropdown is a shortlist plus the existing "Custom…" free-text field (the full IANA set is ~600 names). Default stays `""` = follow the host, so a fleet that provisions `/etc/localtime` properly needs no override.

**Boot and Settings treat a bad zone differently.** Construction warns and falls back to host-local — refusing to boot over a typo'd zone is worse than showing the wrong clock. A Settings write is *rejected* with `unknown timezone 'Foo/Bar'`, so the UI reports it instead of silently doing nothing.

**One structural change:** `add_on_set_parameters_callback` moved to after `_build_collaborators()`. It was registered before, fine when the only live knob touched `_tts_handler` (assigned just above), but the timezone branch touches `self.brain`, which did not exist yet. No collaborator calls `declare_parameter`, so nothing is skipped by moving it.

## Verification

- `ruff check` + `ruff format --check`: clean, 96 files
- pytest: **274 passed**. The `agent_factory` fixture's fake config needed the new field — an existing test the change legitimately invalidated, not a new one.
- basedpyright: 14 errors, all pre-existing, **none in touched files**
- webapp `tsc --noEmit`: **5177 errors with and without this change** (verified by stashing the catalog edit and re-running); the two in `settings/main.js` are pre-existing
- Settings YAML round-trip: set → `timezone: America/Los_Angeles`, set-empty → `timezone: ''`, clear → template line re-commented, overrides empty
- Live switch on a real `BrainAgent`: `set_timezone("America/Los_Angeles")` → `Mon 2026-08-24 14:26 PDT`; a rejected write leaves the previous zone intact; `""` returns to host-local

## Follow-up

This is the prerequisite for agent-settable alarms — without a clock in context the agent cannot do "7:30" at all, only relative delays. That work is scoped but not started; it needs a process-lifetime owner with disk persistence, since an alarm has to outlive the turn, the skill, a directive switch, brain deactivation, and a reboot.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
@@ -6,6 +6,8 @@ from __future__ import annotations
 
 import math
 from dataclasses import dataclass
+from datetime import datetime, tzinfo
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 from brain_client.common.enums import StrEnum
 from brain_client.perception import pose as pose_math
@@ -57,8 +59,42 @@ class TraceEvent(StrEnum):
     SNAPSHOT = "snapshot"
 
 
+# An unsynced clock (no RTC battery, no network yet at boot) reads 1970, and a
+# confidently wrong time misleads the agent worse than no time at all.
+_CLOCK_VALID_FROM_YEAR = 2025
+
+
+def resolve_timezone(name: str) -> tzinfo | None:
+    """A named IANA zone, or None for the system's local zone.
+
+    An unknown name or missing tzdata falls back to local — a misconfigured
+    zone must not take the brain down with it. Local stays None rather than
+    resolving to a tzinfo here: a snapshot of the host's current offset would
+    freeze the brain on one side of a DST change.
+    """
+    if not name.strip():
+        return None
+    try:
+        return ZoneInfo(name)
+    except (ZoneInfoNotFoundError, ValueError):
+        return None
+
+
+def clock_text(now: datetime) -> str:
+    """The wall clock as the status line shows it, or "" when it can't be trusted.
+
+    Minute precision: at a 3 s turn interval, seconds are noise that reads as
+    something to act on. The zone abbreviation is what makes a robot left on
+    UTC visible in the transcript instead of silently wrong.
+    """
+    if now.year < _CLOCK_VALID_FROM_YEAR:
+        return ""
+    return f"{now:%a %Y-%m-%d %H:%M %Z}".strip()
+
+
 def observation_text(
     *,
+    now: datetime,
     uptime_s: int,
     pose: Pose | None,
     battery: Battery | None,
@@ -68,7 +104,8 @@ def observation_text(
     has_wrist_frame: bool,
 ) -> str:
     """The text half of a turn input: robot status, guidance, and new events."""
-    status = f"[t+{uptime_s}s]"
+    clock = clock_text(now)
+    status = f"[{clock} | t+{uptime_s}s]" if clock else f"[t+{uptime_s}s]"
     if pose is not None:
         status += f" pose: x={pose[0]:.2f}m y={pose[1]:.2f}m heading={math.degrees(pose[2]):.0f}°"
     if battery is not None:
--- a/ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
@@ -59,19 +59,14 @@ class TraceEvent(StrEnum):
     SNAPSHOT = "snapshot"
 
 
-# An unsynced clock (no RTC battery, no network yet at boot) reads 1970, and a
-# confidently wrong time misleads the agent worse than no time at all.
+# Cold boot before NTP, on a robot whose RTC lost the time: the clock reads 1970.
 _CLOCK_VALID_FROM_YEAR = 2025
+_CLOCK_UNSET = "time not known"
 
 
 def resolve_timezone(name: str) -> tzinfo | None:
-    """A named IANA zone, or None for the system's local zone.
-
-    An unknown name or missing tzdata falls back to local — a misconfigured
-    zone must not take the brain down with it. Local stays None rather than
-    resolving to a tzinfo here: a snapshot of the host's current offset would
-    freeze the brain on one side of a DST change.
-    """
+    """An IANA zone by name, or None for the host's own — left unresolved so a
+    DST change is picked up rather than frozen at boot."""
     if not name.strip():
         return None
     try:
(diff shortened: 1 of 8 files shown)


ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
```diff
@@ -57,8 +59,42 @@ class TraceEvent(StrEnum):
     SNAPSHOT = "snapshot"
 
 
+# An unsynced clock (no RTC battery, no network yet at boot) reads 1970, and a
+# confidently wrong time misleads the agent worse than no time at all.
+_CLOCK_VALID_FROM_YEAR = 2025
+
+
+def resolve_timezone(name: str) -> tzinfo | None:
```
ugtthis, line 67: The numbers of lines in this docstring is as much as the code. Can we make the code more self documenting so 6 lines of docstrings arent needed to explain 6 lines of code if possible

ugtthis (approved):
if the changes can be made then lgtm

ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
```diff
@@ -57,8 +59,42 @@ class TraceEvent(StrEnum):
...
+    if not name.strip():
+        return None
+    try:
+        return ZoneInfo(name)
+    except (ZoneInfoNotFoundError, ValueError):
+        return None
+
+
+def clock_text(now: datetime) -> str:
+    """The wall clock as the status line shows it, or "" when it can't be trusted.
```
ugtthis: Maybe rm the 1st line of the docstring.  Do we really need it to explain line 90 -> ` if now.year < _CLOCK_VALID_FROM_YEAR:  return ""`

ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
```diff
@@ -57,8 +59,42 @@ class TraceEvent(StrEnum):
...
+
+    Minute precision: at a 3 s turn interval, seconds are noise that reads as
+    something to act on. The zone abbreviation is what makes a robot left on
+    UTC visible in the transcript instead of silently wrong.
+    """
+    if now.year < _CLOCK_VALID_FROM_YEAR:
+        return ""
```
ugtthis: Having "" doesn't seem like a helpful error message. Can we make it show N/A or something similair so we can at least know something is happening versus it failing silently

ugtthis (approved):
some more things i forgot to mention

ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
DavidDobas: Done — `resolve_timezone` is 6 docstring lines down to 2:

```python
def resolve_timezone(name: str) -> tzinfo | None:
    """An IANA zone by name, or None for the host's own — left unresolved so a
    DST change is picked up rather than frozen at boot."""
```

The DST sentence is the one thing the code genuinely can't say: returning `None` for the host zone looks like a missing case, but resolving it to a `tzinfo` here would snapshot the current UTC offset and freeze the robot on one side of a DST change. The rest was restating the signature, so it went.

I also dropped the "a misconfigured zone must not take the brain down" line — that invariant is stated once at the call site in `agent.py`, no need to repeat it here.

ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
DavidDobas: Agreed, gone. `clock_text` is now:

```python
def clock_text(now: datetime) -> str:
    """The wall clock for the status line, or a marker the agent can act on:
    a confidently wrong time misleads it worse than a stated unknown does."""
    if now.year < _CLOCK_VALID_FROM_YEAR:
        return _CLOCK_UNSET
    return f"{now:%a %Y-%m-%d %H:%M %Z}".strip()
```

The first line was describing the return type, which the signature already does. What survives is the *why* behind the year check, since that is a judgment call and not visible from the code.

The threshold itself now carries the trigger condition instead of the docstring:

```python
# Cold boot before NTP, on a robot whose RTC lost the time: the clock reads 1970.
_CLOCK_VALID_FROM_YEAR = 2025
```

ros2_ws/src/brain/brain_client/brain_client/brain/utils.py
DavidDobas: Good catch, and you were right that this was the important one — an empty string asks the model to notice something *is not there*, which is exactly what it is worst at. It now prints in the slot where the time would go:

```
[Tue 2026-08-25 14:32 PDT | t+123s] pose: x=1.20m y=0.30m heading=45°
[time not known | t+5s]             pose: x=1.20m y=0.30m heading=45°
```

Went with plain words over `N/A` so it reads without a decoder — the system prompt no longer needs a sentence explaining what the marker means, and `observation_text` loses its branch (the status line is one unbranched f-string again).

**When it can fire:** cold boot before NTP syncs, on a robot whose RTC did not hold the time across the power cycle — dead or missing coin cell, or a freshly flashed board. Our stack boots in under a minute, so the window is real, but a healthy robot never sees it.

**One limitation worth naming:** `year < 2025` only catches the catastrophic case. An RTC holding a plausible-but-wrong time (drifted an hour, or restored from systemd's last-known-good stamp) passes the check, and the agent will state that wrong time confidently. The check is a floor, not a correctness guarantee.

The precise fix is to test whether the clock is *synchronized* rather than whether it is plausible: `systemd-timesyncd` touches `/run/systemd/timesync/synchronized` on its first successful sync, so a `Path.exists()` would answer it exactly. I did not build that because if the Jetson runs chrony instead, that file never appears and every robot would permanently report "time not known" — worse than the gap it closes. If someone can confirm the daemon on a real robot, it is a three-line upgrade and I am happy to take it.

PR #705, opened 2026-08-26
fix(control): keep the drive smoother running when system helpers hang

theo-michel:
## Problem

Teleop wheel commands lagged by tens of seconds on mars-the-47th — even from the local network — while skills responded instantly.

Root cause: `publish_robot_info_callback` (1 s timer) shells out to `avahi-resolve`, `nmcli dev wifi`, `bluetoothctl`, and git (~7 subprocess spawns/sec) on the **same single-threaded executor** as the 50 Hz `drive_smoother_timer_`. When `avahi-daemon` wedged (it had been spinning at ~94% CPU for days), every `avahi-resolve` blocked indefinitely, the executor spent nearly all wall time inside `popen`, and the smoother ticked at ~0.2 Hz. Each tick advances the acceleration ramp by one fixed 20 ms step, so the robot accelerated ~250× slower than designed — turning was effectively dead. Skills publish `cmd_vel` from other nodes and were unaffected.

Measured on the robot: 10 s of continuous `/joystick` input produced **2** `/cmd_vel_teleop` messages; after restarting avahi, **119 at 50 Hz**.

## Fix (three layers)

1. **Isolation** — the two shell-out timers plus all admin services (`/set_robot_name`, `/set_volume`, `/set_microphone`, `/shutdown`, `/trigger_update`) move into a mutually-exclusive `slow_group_` serviced by a second `MultiThreadedExecutor` thread. The drive path never waits on a subprocess again. The default group stays mutually exclusive, so the drive-state no-locking invariant is preserved; the slow group being mutually exclusive keeps `robot_info.json` access serialised as before.
2. **Bounded exec** — `exec_command` now `fork`/`exec`s the shell in its own process group with a hard 10 s deadline (`poll` + `killpg`), so even on the slow thread a hung helper costs bounded staleness, never a wedged callback.
3. **Stop the subprocess churn** — hostname, BT MAC, version/tags, and SSID change rarely; they're now cached together and refreshed every 10 s instead of re-spawned every second (the update check keeps its 300 s cache).

## Verification

In the CI image (`innate-os-test:fullbuild`), with `nmcli` and `bluetoothctl` replaced by `sleep 600` stubs on PATH and `/joystick` published at 7 Hz:

- `/cmd_vel_teleop` holds **~49.7 Hz** (old code: sub-1 Hz in this scenario, ~0.2 Hz observed on the robot)
- `/robot/info` still publishes; the hung helpers' fields are simply absent until they recover
- clang-format via pre-commit: clean; Release build: clean (the one `-Wunused-result` warning is pre-existing untouched code)

## Notes

The wedged avahi-daemon itself was restarted on mars-the-47th during debugging (symptom gone), but this node must tolerate a sick system daemon — that's what this PR makes structural.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
+++ b/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
@@ -111,7 +111,6 @@ std::string exec_command(const std::string& cmd) {
         const auto left =
             std::chrono::duration_cast<std::chrono::milliseconds>(deadline - std::chrono::steady_clock::now());
         if (left <= std::chrono::milliseconds::zero()) {
-            killpg(pid, SIGKILL);
             break;
         }
         pollfd pfd{fds[0], POLLIN, 0};
@@ -120,7 +119,6 @@ std::string exec_command(const std::string& cmd) {
             continue;
         }
         if (ready <= 0) {
-            killpg(pid, SIGKILL);
             break;
         }
         char buffer[256];
@@ -134,7 +132,16 @@ std::string exec_command(const std::string& cmd) {
         result.append(buffer, static_cast<size_t>(n));
     }
     close(fds[0]);
-    waitpid(pid, nullptr, 0);
+    // EOF only proves the last pipeline stage closed stdout; an earlier stage may
+    // still be hung, so the reap honours the same deadline before killing the group.
+    while (waitpid(pid, nullptr, WNOHANG) == 0) {
+        if (std::chrono::steady_clock::now() >= deadline) {
+            killpg(pid, SIGKILL);
+            waitpid(pid, nullptr, 0);
+            break;
+        }
+        poll(nullptr, 0, 10);
+    }
 
     // Trim trailing newline
     while (!result.empty() && (result.back() == '\n' || result.back() == '\r')) {
--- a/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
+++ b/ros2_ws/src/mars_bot/mars_control/mars_control/app.cpp
@@ -38,6 +38,7 @@
 #include <memory>
 #include <regex>
 #include <sstream>
+#include <thread>
 #include <cerrno>
 #include <poll.h>
 #include <signal.h>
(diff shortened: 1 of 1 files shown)


DavidDobas:
@greptileai

DavidDobas:
Reviewed by priority: (1) correct technical solution, (2) clean readable DRY code, (3) realistic edge cases, (4) other bugs. Findings below, each with the fix applied in 3b8d2c9.

## 1) Correct technical solution chosen

**Yes, with one gap in application.** Second executor thread + mutually-exclusive `slow_group_` is the idiomatic ROS 2 fix, and the partition is sound: every callback touching the drive-smoother state stays in the default group (no-locking invariant holds), every `robot_info.json` toucher is in the slow group (file access stays serialised). The `fork`/`setpgid`/`poll`/`killpg` exec rewrite is correct — async-signal-safe child, EINTR handled, no zombie or fd leak on any path.

The gap: **the two most hang-prone shell-outs bypassed the new bounded exec** because they still used `std::system`, so layer 2's invariant ("a hung helper costs bounded staleness, never a wedged callback") didn't hold:

- `check_update_available` → `innate update quick-check` → an untimeouted **network `git fetch`**, called from the slow group every 300 s. Blocks for minutes on a half-open network.
- `set_system_hostname` runs `sudo hostnamectl` then `systemctl restart avahi-daemon.service` — restarting *the daemon whose wedging caused this incident* — and a wedged unit makes that restart block for the systemd job timeout. Aggravating: glibc `system()` ignores `SIGINT`/`SIGQUIT` process-wide while running, so during an unbounded fetch the node could only be killed with SIGKILL.

**Fixed:** `exec_command_status()` (bounded exec returning `{output, exit_code}`) now backs both; `exec_command` is a thin wrapper over it. quick-check gets a 30 s budget, hostnamectl and the avahi restart 10 s each. Timeout/spawn failure reports as exit code −1 and is treated as "no update" / "hostname set failed" rather than garbage `WEXITSTATUS`.

## 2) Clean readable DRY code

Collapsing the per-field caches into `SlowFacts` is a real improvement. Two leftovers:

- `get_tag_subject`, `get_tag_body`, and `get_robot_version` each independently re-ran `git tag --list --sort=-version:refname` — three of the ~7 execs per refresh were the identical command. **Fixed:** `cached_slow_facts()` resolves the latest release tag once and passes it into all three; the `get_latest_tag` pass-through wrapper is deleted.
- `get_cached_update_available` remains a second hand-rolled cache in the old underscore style. **Left as is** — it has a genuinely different interval (300 s vs 10 s) and folding it into `SlowFacts` would couple the 30 s network budget to the fast refresh; not worth the churn.

## 3) Realistic edge cases handled

The healthy path was fine; the degraded path — the exact scenario this PR exists for — had a real hole:

- **The refresh timestamp was stamped before the work.** The 10 s interval measured from refresh *start*, so when helpers hang and one refresh burns its full budget, the skipped 1 Hz timer re-fires immediately and `elapsed < interval` is false → it refreshes again, back-to-back. Steady state under the incident's conditions: the slow thread ~100% occupied shelling out, admin services permanently queued behind subprocesses. **Fixed:** the stamp moved to after the gather, so a hung refresh is followed by a full quiet interval — worst case a ~50% duty cycle instead of 100%.
- **Worst case one refresh was ≈ 70 s** (7 execs × 10 s), all serialised with `/shutdown`, `/set_volume`, `/set_microphone`, `/set_robot_name` — a user pressing shutdown waited behind a polling timer's subprocesses. **Fixed:** default `EXEC_COMMAND_TIMEOUT` dropped 10 s → 2 s (~100× the healthy latency of every local helper), and with the shared tag resolution a worst-case refresh is now ~5 execs × 2 s ≈ 10 s. Callers of genuinely slow commands pass their own budget.
- One behavior to be aware of (not changed): during a hang, `/robot/info` doesn't publish with fields absent — it stops publishing entirely for the hang's duration, then resumes. With the tighter budgets the worst gap is now ~10 s; worth confirming the app doesn't treat that as "robot offline".

## 4) Other bugs

- **Worker-thread exceptions bypassed `main`'s error boundary.** `MultiThreadedExecutor::spin()` runs the extra thread as a raw `std::thread` with no try/catch in `run()`, so a throw from a callback dispatched there hit `std::terminate` instead of the `catch` in `main` — and since the *default* group could be serviced by either thread, this covered the drive callbacks too. **Fixed:** replaced the 2-thread MT executor with two `SingleThreadedExecutor`s — the drive executor owns the node (main thread, original try/catch intact), the slow group (created with `automatically_add_to_executor_with_node=false`) is spun by a dedicated `std::thread` with its own logging boundary that calls `rclcpp::shutdown()` on the way out so one side dying takes the whole node down, not half of it. Side benefit: the drive path now has a truly dedicated thread, and the drive-state no-locking argument gets simpler, not subtler.
- **`publish_robot_info_callback` lost its inner `try`** around the version/tag block — a throw from the fact-gather degraded the whole payload to `{}` instead of just omitting those fields. **Fixed:** restored a narrow inner boundary around the `SlowFacts` block.

## Verification

- Release build of `mars_control` (plus dependents `mars_msgs`, `brain_messages`, `mars_arm`, `mars_bringup`, `mars_sim`) clean in the ROS Humble container, no new warnings.
- `pre-commit` clang-format: clean.

DavidDobas:
@greptileai

DavidDobas (approved):
Implemented fixes, approving

PR #706, opened 2026-08-26
feat(teleop): echo-cancelled full-duplex talkback

DavidDobas:
## What

Lifts talkback's half-duplex limitation: with `enable_echo_cancel:=true`, the operator keeps hearing the robot's mic **while** they talk. The echo that half-duplex guarded against — the robot's speaker being within earshot of its own mic — is cancelled at the capture instead, by WebRTC's Audio Processing Module (`webrtcdsp` + `webrtcechoprobe`, already installed via `gstreamer1.0-plugins-bad`, the package `webrtcbin` comes from).

**Defaults off.** With the parameter false the duck stays and behavior is unchanged; flip it on the bench to validate cancellation quality before making it the default.

## How

AEC needs one reference signal of what the speaker actually plays, so the per-peer playback branches collapse into **one speaker pipeline**, built at startup and left PLAYING (ALSA is dmix; holding the device open costs nothing):

```
appsrc (RTP) → rtpopusdepay → opusdec → volume → [webrtcechoprobe] → queue → alsasink
```

Peers now attach a thin RTP **tap** (`queue → appsink`) on `pad-added` instead of a full decode branch; the tap forwards packets into the shared appsrc only while that peer's talk gate is open. The mic pipeline gains, when enabled:

```
… caps(48k mono) → webrtcdsp echo-cancel=true delay-agnostic=true → opusenc …
```

`delay-agnostic` because capture and playout are separate pipelines on separate clocks, and dmix hides the true playout latency — the canceller estimates alignment itself.

## What this deletes

The valve, and with it the entire gate-snapshot race class from the last review rounds: the gate is now only ever **read**, per packet, on the tap's streaming thread. There is no write-after-read left to interleave, so `TalkGate` shrinks back to an atomic bool. The tap callback takes no locks (`~Peer` joins that thread), matching the existing fan-out discipline.

## Ordering that matters

`webrtcdsp` **hard-errors at start** if its probe doesn't exist (verified live on the robot, not just from docs). Therefore: the speaker pipeline (which owns the probe) is built before the mic pipeline, and `enable_echo_cancel` is forced off whenever talkback is disabled — a mic pipeline with a probe-less DSP can never be built.

## Verified / not verified

- ✅ Builds clean, zero warnings; clang-format clean.
- ✅ The exact speaker chain parses and reaches PLAYING with the probe (gst-launch, on the robot).
- ✅ `webrtcechoprobe` + `webrtcdsp echo-cancel=true delay-agnostic=true` pair and run in one process.
- ❌ Cancellation quality with real speaker/mic acoustics — the reason the default is off.
- ❌ Talkback audio end to end — same standing caveat as the base branch.

## Known limit

Two peers talking simultaneously now interleave RTP into one depayloader and glitch; one talker at a time is the supported shape. (The old design mixed N talkers via dmix, but N sinks cannot host a single AEC probe.)

### DIFF
--- a/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_streamer.hpp
+++ b/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_streamer.hpp
@@ -175,11 +175,14 @@ class WebRTCStreamer : public rclcpp::Node {
     std::string speaker_pipeline_description() const;  // audiomixer -> volume -> [probe] -> speaker
     static std::string talk_decode_description();      // one peer's depay -> decode -> PCM appsink
     bool build_speaker_pipeline();                     // built + PLAYING at startup; peers mix into it
+    void teardown_speaker_pipeline();                  // only with no peers alive (they hold inputs inside it)
+    void restart_speaker_pipeline();                   // health poll: NULL->PLAYING after a bus error (throttled)
     // Give/take this peer its own appsrc + audiomixer pad. Both run on the executor thread with
     // peers_mutex_ held; detach is safe only after ~Peer has joined the thread that pushes.
     bool attach_speaker_input(Peer* peer);
     void detach_speaker_input(GstElement* src, GstPad* pad);
-    void set_peer_talk(Peer* peer, bool on);  // flip the gate; caller holds peers_mutex_
+    void set_peer_talk(Peer* peer, bool on);         // flip the gate; caller holds peers_mutex_
+    bool peer_audio_active(const Peer& peer) const;  // listen intent minus the talk duck (lifted by AEC)
     void reconcile_talk();  // publish /talkback/is_playing when the last talker stops (or the first starts)
 
     // ---- Shared audio (mic) — encoded once like the cameras, fanned out, gated for privacy ----
@@ -269,6 +272,7 @@ class WebRTCStreamer : public rclcpp::Node {
     GstElement* talk_mixer_ = nullptr;        // ref'd; peers request their input pads from it
     std::atomic<int> want_talk_{0};           // # peers holding the talk button; >0 => the robot is audibly speaking
     bool talk_playing_ = false;               // last published /talkback/is_playing state
+    int64_t last_speaker_restart_ns_ = 0;     // restart_speaker_pipeline backoff
 
     // ---- Peers ----
     std::map<std::string, std::unique_ptr<Peer>> peers_;
--- a/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_streamer.hpp
+++ b/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_streamer.hpp
@@ -84,10 +84,15 @@ struct Peer {
     bool audio_requested = false;            // the operator's LISTEN intent; audio_active is it minus the talk duck
     bool with_talk = false;                  // the audio m-line was negotiated sendrecv (talkback possible)
     bool talk_active = false;                // operator is holding talk: their mic plays out the robot's speaker
-    // Privacy gate: the peer's RTP tap forwards to the speaker only while this is true. Read per sample
-    // on the peer's streaming thread (via a copy tagged on the tap), which must NOT take peers_mutex_ —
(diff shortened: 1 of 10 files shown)


DavidDobas:
## The problem

The "Known limit" at the bottom of the description — one talker at a time — gave up more than it looks, and it wasn't necessary.

Every talking peer's RTP landed in **one** `rtpopusdepay`, so two operators holding talk meant two SSRCs, two sequence-number spaces and two timestamp bases interleaved into a single depayloader that has no way to demux them. Three consequences:

- **Multi-operator talkback is gone.** That's the property the `client_id` work bought — peers stopped having to take over the stream from each other. Nothing enforced the new one-talker rule either: two gates could be open and `want_talk_` counted past 1 happily.
- **Handoffs, not just simultaneity.** Even strictly sequential talkers (A speaks, releases, B speaks) push different SSRCs through the same depayloader, so every handoff crosses a sequence discontinuity.
- **Loss concealment died quietly.** The tap pulled buffers from an appsink and pushed buffers into an appsrc — GStreamer *events* don't survive that hop, so the jitterbuffer's GAP events never reached `opusdec` and `plc=true` had nothing to conceal with. That compounds, because the tap's own `max-buffers=8 drop=true` is a loss source.

The reasoning in the description — "N sinks cannot host a single AEC probe" — is true. The conclusion doesn't follow: AEC needs one **sink**, not one **decoder**.

## The solution

Put an `audiomixer` in front of the speaker and decode per peer:

```
before: peer A ─┐
        peer B ─┴─► one appsrc ─► rtpopusdepay ─► opusdec ─► volume ─► probe ─► alsasink
                    two SSRCs, one depayloader

after:  peer A ─► depay ─► opusdec plc=true ─► caps ─┐
        peer B ─► depay ─► opusdec plc=true ─► caps ─┴─► audiomixer ─► volume ─► probe ─► alsasink
                    each in its own peer pipeline        one sink, one probe, N talkers
```

Each peer depayloads and decodes inside its own transport pipeline and hands PCM across at fixed caps (S16LE/48k/mono, so every mixer input agrees and the probe hands `webrtcdsp` the shape it expects). The probe does not move, so the AEC pairing is unchanged — that was the main risk, and it holds.

What comes back: N simultaneous talkers, clean handoffs, and `plc=true` actually concealing loss now that GAP events reach `opusdec` upstream of the appsink→appsrc hop. Each peer's mixer input also carries the `leaky-type=downstream, max-bytes=256k` bound every other appsrc in the node has, so a stalled speaker drops instead of growing.

Pressing talk stays a pure atomic store — no pipeline surgery on the press. A peer takes a mixer input for its whole life, and an idle live input costs nothing: the mixer aggregates on the clock and reads it as silence. Inputs attach after every failure return in `create_peer_transport` (so none can be orphaned) and release in `destroy_peer` *after* the erase, since `~Peer` is what joins the thread that pushes into them.

## Verified on the robot's GStreamer (1.20.3)

- Two live appsrcs into one `audiomixer` **sum** (1000 + 2000 → 3000) — genuinely mixing, not interleaving.
- An attached-but-idle input does not stall the mix.
- Add/remove of an input and its request pad while PLAYING works; the mix continues and recovers.
- End-to-end rehearsal with the exact strings the node builds — two peers decoding real Opus RTP in separate pipelines into the speaker, while a `webrtcdsp echo-cancel=true delay-agnostic=true` mic pipeline runs against that speaker's probe: no bus errors anywhere, output in every talking phase, silence when nobody talks.
- Compiles clean with `-Wall -Wextra`; clang-format clean.

Unverified, as before: talkback audio end to end on hardware, and cancellation quality with real acoustics — the reason `enable_echo_cancel` stays off by default.

## Not addressed

Two review findings about the lifetime-scoped speaker pipeline rather than the mixing shape, unchanged by this commit:

1. It is built once at startup and never rebuilt — an ALSA error or EOS kills talkback until the node restarts, since the health poll only drains its bus to the log.
2. The `!enable_audio_` path that disables talkback leaves the speaker pipeline PLAYING (holding the device) and does not re-clear `enable_echo_cancel_`.

Pushed as 3bd6141e. The "Known limit" paragraph in the description is now stale — N operators can talk at once again — and should be dropped before merge.

PR #664, opened 2026-08-18
feat(webrtc): loss-resilient, network-adaptive teleop video

theo-michel:
Remote teleop video was unusable away from the LAN: multi-second freezes, then (after partial fixes) full-frame corruption. This PR makes the stream **loss-resilient and network-adaptive**: clean networks keep today's lowest-latency behavior; degraded paths trade a little latency and bitrate for a stream that stays watchable. Everything was built and verified live against mars-the-47th from a genuinely bad vantage point (240 ms RTT, 4–6% packet loss — the office uplink's fault, see "Context" at the bottom).

**Headline result, same path, same measurement harness:**

| per ~50 s window | before | after |
|---|---|---|
| Frozen time | 17–29 s | **0.5–0.7 s** |
| PLIs (decoder distress) | 35–99 | **0** |
| fps decoded | 4.6–8.5 | **~15, every received frame** |
| Wire loss | ~4–6% | unchanged — repaired in-band |

## What changed, layer by layer

**1. Repair machinery that was missing or broken (bug-fix tier — benefits every viewer):**
- `webrtcbin` negotiated **no loss repair at all** (gst defaults). Video transceivers now negotiate NACK/RTX + ULPFEC/RED, set at creation time where webrtcbin's internals read them. Trap defused: gst's `fec-percentage` *defaults to 100* (double bitrate), so it's always set explicitly. (`07d50421`, `6da4525c`)
- **Browser keyframe requests never reached the encoder** — PLIs became `GstForceKeyUnit` events that died at the transport appsrc (the encoder lives in a separate pipeline). A pad probe forwards them, throttled to 1 IDR/500 ms per camera so PLI storms can't melt the link. Periodic keyframes drop from 0.5 s to a 4 s backstop: clean links carry almost no keyframes, lossy links get them on demand. (`0d819a83`)
- **VP8 picture IDs were missing** (`picture-id-mode=15-bit`): without them Chrome can't prove frame continuity after loss and silently discards every repaired frame — NACK worked, repairs were thrown away. (`cbfd1506`)

**2. Adaptivity — the "good network = low latency, bad network = usable" contract:**
[...]

PR #707, opened 2026-08-26
fix(webrtc): stop artifacting on tunneled remote links

karmanyaahm:
Follow-up to #664: artifacts persisted on a real remote link. Root-caused live on R7-27 with the
viewer connected over Tailscale (RTT 226–445 ms). Three independent bugs:

## 1. RTP packets didn't fit the path (the dominant cause)

The media path is a WireGuard tunnel: **MTU 1280**, so 1252 usable on the wire. `rtpvp8pay`'s
default `mtu=1400` (+1 RED, +10 SRTP) fragmented **every full-size packet** — and either lost
fragment kills the whole RTP packet. Measured on the live path:

| ICMP payload | loss |
|---|---|
| 56 B | 0.83% |
| 1200 B | 1.67% |
| 1400 B (fragmented) | **2.5%** |

FEC packets are ~14 B *bigger* than the media packets they protect, so the repair traffic
fragmented and died at the same rate. → `mtu=1200`.

## 2. The sender adaptation was loss-blind and flapped on RTT

`on_peer_stats` read `rb-fractionlost` (uint) — that field only exists on webrtcbin's *internal*
input stats, never in the `get-stats` reply (verified by disassembling the shipping
`libgstwebrtc.so`: the reply carries **`fraction-lost` as a double in [0,1)**). So
`rtcp_loss_promille_` never left −1, "viewer loss 0.0%" in the logs was the sentinel, and the
DEGRADED trigger degenerated to `rtt >= 250` — which a tunneled path straddles forever.
Observed: **17 GOOD↔DEGRADED flaps in 8 minutes**, each entry = bitrate cut + 100% FEC +
1 Hz forced keyframes on a link that had no loss spike.

Now: the real loss field is read, loss is the *only* trigger (RTT is a property of the path, not
damage), and an absent report logs `loss n/a` instead of a reassuring `0.0%`.

## 3. Encoder tuning made every recovery keyframe hurt

- `error-resilient=partitions` was a no-op (only splits token partitions; `token-parts` is 1).
  `default` is the actual libvpx loss-resilience mode.
- IDR size was unbounded: a recovery keyframe ate the 600 ms CBR buffer and libvpx answered with
  a quantizer slam on the following frames — the blocky pulsing after every PLI/DEGRADED entry.
  `max-intra-bitrate=300` caps it at 3× the per-frame budget (same neighborhood libwebrtc uses).
- `main_fps` defaulted to 30 but the driver captures 15 (`stereo_depth_estimator.yaml`), so the
  keyframe backstop was silently 8 s and rate control got a doubled per-frame budget.

## Also: duplicate payload types across the BUNDLE

The live offer carried `pt 97 = red(video0)` **and** `VP8(video1)`, `pt 98 = ulpfec(video0)`
**and** `opus` — webrtcbin assigns each m-line's red/ulpfec/rtx PTs sequentially after that
m-line's media PT, and 96/97/98 left no room. RFC 8843 §9.1 forbids duplicates in a bundle;
Chrome tolerates it today via the signaled `a=ssrc` lines, newer Chrome rejects the offer.
Cameras now stride 5 PTs apart (96, 101, …) reserving each m-line's aux block; audio moves
to 127. Updated the two tests that pinned the old scheme.

## Verified

- `colcon build --packages-select mars_cam` clean; `test_webrtc_internal` 7/7 pass.
- New pipeline strings parse and apply on **both** stacks: system GStreamer 1.20.3 and
  `/opt/gst` 1.24.12 (`error-resilient=default`, `max-intra-bitrate=300`, `mtu=1200`, `pt=127`).
- Not yet soak-tested live — needs a container restart on a robot with a remote viewer.

## Test plan

1. `innate restart`, connect the webapp over the remote path, press `p`.
2. Expect: loss well under 1% (no fragmentation), no 30 s GOOD↔DEGRADED cycle in the
   `webrtc_streamer` logs, `viewer loss` showing real numbers (or `n/a`, never a fake 0.0%),
   and no blocky pulse after keyframes.
3. Sanity-check a LAN viewer too: nothing here should change the clean-path behavior beyond
   ~17% more (smaller) packets from mtu=1200.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_internal.hpp
+++ b/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_internal.hpp
@@ -60,11 +60,14 @@ inline guint cam_ssrc_for_index(size_t index) {
     return 0x1A2B3C00u + static_cast<guint>(index) + 1u;
 }
 
-// RTP payload type per camera (96, 97, 99, 100, …) — skips 98, which the audio (opus) payloader uses.
+// PTs must be unique across the whole BUNDLE (RFC 8843 §9.1), INCLUDING the red/ulpfec/rtx/rtx PTs
+// webrtcbin auto-assigns per video m-line — sequentially after that m-line's media PT, at link time.
+// A 5-PT stride (96, 101, 106, …) leaves each m-line its 4 aux slots; audio sits above them all.
 inline int cam_pt_for_index(size_t index) {
-    const int pt = 96 + static_cast<int>(index);
-    return pt >= 98 ? pt + 1 : pt;
+    return 96 + 5 * static_cast<int>(index);
 }
+
+inline constexpr int kAudioPt = 127;
 
 // Chrome/Firefox often obfuscate their host ICE candidates as "<uuid>.local" mDNS names (one per local
 // interface). These parsing helpers are kept for diagnostics and for a future non-destructive fast path:
--- a/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_internal.hpp
+++ b/ros2_ws/src/mars_bot/mars_cam/include/mars_cam/webrtc_internal.hpp
@@ -69,6 +69,10 @@ inline int cam_pt_for_index(size_t index) {
 
 inline constexpr int kAudioPt = 127;
 
+// The dynamic PT range is 96-127; camera 6 would land on 126 and push its aux block onto (and past)
+// kAudioPt, so the stride fits exactly 6 cameras. configure_cameras enforces this.
+inline constexpr size_t kMaxCameras = 6;
+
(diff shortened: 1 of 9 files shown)


karmanyaahm:
Pushed one more commit for the receiver side: the webapp pinned `jitterBufferTarget` to 0 until the first adapt poll at t=3 s, so the first seconds of every connection (and every watchdog rebuild) played out unbuffered — on a ~240 ms path that renders every early loss as macroblocking. Tracks now attach at the last-known target, the poll runs at 500 ms until the first RTT sample lands, and rebuilds carry the learned target over. `tsc --noEmit` unchanged (0 errors in webrtcSession.js).

theo-michel:
Closing to reopen under the author's account — same branch, review fixes included. Continued in the superseding PR.

theo-michel:
Superseded by #708.

PR #708, opened 2026-08-26
fix(webrtc): stop artifacting on tunneled remote links

theo-michel:
Follow-up to #664: artifacts persisted on a real remote link. Root-caused live on R7-27 with the
viewer connected over Tailscale (RTT 226–445 ms). Three independent bugs:

## 1. RTP packets didn't fit the path (the dominant cause)

The media path is a WireGuard tunnel: **MTU 1280**, so 1252 usable on the wire. `rtpvp8pay`'s
default `mtu=1400` (+1 RED, +10 SRTP) fragmented **every full-size packet** — and either lost
fragment kills the whole RTP packet. Measured on the live path:

| ICMP payload | loss |
|---|---|
| 56 B | 0.83% |
| 1200 B | 1.67% |
| 1400 B (fragmented) | **2.5%** |

FEC packets are ~14 B *bigger* than the media packets they protect, so the repair traffic
fragmented and died at the same rate. → `mtu=1200`.

## 2. The sender adaptation was loss-blind and flapped on RTT

`on_peer_stats` read `rb-fractionlost` (uint) — that field only exists on webrtcbin's *internal*
input stats, never in the `get-stats` reply (verified by disassembling the shipping
`libgstwebrtc.so`: the reply carries **`fraction-lost` as a double in [0,1)**). So
`rtcp_loss_promille_` never left −1, "viewer loss 0.0%" in the logs was the sentinel, and the
DEGRADED trigger degenerated to `rtt >= 250` — which a tunneled path straddles forever.
Observed: **17 GOOD↔DEGRADED flaps in 8 minutes**, each entry = bitrate cut + 100% FEC +
1 Hz forced keyframes on a link that had no loss spike.

Now: the real loss field is read, loss is the *only* trigger (RTT is a property of the path, not
damage), and an absent report logs `loss n/a` instead of a reassuring `0.0%`.

## 3. Encoder tuning made every recovery keyframe hurt

- `error-resilient=partitions` was a no-op (only splits token partitions; `token-parts` is 1).
  `default` is the actual libvpx loss-resilience mode.
- IDR size was unbounded: a recovery keyframe ate the 600 ms CBR buffer and libvpx answered with
  a quantizer slam on the following frames — the blocky pulsing after every PLI/DEGRADED entry.
  `max-intra-bitrate=300` caps it at 3× the per-frame budget (same neighborhood libwebrtc uses).
- `main_fps` defaulted to 30 but the driver captures 15 (`stereo_depth_estimator.yaml`), so the
  keyframe backstop was silently 8 s and rate control got a doubled per-frame budget.

## Also: duplicate payload types across the BUNDLE

The live offer carried `pt 97 = red(video0)` **and** `VP8(video1)`, `pt 98 = ulpfec(video0)`
**and** `opus` — webrtcbin assigns each m-line's red/ulpfec/rtx PTs sequentially after that
m-line's media PT, and 96/97/98 left no room. RFC 8843 §9.1 forbids duplicates in a bundle;
Chrome tolerates it today via the signaled `a=ssrc` lines, newer Chrome rejects the offer.
Cameras now stride 5 PTs apart (96, 101, …) reserving each m-line's aux block; audio moves
to 127. Updated the two tests that pinned the old scheme.

## Verified

- `colcon build --packages-select mars_cam` clean; `test_webrtc_internal` 7/7 pass.
- New pipeline strings parse and apply on **both** stacks: system GStreamer 1.20.3 and
  `/opt/gst` 1.24.12 (`error-resilient=default`, `max-intra-bitrate=300`, `mtu=1200`, `pt=127`).
- Not yet soak-tested live — needs a container restart on a robot with a remote viewer.

## Test plan

1. `innate restart`, connect the webapp over the remote path, press `p`.
2. Expect: loss well under 1% (no fragmentation), no 30 s GOOD↔DEGRADED cycle in the
   `webrtc_streamer` logs, `viewer loss` showing real numbers (or `n/a`, never a fake 0.0%),
   and no blocky pulse after keyframes.
3. Sanity-check a LAN viewer too: nothing here should change the clean-path behavior beyond
   ~17% more (smaller) packets from mtu=1200.

---

_Supersedes #707, which was opened from the wrong account. Both greptile findings there (loss-blind DEGRADED exit on RTT-only reports; unbounded camera PT assignment) are addressed in ea1d17db._


karmanyaahm:
One more from live testing: the boot splash showed a dead, fully-filled bar for the whole load on Theo's machine — macOS Reduce Motion swaps the sweep for a full-width span whose opacity pulse was itself killed by the global `* { animation: none !important }` rule. Fixed: centered 40% pulsing segment under reduced motion (opacity-only re-assert over the star kill), plus a 0.4 s appear-delay so fast LAN mounts never flash a splash at all. Verified live in-browser under prefers-reduced-motion.

karmanyaahm:
Added the follow-up we discussed after re-measuring: `0a2b5ca8` makes DEGRADED genuinely shed load. The old recipe (60% bitrate × 100% FEC + 1 Hz forced keyframes) offered 1800 kbps against GOOD's 1875 — a 4% cut with double the packets and an IDR burst every second — and the one-step exit re-congested the path immediately (34 GOOD↔DEGRADED flips/hour measured live with honest loss numbers, 3–14% viewer loss with two tabs on one tunnel).

Now a three-rung ladder: **DEGRADED** = 40% bitrate + 2×-base FEC (~900 kbps/peer, half of GOOD), **RECOVERING** = 70% + base FEC, climb one rung per 15-clean-second window, drop straight to the bottom on 2 bad ticks. Forced keyframes removed from every rung — recovery stays PLI-driven (500 ms throttle) with the 4 s backstop.

Deployed on mars-the-27th for live testing.

DavidDobas:
Review, by priority: (1) correct technical solution, (2) clean readable DRY code, (3) realistic edge cases, (4) other bugs. Findings only — no fixes pushed.

## 1) Correct technical solution chosen

**Yes — all four diagnoses check out.**

- **`mtu=1200`**: correct and well-evidenced (1200 + 1 RED + 10 SRTP = 1211, FEC ~1225, both under WireGuard's 1252 usable).
- **`fraction-lost` as double**: correct. GStreamer's stats code converts the RR's `rb-fractionlost` (0–255 uint, internal source stats only) into `fraction-lost = x/256.0` on the `remote-inbound-rtp` reply — the old `gst_structure_get_uint(s, "rb-fractionlost", …)` could never match, exactly as diagnosed. `fl*1000` → promille against the 30‰ threshold is right.
- **Loss-only trigger + 3-rung ladder** with one-rung climbs and real shedding (40%, vs the old 60%×2×FEC that shed ~nothing net): sound control design.
- **PT stride**: correct per RFC 8843; `kMaxCameras = 6` is enforced in `configure_cameras` and the tightness is tested. Note pt 98→127 for opus only matters to a client that hardcodes PTs instead of reading SDP — webapp and test consumer are updated; worth a quick check that the mobile app doesn't pin 98 anywhere.

Two deliberate trade-offs to confirm as intended: (a) dropping RTT entirely means a bufferbloat path that *queues* instead of dropping won't degrade until queues overflow into loss; (b) removing the 1 Hz DEGRADED keyframe makes recovery fully PLI-dependent (browsers retry PLIs, so corruption age stays bounded in practice). Both defensible, both one-way doors.

## 2) Clean readable DRY code

Good — the ladder is readable and the atomics story is coherent (`adapt_rung_` executor-thread-only, `degraded_` atomic for the webrtcbin threads). Two things:

- `degraded_fec_pct()` is a member function returning a literal `100u`, referenced from three files as if it computed something — a named constant says the same with less machinery.
- **Scope**: the `boot-splash` CSS rework and the reduced-motion changes have nothing to do with artifacting on tunneled links — they belong in their own PR, especially since one is a regression (§4).

## 3) Realistic edge cases handled

Mostly yes — the webapp jitter-buffer chain is careful (superseded-pc guards correct, carried-over target on rebuild is a nice touch, audio receivers explicitly excluded). Three gaps:

- **The arm camera has the same fps invariant broken in the opposite direction.** The fix sets `main_fps: 15` to match the driver — but `kDefaults` declares arm at 15 while `arm_camera_driver` captures at **30** (`stereo_depth_estimator.yaml`). By this PR's own reasoning that halves arm's per-frame rate-control budget and makes its keyframe backstop 2 s instead of 4 s. Fix arm too, or say why it's exempt.
- **A peer joining mid-RECOVERING gets base FEC, not 2×**: `configure_video_transceivers` and the `on-new-transceiver` path key off the boolean `degraded_` only, so a RECOVERING-rung arrival sits below the fleet's FEC until the next rung transition (up to 15 s, or indefinitely on a stable rung).
- The get-stats reply repeats the *last received RR's* `fraction-lost` until the next RR arrives, so "2 consecutive bad ticks" can be one RR sampled twice — DEGRADED entry is effectively 1 report on sparse-RTCP paths. Pre-existing pattern (RTT behaved the same); just know the 2-tick debounce is weaker than it looks.

## 4) Other bugs

- **`prefers-reduced-motion` regression (webapp/css/app.css).** The PR deletes the old accommodation — which replaced the *translating* boot sweep with a static opacity pulse — and force-re-enables the full `boot-sweep` movement animation with `!important` under reduce-motion. Keeping an indicator alive is right ("frozen bar reads as hung"); keeping the *moving* one is not — suppressing movement is precisely what the preference exists for, and the deleted block solved it correctly. The opacity-only exceptions (`breathe`, `boot-appear`) are fine; the sweep should go back to a pulse.
- **The jitter poll chain can die permanently on a single throw.** `#adaptJitterBuffer` is well guarded except `receiver.track.kind` — `receiver.track` can be null (post-stop receivers), which would reject `tick`, and the setTimeout chain (unlike the old `setInterval`) never re-arms after an unhandled rejection. One `?.` or a `.catch` on the tick body restores the old resilience.

**Bottom line:** the transport-layer work is excellent and root-caused with unusual rigor — mergeable on sight once the arm-fps question is answered. The CSS hunk deserves its own PR, with the reduce-motion sweep fixed before it lands anywhere.

theo-michel:
Fixed the jitter poll chain problem

DavidDobas (approved):

theo-michel:
Tested seemed to work

PR #715, opened 2026-08-27
feat(stt): Scribe realtime by default — the transcript lands ~0.75 s after the last word

theo-michel:
## What

Replaces the batch-by-default STT pipeline (~1.1 s from last word to transcript) with **ElevenLabs Scribe v2 realtime over a warm socket, committed by our own silero endpointer** (~0.75 s: 0.5 s VAD silence window + 250–380 ms commit→transcript), at equal accuracy on the robot's far-field mic. Chosen over Cartesia ink-2, Gemini 3.5 Transcribe Live, and four OpenAI variants in a 25-run benchmark on a 50-clip labeled corpus from this mic plus a live room rig (per-run data: `data/speech_debug/results/` on R7-27).

## How

- **Manual commit beats vendor VAD on this mic, measured**: ElevenLabs' server VAD delivers 0.73–0.9 s at *any* silence setting (~0.5 s decision overhead) and loses 11 WER points on noisy audio (it trims real words). Every backend now endpoints locally with the same silero detector.
- **Slow AGC on the capture path** (instant attack, ~16 s release, boost-only, +24 dB cap): the mic runs ~−36 dB RMS and normalizing it alone was worth 3–11 WER points across *every* engine tested. Ducked audio bypasses the AGC so the robot's own voice can't slam the gain. 0/600 silero false-triggers on +18 dB-boosted room tone.
- **Warm-socket discipline**: silence frames stream during ducking (Scribe needs ~2 s of received audio before it processes; a paused stream goes cold), and a safety commit fires after 30 s of uncommitted audio so the vendor's ~36 s force-commit can never cut mid-word.
- **Batch is the automatic fallback**: after 3 connects without a `session_started`, the reconnect loop settles for `elevenlabs_batch` so the robot keeps hearing; the next voice session retries realtime. The fallback got fast too: raw 16 kHz PCM upload + no timestamps/audio-tags = **0.38–0.45 s** warm (was ~1.1 s).
- **The best part is no part**: the OpenAI backend is deleted — it spoke the retired beta realtime protocol with a superseded model and could no longer connect. Its params (`openai_realtime_*`, `stt_realtime_vad_*`) go with it.
- Keyterms + a `previous_text` vocabulary hint bias every ElevenLabs path. Note: the Innate proxy relay currently collapses repeated query params, so only one keyterm survives the trip — the client detects this from the session echo and warns; `previous_text` (body content) carries the vocabulary meanwhile. Fix is cloud-side and in progress.

## Verified

- `ruff` clean; **274/274** brain_client tests pass; `basedpyright` adds no new errors in touched files (14 pre-existing errors in untouched files).
- Live on the robot stack (rebuild → restart → sentences played through a second robot's speaker across the room): `🎤 Scribe committed in 310 ms: 'Okay, so, like, can you just try it one more time?'` — verbatim, fillers intact, 262–376 ms across the run.
- Fallback drill: bad model → 3 counted session failures → batch takeover → correct transcript at 379 ms.

## Rollback

`stt_backend: "elevenlabs_batch"` in settings (or the launch arg) — one line, no code.

## Follow-ups (not in this PR)

- Innate proxy: forward repeated query params (keyterms) — in progress cloud-side; the robot's warning disappears on its own when it lands.
- VAD window tuning: at 0.5 s, slow-paced speech already splits at pauses — the per-utterance `committed in` logs make split-rate measurable during daily driving before trialing 0.4 s or 0.6 s.
- Reconnect stress (150-cycle) and a day's cost measurement of continuous streaming during active voice sessions.


axelpey:
Closing because this work is bundled into #717, which is the canonical PR for the voice-latency push.

PR #717, opened 2026-08-27
feat(voice): answer in one turn, ~3.2 s from last word to first word — bundles #715 + #716

theo-michel:
## What

One PR for the whole voice-latency push: merges **#715** (Scribe realtime STT) and **#716** (chunked history compaction for Gemini's implicit cache — still gemini-3.6-flash, still images) and adds the two remaining fixes from the latency investigation. Measured live on mars-the-47th, median end-of-speech → first audio of the answer:

| Stack | Total (n=6) |
|---|---|
| main, before any of this | **5.75 s** (5.56–6.58) |
| + same-turn prompt fix + fast TTS | **4.08 s** (3.52–6.21) |
| + #715 realtime STT + #716 compaction (this PR) | **3.18 s** (2.93–3.48) |

Stage medians on this branch: endpointer silence 500 ms → Scribe commit→transcript **274 ms** → brain wake 20 ms → **one** Gemini turn (speech + head nod together, 6/6 runs) 1973 ms → TTS first byte 284 ms.

## The two new commits

**fix(brain): answer in the turn you are asked** — the Intro Agent's "always use a head emotion" directive made the model substitute the nod *for* the answer: 6/6 live runs spent a full turn on `head_emotion` with no speech and answered a turn later (~2 s wasted per question). A same-turn rule appended **after** the directive block (placement matters: 16/16 there vs 5/8 as a Rules bullet, p=0.0002, n=16 vs n=8) makes speech + tool call one response, answer first, working call before the emotion so a forced nod can't steal the skill slot from a real action. Side-effect checks: pure-action commands keep correct call order and args; idle turns stay silent; stray STT fragments still ignored.

**feat(tts): raw 16 kHz fast path with speed 1.5 on the speaker** — headerless PCM into `aplay` with explicit format args + Cartesia `generation_config.speed=1.5`. TTFB unchanged within noise (277 vs 291 ms, n=10) but the stream is 2.7× smaller and clips ~10% shorter, so the robot finishes talking and un-ducks the mic sooner. Sim path keeps WAV (browser decoders need the container). `/tts/sse` was measured and rejected — the proxy buffers it whole (729 ms TTFB). Sonic 3.6 was measured and rejected — +63 ms TTFB vs 3.5, interleaved n=8.

## Deliberately not in this PR

Speculative Gemini send on the final STT partial (measured worth ~0.5 s, needs cancel-on-resume complexity), `history_max_image_turns: 0` (−0.26 s but loses vision memory; #716 makes the frames cheap to keep), model swap to 3.5-flash-lite, prompt caching beyond #716 (measured: explicit caches buy ~36 ms for a 1.6 s per-turn rebuild; the Interactions API is +0.6 s slower — TTFT has a 707 ms floor regardless of prefill).

## Verified

- `ruff` clean, **274/274** brain_client tests pass, `basedpyright` adds no new errors (14 pre-existing in untouched files).
- Live on the robot after build+restart: the 6-run session above; the robot says the answer *while* nodding on every question.
- Merge note: local `gh` predates stacked PRs, so this bundles the two branches by merge — if #715/#716 land first, this diff shrinks to the two commits automatically.

Latency report with all measurements: internal artifact (ask Theo for the share link).


theo-michel:
### Added: the agent chat surviving a dropped websocket (b1110efe)

Chased down a report of the chat freezing mid-conversation — the robot kept working, the webapp just stopped showing it, and only a page reload brought it back. **Neither cause is from this branch**; both are in `main` and predate the branch point (`91fca861`). Folding the fixes in here since they were found while testing it.

#### 1. The websocket was being killed by its own heartbeat

`webapp/proxy/https_server.py` relays `/ws` → rws with `heartbeat=20.0` on both legs. aiohttp derives the pong deadline as `heartbeat / 2` ([`web_ws.py:129`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_ws.py)), so a PONG arriving later than **10 s** hard-closes the connection with `No PONG received`. On WiFi already carrying two video tracks and an audio track, that fires on perfectly healthy sockets.

Measured client lifetimes across three boots, nothing done to provoke them:

| boot | client | lifetime |
|---|---|---|
| `81584` | 0 | 136 s |
| `81584` | 1 | 156 s |
| `93318` | 0 | 147 s |

Each ends with rws logging `Failed to send ping: Bad Connection` — the downstream echo, logged when it next touches a socket the relay already tore down. Browser console shows the matching `WebSocket connection to 'ws://mars-the-2nd.local/ws' failed` followed by the WebRTC session re-handshaking on the restored socket.

`WS_HEARTBEAT = 60.0` gives 30 s of slack. Dead-peer reaping goes from ~20 s to ~60 s, still far inside the kernel TCP timeout the comment there is guarding against.

Introduced 2026-07-22 in `2e188bb3`.

#### 2. …and the transcript never came back

`rosClient` reconnects fine — backoff caps at 10 s, subscriptions are replayed with `retryCount` reset, handler exceptions are isolated per handler. That's why video recovered on its own.

What didn't recover was the chat, because `loadHistory()` was latched:

```js
let historyLoaded = false;
async function loadHistory() {
  if (historyLoaded) return;   // fires on every reconnect, runs exactly once
```

`replay()` has one caller and it's behind that flag. So on reconnect the live stream resumed but everything published during the gap was gone from the DOM permanently. A reload was the only thing that cleared it — exactly the reported symptom.

Now refetches on every connect; the flag only prevents two fetches overlapping. This is lossless: the brain's snapshot already carries chat, thoughts, system messages **and** skill runs (`brain_client_node.py:385` and `:520` both append into `chat.history`), so replaying a fresh snapshot can't drop anything the live stream had rendered.

Introduced 2026-08-21 in `58ceb911` (#687). Worth noting the timing — before #687 there was no backfill at all, so a gap just looked like a gap. #687 added the backfill and latched it, which is what turned an intermittent drop into a chat that appears frozen.

#### One thing this branch does aggravate

Not a cause, but related: `history_max_entries` 60 → 2000 roughly doubles `/brain/trace` payloads. Measured live at **627 KB mean / 748 KB max** per `turn_request`, ~230 KB/s sustained — that's the full Gemini request body on the wire every turn. It adds LAN pressure that makes a tight pong deadline easier to miss, though drops happen with zero trace subscribers too.

Two things make it worse than it needs to be, both out of scope here:
- `brain/main.js` `setVisible(false)` pauses the animation loop and the camera fallback but **never drops the `/brain/trace` subscription** — open Inspect Brain once and the tab parses ~1 MB of JSON every few seconds for the rest of the session.
- `ChatManager.history` is unbounded and `get_chat_history` serialises all of it.

Happy to split those into a follow-up.

#### Verification

- `ruff check` + `ruff format --check` clean on `https_server.py`
- `tsc --noEmit`: 29 pre-existing errors in `js/` before and after, none in `agentPanel.js`

DavidDobas:
Ran the merged branch's pruning path with realistic turn inputs to sanity-check the 2000 cap. The image side holds up exactly as described — masking is O(1) in the cap, so going 400 → 2000 adds text entries and zero image bytes (at 2000 entries only 4 still carry JPEG bytes). Three things on the text side.

### 1. Entries measure ~40 tokens, not ~17

Steady-state supervision turn (skill running, robot silent), built from the real `observation_text`:

```
 26 tok  status line (clock, pose, running skill)
 27 tok  guidance line
 10 tok  "(second image is the arm wrist camera)"
~15 tok  two frame placeholders, once masked
  0 tok  model entry: ''
-------
~79 tok / turn  =  ~40 tok/entry
```

History oscillates between 1000 and 2000 entries (compaction evicts to half the cap), so that's **~40k–80k prompt tokens**, not ~34k. Still under the ~100k line, but with ~25% headroom rather than 3×.

### 2. The guidance line is ~1/3 of the text budget, and it's pure duplication

`guidelines_when_running` is constant for the life of a skill, but it gets stored into every turn's entry — 27 of those 79 tokens, ~54k tokens at full depth. Not a cache problem (appending doesn't break the prefix), but it's carried weight forever.

Hoisting it into `systemInstruction`, where it changes on skill start/stop instead of every turn, roughly halves the text budget and would make 2000 comfortable rather than marginal. Cheaper win than tuning the cap.

### 3. Half the entries are empty strings

On a silent supervision turn, `absorb` strips the thought parts and stores `{"role": "model", "parts": [{"text": ""}]}`. Zero content, but it still consumes a slot against the cap — so 2000 entries is ~1000 turns of real memory, ~667 once tool calls make it 3 entries/turn. Worth saying out loud in the config comment, since "2000 entries" reads as more depth than it buys.

<details>
<summary>What the wire actually looks like after 25 turns</summary>

```
[ 0] user   '[Fri 14:03 UTC | t+900s] pose: x=3.21m y=-1.04m heading=109°
              - user said: "pick up the socks and put them in the box"'
             '[older camera frame removed]'
[ 1] model  'On it — picking up the socks now.'
             functionCall run_skill({"skill":"pick_any_object",...}) +thoughtSignature
[ 2] user    functionResponse -> {'outcome': 'started'}
[ 3] user   '[… t+905s] pose: … | running skill: pick_any_object
              (guidance while this skill runs: Report what you see in the gripper. …)
              (second image is the arm wrist camera)'
             '[older camera frame removed]'
             '[older wrist camera frame removed]'
[ 4] model  ''
[ 5] user   '[… t+910s] … (same three lines, verbatim) …'
             '[older camera frame removed]'
             '[older wrist camera frame removed]'
[ 6] model  ''
             … 20 more identical pairs …
[49] user   '[… t+1020s] …'   <IMAGE>  <IMAGE>   <- only the newest 3-4 hold real bytes
[50] model  ''
```

</details>

One thing I couldn't measure without live traffic: `absorb` keeps `thoughtSignature` verbatim on model parts, and those are unbounded from our side — they'd land on every action turn. `last_usage["prompt"]` (added in this PR) is exactly the instrument for it: ~20 min of a real manipulation session on R7-27 reading `promptTokenCount` settles both that and the 40 tok/entry number above.

Numbers are simulated against the real `absorb`/`_prune` code path, not measured on a robot — so treat them as sizing, not ground truth.

DavidDobas:
### Added: the agent chat surviving a dropped websocket (a7b44755)

Chased down a report of the chat freezing mid-conversation — the robot kept working, the webapp just stopped showing it, and only a page reload brought it back. **Neither cause is from this branch**; both are in `main` and predate the branch point (`91fca861`). Folding the fixes in here since they were found while testing it.

#### 1. The websocket was being killed by its own heartbeat

`webapp/proxy/https_server.py` relays `/ws` → rws with `heartbeat=20.0` on both legs. aiohttp derives the pong deadline as `heartbeat / 2` ([`web_ws.py:129`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_ws.py)), so a PONG arriving later than **10 s** hard-closes the connection with `No PONG received`. On WiFi already carrying two video tracks and an audio track, that fires on perfectly healthy sockets.

Measured client lifetimes across three boots, nothing done to provoke them:

| boot | client | lifetime |
|---|---|---|
| `81584` | 0 | 136 s |
| `81584` | 1 | 156 s |
| `93318` | 0 | 147 s |

Each ends with rws logging `Failed to send ping: Bad Connection` — the downstream echo, logged when it next touches a socket the relay already tore down. Browser console shows the matching `WebSocket connection to 'ws://mars-the-2nd.local/ws' failed` followed by the WebRTC session re-handshaking on the restored socket.

`WS_HEARTBEAT = 60.0` gives 30 s of slack. Dead-peer reaping goes from ~20 s to ~60 s, still far inside the kernel TCP timeout the comment there is guarding against.

Introduced 2026-07-22 in `2e188bb3`.

#### 2. …and the transcript never came back

`rosClient` reconnects fine — backoff caps at 10 s, subscriptions are replayed with `retryCount` reset, handler exceptions are isolated per handler. That's why video recovered on its own.

What didn't recover was the chat, because `loadHistory()` was latched:

```js
let historyLoaded = false;
async function loadHistory() {
  if (historyLoaded) return;   // fires on every reconnect, runs exactly once
```

`replay()` has one caller and it's behind that flag. So on reconnect the live stream resumed but everything published during the gap was gone from the DOM permanently. A reload was the only thing that cleared it — exactly the reported symptom.

Now refetches on every connect; the flag only prevents two fetches overlapping. This is lossless: the brain's snapshot already carries chat, thoughts, system messages **and** skill runs (`brain_client_node.py:385` and `:520` both append into `chat.history`), so replaying a fresh snapshot can't drop anything the live stream had rendered.

Introduced 2026-08-21 in `58ceb911` (#687). Worth noting the timing — before #687 there was no backfill at all, so a gap just looked like a gap. #687 added the backfill and latched it, which is what turned an intermittent drop into a chat that appears frozen.

#### One thing this branch does aggravate

Not a cause, but related: `history_max_entries` 60 → 2000 roughly doubles `/brain/trace` payloads. Measured live at **627 KB mean / 748 KB max** per `turn_request`, ~230 KB/s sustained — that's the full Gemini request body on the wire every turn. It adds LAN pressure that makes a tight pong deadline easier to miss, though drops happen with zero trace subscribers too.

Two things make it worse than it needs to be, both out of scope here:
- `brain/main.js` `setVisible(false)` pauses the animation loop and the camera fallback but **never drops the `/brain/trace` subscription** — open Inspect Brain once and the tab parses ~1 MB of JSON every few seconds for the rest of the session.
- `ChatManager.history` is unbounded and `get_chat_history` serialises all of it.

Happy to split those into a follow-up.

#### Verification

- `ruff check` + `ruff format --check` clean on `https_server.py`
- `tsc --noEmit`: 29 pre-existing errors in `js/` before and after, none in `agentPanel.js`

DavidDobas (approved):
Approving because it works well, but we should think about the lot of empty messages in history

theo-michel:
Fixed the issues showed by David

PR #718, opened 2026-08-27
optimizing Process ram

theo-michel:
**Draft — analysis tooling only. No consolidation changes in here yet.**

We're past 20 ROS nodes and every process costs RAM whether or not it's doing
anything. Before changing any of that, I wanted to see the actual shape of the
fleet. Nothing in ROS reports it: `ros2 node list` knows nodes but not
processes, `ps` knows processes but not nodes.

This adds `/fleet.json`, which joins the two — it walks each tmux pane's
process tree, reads `/proc/<pid>/smaps_rollup`, and correlates against
`ros2 node list` + `ros2 component list` — and `/fleet.html` to render it.

### What it shows

Three groupings: **by tmux window** (the topology `innate view` shows you), **by
memory**, and **by nodes per process**. The last one is the point.

Measured on mars-the-47th, 6 processes carry 34 of the 67 nodes:

| nodes | PSS | process |
|---|---|---|
| 16 | 196 MB | `camera_container` |
| 5 | 521 MB | `brain_client_node` |
| 4 | 182 MB | `skills_action_server` |
| 4 | 49 MB | `bt_navigator` |
| 3 | 80 MB | `mode_manager` |
| 2 | 49 MB | `arm_sdk_server` |

Against **33 processes hosting one node each, costing 2,258 MB**, plus 16
`ros2 launch` supervisors at 431 MB that are idle from boot onward.

The camera container is the argument in one line: 16 nodes for 196 MB, while
the nav pane spends more than twice that on 15 nodes in 15 processes.

### Why PSS

RSS charges every shared page in full to every process mapping it, so summing
it across 40 ROS processes counts libc, rclcpp and Zenoh dozens of times over.
PSS divides shared pages by the number of sharers, so these numbers sum
honestly.

Worth knowing when reading them: PSS also means per-process savings don't sum
into system-wide savings. Delete a process and some of "its" memory is
re-attributed to the survivors rather than freed.

### Attribution

Exact where it can be — `__node:=` plus `__ns:=`, container components, and
duplicate names in `ros2 node list` (several nodes here spin in-process helpers
that reuse the node name, which is why the launch files mute the resulting
"Publisher already registered" warning).

A node started by `ros2 run` has no `__node:=` to read, so its ROS name is
inferred from the executable path by token overlap and flagged as inferred in
the detail panel.

The ~12 helper nodes that genuinely can't be pinned to a PID by name — TF
listeners, nav2 costmaps — are listed separately in the footer rather than
guessed at, so the totals never quietly disagree with `ros2 node list`.

### Notes

- A scan costs ~3.6 s of `ros2` CLI, hence the 4 s server-side cache — a mashed
  Refresh button or several open tabs must not queue up one scan each.
  Auto-refresh is 10 s.
- Read-only. Touches no ROS node; the only change outside the two new files is
  two lines in `https_server.py` registering the route.
- Uses the webapp's existing design tokens rather than a new palette.

### Try it

```
https://mars-the-47th.local/fleet.html
```

### Verified

`/`, `/config.json`, `/settings.json` and the rest still return 200; page and
endpoint serve over both 4080 and 4443; JSON contract has every field the page
reads; ruff clean. Not render-tested — there's no browser on the Jetson.

### Still to come

The consolidation work itself. Findings so far, in rough order of payoff:
`torch` resident from boot in `manipulation_server` (353 MB); the nav2 stack as
nine separate processes when every one of those nodes is already a registered
component; `cupy` permanently resident in `grid_localizer` (186 MB); and an
idle `stereo_calibration_manager` holding two raw stereo subscriptions it
discards — ~28 MB/s serialized out of the camera container for nothing.

One caveat found the hard way and worth recording before anyone composes
anything: `ComposableNodeContainer` makes `launch_ros` stand up a full ROS node
*inside the launch supervisor* to call the load service, and it never exits
(20 threads vs 2). On a two-node container that tax cancelled the saving
exactly. Compose in bulk or not at all.

### DIFF
--- a/webapp/fleet.html
+++ b/webapp/fleet.html
@@ -1,0 +1,575 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>Node Fleet · Innate OS</title>
+<link rel="icon" href="/public/innate-logo.avif">
+<style>
+:root {
+  --bg: #0a0a0c;
+  --panel: #111114;
+  --panel-2: #17171b;
+  --hairline: rgb(255 255 255 / 8%);
+  --hairline-strong: rgb(255 255 255 / 16%);
+  --text: #e7e7ea;
+  --muted: #8a8a93;
+  --faint: #5c5c66;
+  --accent: #e8a33d;
+  --accent-faint: rgb(232 163 61 / 12%);
+  --ok: #56c28c;
+  --danger: #d96a5a;
+  --font-ui: system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
+  --font-mono: ui-monospace, "SF Mono", "Cascadia Code", Menlo, Consolas, monospace;
+
+  --w0: #e8a33d; --w1: #5fb8a3; --w2: #6b93d6; --w3: #a389d4; --w4: #56c28c;
+  --w5: #d96a5a; --w6: #b58ce6; --w7: #7fc4e0; --w8: #d4b45a; --w9: #8f9bb3;
+}
+
+* { box-sizing: border-box; }
+
+html, body { height: 100%; }
+
+body {
+  margin: 0;
+  background: var(--bg);
+  color: var(--text);
+  font-family: var(--font-ui);
+  font-size: 14px;
+  line-height: 1.5;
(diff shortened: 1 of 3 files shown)


theo-michel:
Closing as stale — the node fleet viewer never went past draft. The branch `theo/optimize-process-ram` keeps the work if it is wanted later.

PR #719, opened 2026-08-27
fix(skills): pickAnything keeps picking the object it first saw (object permanence)

theo-michel:
## Problem

With several identical objects in view (two white socks), every Gemini look in `pick_any_object` took the **"best first"** match. Gemini's ordering is not stable between calls, so the skill could silently switch instances between the search, the re-localizations during positioning, and the wrist seed — and grab the wrong twin.

## Mechanism

A minimal single-target tracker on top of the existing pipeline:

- **Parse all matches** — `innate.vision` gains `parse_det_cands` / `parse_det_boxes` (the singular parsers remain as thin wrappers), and both Gemini prompts ask for ALL matches.
- **Odom-anchored memory** — the first detection anchors `_last_seen_odom`; every later look back-projects each candidate to the floor and keeps the one nearest the memory. Odom is the right frame: continuous (no relocalization jumps mid-run), cm-scale drift over a pick run, and every Gemini look happens with the base stopped, so the odom read matches the frame's capture pose. Without odometry it degrades exactly to today's behavior.
- **Validation gate (`mem_gate_m` = 0.5 m)** — objects don't teleport: a lone match farther than the gate from the last sighting is a *different* instance, so the skill coasts ("not seen this frame") instead of re-anchoring on a twin. This closes the hijack case where the tracked sock is missed for one frame (arm occlusion, frame edge) while the twin is detected. Sized above worst-case back-projection disagreement between consecutive looks (~0.3 m/deg of tilt error at 2 m, largely systematic, and the memory refreshes every localize).
- **Wrist stage** — seeds from the box nearest the servo aim point rather than Gemini's first. Grounded because `_position_above` only returns once the target is proven inside the accept box and the arm is aimed at its bearing.
- **Verify stage** — grasp verification now judges only the grab spot directly ahead. Previously, after a *successful* pick, the twin still on the floor made Gemini truthfully answer "YES, the sock is on the floor" → the pick was reported as a miss → teardown opened the gripper and dropped the held sock.
- **Observability** — the skill logs `N matches; kept #k (d m from last sighting)`; any run keeping a non-first candidate is direct evidence the old code would have switched socks.

## Prerequisite framework fix (first commit)

Annotating a skill attribute `tuple[float, float] | None = None` crashed skill loading: py3.10's `isinstance(tuple[float, float], type)` quirk let a generic alias reach `issubclass()` in the feed materializer. It now filters with `get_origin` — a subscripted generic is never a feed. Any skill author could have hit this.

## Verification

- ruff check + format clean on all touched files
- Full brain_client suite: **274 passed** (fullbuild image, overlay recipe)
- basedpyright: identical error counts to origin/main under both the brain_client scope and the repo pyrightconfig scope (all pre-existing env artifacts) — zero new errors
- Skill loads with zero `_declaration_issues` and exactly the four real feeds
- Unit-level checks (run in-container, then deleted per repo convention): both matches parsed; memory beats Gemini order; memory survives a 90° base rotation; lone far twin coasted with the anchor untouched; near lone match accepted; unprojectable candidates coast; no-memory falls back to best-first

End-to-end two-sock trials in the MuJoCo sim are running now; results will be posted on this PR.

## Hardware note

The approach assumes odom drift stays small relative to twin separation within one run (holds easily for the drives involved). A real two-socks-close-together run on a robot is the convincing final check.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/skills/types.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/skills/types.py
@@ -645,7 +645,10 @@ def _materialize_feed_annotations(cls) -> None:
             )
             continue
         resolved, optional = _split_optional(annotation)
-        if isinstance(resolved, type) and issubclass(resolved, Skill) and resolved is not Skill:
+        # py3.10 quirk: a subscripted generic (`tuple[float, float]`) passes
+        # isinstance(..., type) but crashes issubclass — and is never a feed.
+        is_class = isinstance(resolved, type) and get_origin(resolved) is None
+        if is_class and issubclass(resolved, Skill) and resolved is not Skill:
             if optional:
                 # The feed rule (`| None` = best effort) does not extend to
                 # composition: a declared sub-skill is always wired and its
@@ -660,7 +663,7 @@ def _materialize_feed_annotations(cls) -> None:
             setattr(cls, name, descriptor)
             descriptor.__set_name__(cls, name)
             continue
-        if isinstance(resolved, type) and issubclass(resolved, TrainedSkill) and resolved is not TrainedSkill:
+        if is_class and issubclass(resolved, TrainedSkill) and resolved is not TrainedSkill:
             # a generated physical-skill ref: same declaration shape as a code
             # sub-skill, materialized as the PhysicalSkill descriptor
             if optional:
@@ -679,7 +682,7 @@ def _materialize_feed_annotations(cls) -> None:
                 "(`gripper_open: GripperOpen`); bare `Skill` declares nothing."
             )
             continue
-        entry = feed_types.get(resolved) if isinstance(resolved, type) else None
+        entry = feed_types.get(resolved) if is_class else None
         if entry is None:
             hint = _feed_attr_hints().get(name)
             if hint is not None:
--- a/ros2_ws/src/brain/brain_client/innate/vision.py
+++ b/ros2_ws/src/brain/brain_client/innate/vision.py
@@ -68,16 +68,29 @@ def _box_corners_px(det):
     return (x0 / 1000.0 * IMG_W, y0 / 1000.0 * IMG_H, x1 / 1000.0 * IMG_W, y1 / 1000.0 * IMG_H)
 
 
-def parse_det_px(text):
-    """Best (u, v) from a Gemini detection reply."""
+def parse_det_cands(text):
+    """All detections -> [(u, v, grip_strength | None)], best first.
+    (u, v) is the grasp_point when given, else the box center."""
+    cands = []
     for det in parse_dets(text):
         gp = det.get("grasp_point")
         if gp and len(gp) >= 2:
-            return (_norm1k(gp[1]) / 1000.0 * IMG_W, _norm1k(gp[0]) / 1000.0 * IMG_H)
-        c = _box_corners_px(det)
-        if c:
-            return ((c[0] + c[2]) / 2.0, (c[1] + c[3]) / 2.0)
-    return None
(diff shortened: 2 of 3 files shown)


theo-michel:
## Sim validation: two-sock trials in the MuJoCo twin — PASS

Ran 14 trials (12 valid) with two **identical** sock props in the sim apartment, ground truth streamed from the world server, all but the first two on the current branch tip (`29ac913`). Full logs + 75 Hz ground-truth JSONL were captured per trial.

**Verdict: object permanence holds.** Zero mid-run instance switches in 12/12 valid trials — every accepted localization mapped to the same sock as the first, and every completed pick lifted the sock the robot first saw.

The counterfactual fired three times — Gemini ranked the **twin** first and the memory overrode it:

```
[PickAnyObject] px=(292,186) -> base_link (1.032,0.093)
[PickAnyObject] 2 matches; kept #2 (0.09m from last sighting)
[PickAnyObject] px=(365,351) -> base_link (0.365,-0.012)
[PickAnyObject] wrist stage: reached stop z (z=0.050)
[PickAnyObject] verify: floor='NO' j6=0.482 (2 cams) -> HELD
```

(T7a, verbatim; socks 0.22 m apart.) Ground truth: the tracked sock rose z 0.03→0.098 and traveled with the robot; the twin never moved. On the old `cands[0]` code these three runs would have switched to the wrong sock.

**Verification with the twin in view: PASS.** Four HELD verdicts with the twin still on the floor (0.22–0.80 m away, in frame) — the old prompt would have read the twin as "the sock is still on the floor" and dropped the held one. And on the one genuinely empty grasp it correctly said NOT HELD (`floor='YES' j6=-0.334`), so the scoping didn't create false positives.

| trial | separation | prompt | picked (GT) | kept #k>1 | verify | outcome |
|---|---|---|---|---|---|---|
| T2b/T2c | 0.80 m symmetric | sock | first-seen ✅ | — | HELD | success ×2 |
| T6a | 0.36 m | block | first-seen ✅ | — | HELD | success |
| T7a | 0.22 m | block | first-seen ✅ | **#2 ×1** | HELD | success |
| T4a | 0.80 m mirrored | sock | (empty grasp) | — | NOT HELD ✅ | grasp missed |
| T1b/c, T3a/b, T5a, T7b | 0.40–0.80 m | sock | none | — | — | could not centre* |
| T7c | 0.22 m mirrored | block | none | #2 ×2 | — | could not centre* |

\* Every "could not centre" was **detection**, not selection: the sim sock prop is a 4×4×6 cm grey box, and Gemini often refuses to call it "the sock" on the close-in confirm look (half-occluded by the arm). With a prompt matching what it actually looks like ("the small grey block") the same placements succeeded 2/3. Selection stayed consistent in all of them, and the coasting/gate-rejection path never fired in any trial. Same-behavior-on-main, unrelated to this diff.

**Known limit, observed once (T7c):** at 0.22 m separation, one look returned *only* the twin, which sat inside the range-scaled gate — so it was accepted. Below the gate radius only nearest-neighbor applies, and nearest-neighbor can't help when the true target is absent from the candidate list. Working as designed; re-identification would be the next tier if sub-gate twins ever matter in practice.

**Still untested:** the coast/re-anchor paths (never triggered in sim), and detection quality on a real sock — recommend one hardware run with two socks ~0.5 m apart before calling it done.

Repro notes: needs a second sock prop (`sim/props/13_sock_twin.py`, identical body, distinct name — left uncommitted since adding a prop forces a full world-model recompile for everyone); place via the world-server socket (`{"op":"drop_prop_at",...}` on :8800) in the y≈−1.17 band (the runbook's −1.60 band is behind the sofa); trials also surfaced a sim bug where the TCP reset path wedges the arm trajectory servo — filing separately.

DavidDobas:
@greptileai

axelpey (approved):
Tested a couple times and works much better, thanks!

PR #723, opened 2026-08-28
Add searchable keyboard navigation to Teleop skills

axelpey:
## Summary

- add a focused search field to the Teleop skills dropdown
- open it with Cmd/Ctrl+K and show the shortcut on the button
- filter the live skill roster as the user types
- highlight the first result and support Up/Down plus Enter activation
- preserve the existing inline parameter form behavior for parameterized skills

## Validation

- one combined search + selection + ArrowDown + Enter-target regression test
- all existing `webapp/tests/*.test.js` tests
- `node --check webapp/js/teleop/skillsMenu.js`
- `git diff --check`
- live simulator smoke test for focus, filtering, highlighting, and Enter activation

## Follow-up

Keyboard traversal within a parameterized skill form is intentionally left for a follow-up change.


DavidDobas (approved):
Remove test if not needed

theo-michel:
@greptileai

PR #725, opened 2026-08-28
fix(sim): three small fixes to the sim stage's overlays

DavidDobas:
Three independent papercuts on the Agent page in simulation, one commit each.

### 1. Camera tiles are always live in sim (`6a3a321c`)

The strip's `off` rung is a bandwidth control from #664: on a real robot each live tile is a full-bitrate WebRTC stream, so a fresh profile opens with the primary camera only and the rest read **"tap to view"**.

In sim there is no stream to shed. A tile is a scissor render of the scene already on screen, round-robined one per every other frame in `simStage`, blitted into a canvas the tile mounts directly — two live thumbnails cost the same total GPU work as one, each refreshing at half the rate. The off rung was charging a click for nothing.

Sim cameras skip it: `reconcile()` enables the whole roster, and the tiles carry no close `×` since there is nothing to turn off. The map keeps its `×` — that widget is genuinely heavy (lazy import, ROS subscriptions, canvas).

**Real robots are untouched** — `simCams` is false and every path is byte-identical to #664 as merged: the ladder, the off tiles, the `×`, the v2 migration. Profiles that already persisted the sim default heal themselves on the first `/webrtc/active_streams` message, so nobody has to clear localStorage.

@theo-michel — flagging since this is your area: #664 restored the ladder globally and the sim carve-out went with it, but the PR only argues about WebRTC bitrate, so this reads as collateral rather than something you'd have wanted. Shout if I've misread it.

### 2. Scene setup and Challenges open exclusively (`a7965bb2`)

Their toggles sit side by side in the stage's bottom-left corner and both panels expand upward from there, up to 360px wide — so opening the second laid it over the first, headers and close buttons stacked:

Opening either now closes the other. They ship in separate bundles (the sim viewer is vite-built and imported at runtime by `robotSession.js`), so the handshake is a document event rather than a shared module; the event name and panel ids are a contract between the two files, stated in both.

Closing scene setup this way also drops any armed prop, which would otherwise keep following the cursor with its panel gone.

The challenge panel yields **even mid-run**: deliberately opening scene setup is a clear enough intent, and nothing is lost — the judge state is server-side, so reopening shows the live goals and timer. That is deliberately the opposite of the existing `dismiss()`, which refuses to close a running challenge, because there the trigger is a stray click on the scene rather than an intent.

### 3. Placement defaults to "choose spot" (`f250a30a`)

Near-robot drops everything at the same reach offset in front of the robot, so setting up a scene meant placing, driving, placing again. Choose-spot is the mode that actually builds a scene, and it is one click away from the same result — objects still land wherever you click, including in front of the robot. Both modes are unchanged; only which one the panel opens in.

## Verification

- `sim/viewer`: `npm run typecheck` clean, `npm run build:lib` clean
- `webapp`: `tsc --noEmit` — 0 errors in both touched files; repo total 5177 before and after, an unchanged baseline (almost entirely `public/vendor/three.module.min.r160.js`, which `tsconfig.json` does not exclude)
- `webapp`: `node --test tests/*.test.js` — 8/8 pass

Not exercised in a running sim — worth a click-through of the two panels before merge.

## Noticed while here, not fixed

`webapp` has no `tsc` in CI and no npm script for it, and `tsconfig.json` doesn't exclude `public/vendor/`, so the check reports 5177 errors and can't work as a gate. Adding `"exclude": ["public/vendor/**"]` would drop it to the ~30 real ones in `armsdk/viz.js`, `brain/main.js`, `settings/main.js`, `teleop/agentState.js`, `keyboardDrive.js`. Happy to do it as a follow-up.

### DIFF
--- a/sim/viewer/src/simStage.ts
+++ b/sim/viewer/src/simStage.ts
@@ -172,7 +172,7 @@ export function createSimStage(parent: HTMLElement, session: SimSession): { audi
   modeSwitch.append(robotButton, spotButton);
   const placementHint = document.createElement("p");
   placementHint.className = "sim-placement-hint";
-  let placement: PlacementState = { kind: "near-robot" };
+  let placement: PlacementState = { kind: "choose-prop" };
   const selectedProp = (): string | null =>
     placement.kind === "following" || placement.kind === "rotating" ? placement.prop : null;
   const refreshPlacementHint = () => {
--- a/sim/viewer/src/simStage.ts
+++ b/sim/viewer/src/simStage.ts
@@ -18,6 +18,14 @@ const THUMB_FRAME_DIV = 2;
 // visible gain (~75Hz interpolated state) and the load jitters everything.
 const MIN_FRAME_MS = 1000 / 62;
 
+// Scene setup and the webapp's challenge panel expand over the same corner of
+// the stage, so at most one may be open. They ship in separate bundles -- this
+// one is vite-built and imported at runtime -- so the handshake is a document
+// event rather than a shared module: the event name and the detail.panel values
+// are a contract with webapp/js/agent/challengePanel.js.
+const PANEL_OPEN_EVENT = "innate:panel-open";
+const PANEL_ID = "sim-scene-setup";
+
 const VIEW_FOR: Record<string, CameraView> = { main: "main", arm: "arm", orbit: "orbit" };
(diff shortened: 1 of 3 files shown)


theo-michel (approved):
Makes sense

PR #728, opened 2026-08-28
Fix joystick hitbox to match round pad

axelpey:
## Summary
- size the joystick interaction target to the visible pad
- make the hit target circular and prevent child visuals from extending pointer hit-testing
- use the same responsive pad size for desktop and mobile

## Testing
- all `webapp/tests/*.test.js` pass
- browser hit-test: center and circular edge hit; corners and the former oversized region do not
- `git diff --check`

## Note
- optional whole-tree `tsc --noEmit` remains blocked by pre-existing errors in Arm SDK, settings, and vendored Three.js files


DavidDobas (approved):
LGTM

PR #733, opened 2026-08-29
Fix corner map framing on the robot

axelpey:
## Summary

- keep the corner map centered on the live robot pose
- clear stale full-map pan state when the map returns to its thumbnail
- preserve normal panning when the map is expanded

## Validation

- all `webapp/tests/*.test.js` pass
- focused corner-map framing regression test passes
- touched files introduce no TypeScript errors (the full project check still reports existing unrelated errors)


axelpey:
Screenshot 2026-08-28 at 10 29 41 PM

theo-michel:
Remove test pls (Done)

theo-michel (approved):
had a review, asked claude to fix the issues it found, should be good now I guess

PR #738, opened 2026-08-29
fix(webapp): the chat transcript survives drops, refreshes and noise

DavidDobas:
Three independent faults were making the agent chat drift out of sync with what the robot was actually doing. They look like one bug from the outside — "the chat stops updating and only a reload fixes it" — but they have separate causes and separate fixes.

None of them are regressions from recent work; all three predate #717.

---

## 1. A dead socket the client never noticed

**Symptom.** Mid-conversation the chat stops updating. The robot keeps working — a reload shows everything that was missed. No error, no disconnect indicator.

**Cause.** `rosClient` reconnects correctly *when it is told the socket closed*. The problem is the case where it never is. When a TCP connection dies without a FIN — phone sleeps, WiFi roams, the proxy restarts abruptly — the browser sits at `readyState === OPEN` until the OS TCP timeout, which on mobile can be many minutes or effectively never while suspended.

The reason the client can't tell: **WebSocket ping/pong is invisible to page scripts.** The `/ws` relay pings every 60s and the browser's network stack answers automatically, but JavaScript sees none of it. So an idle-but-healthy socket and a half-open dead one are literally indistinguishable at the JS layer — silence proves nothing.

The consequence is that `onclose` never fires, `#handleClose` never runs, state stays `"connected"`, `onStateChange` never emits — and therefore the history refetch added in #717 **never gets a chance to run.** That fix only heals drops the client observes; this is the class it can't see.

**Fix, part one — stop trusting the socket.** The panel now reconciles against the brain's own record: on reconnect, on tab wake, and every 30s while the tab is visible. A snapshot whose newest entry is not newer than what is already on screen costs one service call and no DOM work:

```js
const raw = String(res?.history ?? "");
if (raw === lastSnapshot) return;
```

Comparing the snapshot to the last one replayed is deliberately dumb. An earlier version of this compared timestamps — newest-in-history against newest-rendered — and it latched itself off: a skill republishing `running` is many card updates on screen but one deduped history entry, so the rendered clock ran permanently ahead of anything the snapshot could contain, and the reconcile then skipped every poll for the rest of the session. Nothing the live stream does can move a string comparison.

This is the fix that matters, because it does not care *why* a message went missing — dead socket, dropped frame, a cause not yet found. It converges regardless.

Replaying more often is the cost, so `replay()` now keeps the reader's scroll position unless they were already at the bottom, and a failed reconcile logs instead of being swallowed.

It also improves the reconnect path from #717: that one rebuilt the whole transcript and yanked the reader to the bottom even when nothing had been missed.

**Fix, part two — notice the dead socket anyway.** Reconciliation heals the transcript but leaves the socket dead, so subscriptions, publishes and action goals stay broken until something reconnects. The `/ws` relay now emits `{"op":"keepalive"}` every 20s — a normal text frame, so `#handleFrame` sees it and stamps `#lastFrameAt` — and `rosClient` reconnects once three in a row are missed.

Three details:

- **Proxied sockets only.** `#urlFor` connects directly to `ws://<ip>:9090` when the target is not the serving host, and rws sends no keepalive. Arming the watchdog there would tear down healthy idle connections every 70s, so `#keepaliveExpected` gates it. Direct connections keep today's behaviour — no liveness detection, but no false teardowns. (Caught in review; see 6995023b.)
- **Also checked on tab wake.** Background tabs freeze timers, so the periodic check cannot be trusted to have run while away.
- **Not on `/worldstate`.** The sim viewer's parser has no reason to tolerate a frame that isn't world state.

Client and server never disagree on the cadence, because the webapp is served by the same process that sends the keepalive. Verified against the live proxy: keepalives arrive at 20.0s intervals.

---

## 2. Skill cards vanished on refresh

**Symptom.** Skill cards ("navigate to position — Failed", "drop in box — Failed") render live but disappear on reload. Chat messages survive; only the cards are lost.

**Cause.** There are two writers for skill status, and only one recorded history:

- **Manual runs** (pressing a skill in the app) went through `_on_manual_skill_event`, which appended a `task_activated` entry.
- **Agent-run skills** go through `runner.py`, which publishes to `/brain/skill_status_update` for the live UI and never touches `chat.history`.

That isn't an oversight in the runner so much as a process boundary: the runner deliberately leaves the echo to the skills server (see the comments at `runner.py:66` and `:182`), and the skills server is a **different node in a different process** — it has no way to reach the brain node's chat history. It also never publishes terminal events for agent runs itself; the action server does.

Measured on the live robot, a session's history read:

```
user            91
skill_output    30
system           6
robot            5
task_activated   0     ← every card the agent ran
```

**Fix.** The brain node now subscribes to `/brain/skill_status_update` and records from there. That is the one place every run appears regardless of who started it, and it is the only place carrying `args` — which the cards need for their parameter line (`x 2.65 m · y 4.23 m · −1° · Map`), and which the old manual-only entries were dropping anyway. Both publishers can announce the same run, so entries are deduped on `(run id, status)` over a bounded window.

---

## 3. Room tone became chat messages

**Symptom.** A small bubble containing a single `.` appears in the transcript.

**Cause.** Not a rendering artifact — a real `user` entry with text `"."`. The recogniser punctuates ambient noise, and nothing filtered it: the STT paths only `.strip()` their result, so a punctuation-only transcript is published like any other utterance. Each one also costs the agent a turn answering nobody.

**Fix.** Drop transcripts with no letter or digit anywhere, at `_handle_device_data` — the single point every input device routes through, so it covers all STT backends and any workspace input at once. `"."`, `"?"`, `"..."` are dropped; `"ok"`, `"3"`, `"Mm-hmm."`, `"ça va"` all pass.

---

## 4. Lines the robot spoke aloud never reached the transcript

**Symptom.** The robot says something over the speaker — "Got it.", "Here is the second red sock" — and nothing appears in the chat. Not a sync gap: it was never published, so a reload doesn't recover it either.

**Cause.** `Skill.say()` publishes to `/brain/tts`, and that handler called `chat.speak(text)`. `speak()` only forwards to the TTS engine; it is `emit()` that appends to history *and* publishes `chat_out`. So every line a skill voiced was audible and invisible.

`pick_any_object` alone says five of them ("Looking for {prompt}.", "Picking it up.", "Got it.", "I couldn't get a grip on it.", "My arm isn't responding properly, stopping."), and `follow_aruco` two more.

**Fix.** `_on_tts` emits instead of speaking. `speak` defaults to `True` for the `ROBOT` sender, so the audio path is byte-identical — the line just now also lands in history and on `chat_out`. `_on_environment_speech` already worked this way; its docstring ("the line reaches the chat as the voice starts") is the contract `_on_tts` was missing.

No double-publish risk: the agent's own replies reach TTS in-process through `SpeechStreamer`, never over `/brain/tts`.

---

## Also included

The Brain monitor kept its `/brain/trace` subscription forever once opened — `setVisible(false)` paused the animation loop and the camera fallback but not the heaviest feed. Measured at **627 KB mean / 748 KB max** per `turn_request` and ~230 KB/s sustained, that meant a hidden monitor parsed ~1 MB of JSON every few seconds for the rest of the session. It now follows the same on-screen-only rule as the camera fallback.

Tradeoff: turn history now covers only the stretches the monitor was actually on screen. The docstring says so.

---

## Verification

- `ruff check` + `ruff format --check` — clean
- `pytest` — 281 passed
- `basedpyright` — 15 errors, byte-identical before and after, none in the touched files
- `tsc --noEmit` — 34 errors in `js/`, identical before and after, none new

Behavioural checks against the live robot: the history service contents above, the trace payload sizes, the keepalive cadence off the running proxy (20.0s, confirmed with a websocket client), and history payload size for the poll (17.4 KB / 92 entries, ~194 B/entry).

**Not verified end-to-end in a browser:** the reconciliation healing a real stall. That needs a hard reload to pick up the new `rosClient.js` and `agentPanel.js`, then a stall to occur. The signature to look for is a missing message appearing on its own within 30s instead of needing F5.

**Known follow-up:** `ChatManager.history` is unbounded, so the poll's payload grows over a session. Fine for hours at 194 B/entry, but bounding it — or adding a cheap newest-timestamp service so the poll need not pull the whole transcript — is the clean next step.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
@@ -12,6 +12,7 @@ pubs/subs/service and delegates here.
 from __future__ import annotations
 
 import json
+import re
 import time
 from typing import Any
 
@@ -23,6 +24,10 @@ from brain_client.inputs.loader import InputLoader
 from brain_client.inputs.types import InputDevice
 
 MIC_DEVICE_NAME = "micro"
+# A transcript with no letter or digit anywhere is room tone the recogniser
+# punctuated — a bare "." or "?". Publishing it would post a junk chat bubble
+# and spend an agent turn answering nobody.
+_HAS_CONTENT = re.compile(r"[^\W_]")
 
 
 class InputDeviceManager:
@@ -72,6 +77,9 @@ class InputDeviceManager:
         try:
             text = data if isinstance(data, str) else data.get("text", "")
             if data_type == "chat_in":
+                if not _HAS_CONTENT.search(text):
+                    self._logger.debug(f"🔇 Dropped contentless transcript from '{device_name}': {text!r}")
+                    return
                 data_dict = {"text": text, "sender": "user", "timestamp": time.time()}
             else:
                 data_dict = {"text": data} if isinstance(data, str) else data.copy()
--- a/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
@@ -24,9 +24,8 @@ from brain_client.inputs.loader import InputLoader
 from brain_client.inputs.types import InputDevice
 
 MIC_DEVICE_NAME = "micro"
-# A transcript with no letter or digit anywhere is room tone the recogniser
-# punctuated — a bare "." or "?". Publishing it would post a junk chat bubble
-# and spend an agent turn answering nobody.
+# Room tone the recogniser punctuated — a bare "." or "?". Publishing it posts
+# a junk bubble and spends an agent turn answering nobody.
 _HAS_CONTENT = re.compile(r"[^\W_]")
 
 
--- a/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/inputs/manager.py
@@ -24,8 +24,6 @@ from brain_client.inputs.loader import InputLoader
 from brain_client.inputs.types import InputDevice
 
 MIC_DEVICE_NAME = "micro"
-# Room tone the recogniser punctuated — a bare "." or "?". Publishing it posts
-# a junk bubble and spends an agent turn answering nobody.
 _HAS_CONTENT = re.compile(r"[^\W_]")
 
 
--- a/ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
@@ -402,10 +402,12 @@ class BrainClientNode(Node):
         self.brain.on_custom_input(data)
 
     def _on_tts(self, msg: String) -> None:
+        """Speak a line a skill sent, and show it — emit, not speak: anything the
+        robot says aloud belongs in the transcript, or Skill.say goes unrecorded."""
         text = msg.data
         if text and text.strip():
             self.get_logger().info(f"TTS request received: {text[:50]}...")
-            self.chat.speak(text)
+            self.chat.emit(Sender.ROBOT, text)
 
     def _on_environment_speech(self, payload: dict) -> None:
         """Speak a simulated character: the line reaches the chat as the voice
--- a/ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/nodes/brain_client_node.py
@@ -522,13 +522,11 @@ class BrainClientNode(Node):
         )
 
     def _on_skill_status(self, msg: String) -> None:
-        """Record every skill run in the chat history, so a replayed transcript
-        keeps its cards.
+        """Record every skill run, so a replayed transcript keeps its cards.
 
-        The live topic is the only place an agent-run skill ever appears — the
-        runner leaves the echo to the skills server, which is a different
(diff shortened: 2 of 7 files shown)


DavidDobas:
Good catch — that was a real bug, fixed in 6995023b.

`#urlFor` only routes through the proxy when the target is the host that served the page:

```js
if (this.#isServingHost(ip)) {
  const scheme = location.protocol === "https:" ? "wss" : "ws";
  return `${scheme}://${location.host}/ws`;
}
return `ws://${ip}:${ROSBRIDGE_PORT}`;   // straight to rws, no keepalive
```

On that second path nothing fills the silence, so an idle-but-healthy connection would go 70 s without a frame and the watchdog would tear it down — rejecting in-flight service calls and orphaning action goals every 70 s, forever. Worse than the bug it was added to fix.

The watchdog is now armed only where the keepalive actually reaches:

```js
// A direct rws socket gets no keepalive and is legitimately silent when idle.
this.#keepaliveExpected = this.#isServingHost(ip);
```

`#keepaliveExpected` gates both the interval and `#dropIfStale` itself, so the `visibilitychange` path can't fire on a direct socket either. Direct connections keep exactly today's behaviour — no liveness detection, but no false teardowns.

The remaining gap is that a direct connection still can't detect a half-open socket. That needs a keepalive from rws, which is a third-party binary we don't control, so it's out of scope here. Worth noting the webapp is normally served by the proxy, so the common path is covered.

Verification after the change: `ruff` clean, 281 tests pass, `tsc --noEmit` 34 errors in `js/` (unchanged baseline), `basedpyright` 15 (unchanged baseline).

axelpey (approved):

PR #741, opened 2026-08-29
Fix settings callbacks and add demo utility skills

axelpey:
## Summary

- restore dispatch for slow/admin callbacks so `/set_volume`, `/set_microphone`, and `/robot/info` respond reliably
- add a narrowly scoped `ChangeVolume` skill that can only change speaker volume from 0–100
- add a `CheckBattery` skill and expose both skills from Demo Agent

The volume skill intentionally does not expose microphone, driving-sound, TTS voice, motion, or hardware-safety settings.

## Validation

- `python3 -m py_compile` for the new skills and Demo Agent
- Ruff check and format check
- XML validation and `git diff --check`
- successful `mars_control` release build on `mars-the-blue.local`
- full robot reboot; `ros-app.service` active with `NRestarts=0` and no failed units
- live websocket `/set_volume` call returned success
- `/robot/info` publishes once per second with the persisted volume
- live directive roster shows `change_volume` and `check_battery` on Demo Agent, with no broken agents and `brain_active=false`

## Operational note

The robot web app responds over HTTP and HTTPS. Its local HTTPS certificate is self-signed, so strict browsers may still show a certificate warning; certificate trust is outside this PR.

### DIFF
--- a/workspace/innate_skills/change_volume.py
+++ b/workspace/innate_skills/change_volume.py
@@ -1,7 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (c) 2026 Innate Inc
-import time
-
 from mars_msgs.srv import SetVolume
 
 from innate import Skill, SkillReturn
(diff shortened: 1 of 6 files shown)


workspace/innate_skills/change_volume.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def execute(self, speaker_volume: int) -> SkillReturn:
+        volume = self._integer(speaker_volume, "speaker_volume", 0, 100)
+        response = self._call(SetVolume, "/set_volume", SetVolume.Request(volume_percent=volume))
+        if not response.success:
+            self.fail(response.message or "The robot rejected the volume change.")
+        return response.message or f"Speaker volume set to {volume}%"
+
+    def _call(self, service_type, name: str, request, timeout: float = 5.0):
+        if self.node is None:
```
DavidDobas, line 27: Just thinking if it is called modify settings, we should either make it change more different settings, or rename to ChangeVolume

DavidDobas (approved):

workspace/innate_skills/change_volume.py
axelpey: it was like that at first but there's a lot of decisions to have regarding what more skills to have in there (or not have) and i feel like this should be decided later

PR #746, opened 2026-09-02
feat(skills): pickAnything rolls the gripper onto the object's short side

theo-michel:
## What

`pick_any_object` always closed the gripper at the same roll. Its 81 mm jaw is narrower than most objects' long side, so anything elongated lying across the fingers was missed. The wrist stage now measures the object's main axis from the object-vs-floor mask it already tracks, and rolls the wrist so the fingers close across the object's **short** side.

## How

- **`innate.vision.seg_track`** also returns the blob's `Axis` (major-axis angle in image coordinates, elongation). It is the min-area rectangle of the Otsu-thresholded back-projection component under the CamShift window. CamShift's own ellipse was tried first and rejected: its window only hugs part of a blob that grows during the descent, and the angle drifted 0°→35° over one descent while the mask rectangle stayed within 3–6° at every height.
- **Roll = image angle − 90°.** The wrist camera rolls with the fingers, which close along image horizontal, so the blob's image angle is already relative to the jaw; no frame conversion. The last reading above `AXIS_MIN_Z` (7 cm) is used, because below that the fingers straddle and clip the blob in the wrist view. Blobs rounder than `MIN_ELONGATION` keep roll 0 (socks, balls, cubes behave as before).
- **Rolled grasps descend with a vertical tool axis** (`ROLLED_PITCH`). At `arm_pitch` 1.30 a 90° roll puts one fingertip 3 cm below the other; the low finger lands on a 3 cm bar and the descent stalls with the other pad above it (measured in MuJoCo, reproduced in the ROS stack at ee z 0.047 → miss). The real IK reaches π/2 at the push heights across the pick box.
- **IK yaw = the arm's own j1** for the push/close moves. At vertical pitch RPY is gimbal-locked and a yaw-0 target makes the solver put all of j1 into j5 (up to 22° of roll error, seen as j5 = roll + j1 in the logs). With yaw = j1 the target is exactly reachable and j5 == roll in every trial.

## Sim evidence

Full ROS stack in `innate-sim`, real skill via `innate skill run innate-os/pick_any_object @prompt="the orange bar"`, bar prop 1 m ahead of spawn at four world yaws, ground truth from the world-server stream.

| | before | after |
|---|---|---|
| bar orientations graspable | 1 of 4 (along x only) | 4 of 4 |
| bar runs, final code | | 5 / 7 held |
| sock runs (no-regression) | | 2 / 2 held |

The two misses were marginal pinches (j6 0.20–0.26 at close) that slipped during the twist/lift, not roll errors; measured roll was within 5° of ideal in every run. Rig physics (headless `VirtualMars`) says the jaw tolerates ~15–30° of roll error.

## Hardware checks before trusting this on a MARS

- Roll sign was verified in sim only. If a real wrist camera is mirrored, negate in `_grasp_roll` (one line, noted in its docstring).
- The vertical pitch for rolled grasps is untested on hardware; the IK reaches it for x 0.25–0.31 m at z ≤ 0.10 m, not at x 0.36 m.
- Pre-existing, not fixed here: the wrist CamShift tracker occasionally never locks on a thin diagonal bar, in which case the skill falls back to today's unrolled blind push.

## Checks

ruff + format clean on `brain_client/` and the skill, basedpyright 0 errors on `innate/vision.py`, 281 brain_client tests pass in the sim container.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/robot/manipulation.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/robot/manipulation.py
@@ -344,6 +344,12 @@ class Manipulation:
         """Clamp (x, y) into the graspable reach box."""
         return (max(cls.REACH_X[0], min(cls.REACH_X[1], x)), max(cls.REACH_Y[0], min(cls.REACH_Y[1], y)))
 
+    def reachable(
+        self, x: float, y: float, z: float, *, roll: float = 0.0, pitch: float = 0.0, yaw: float = 0.0
+    ) -> bool:
+        """Whether IK finds a solution for the pose; the arm does not move."""
+        return self._solve_ik(x, y, z, roll, pitch, yaw) is not None
+
     # --- motion ---
 
     def move_to(
--- a/ros2_ws/src/brain/brain_client/innate/vision.py
+++ b/ros2_ws/src/brain/brain_client/innate/vision.py
@@ -200,9 +200,10 @@ def seg_model(hsv, box):
 # +v, i.e. clockwise on screen, in [0, pi)) and major/minor elongation
 # (1 = round).
 Axis = tuple[float, float]
+Window = tuple[int, int, int, int]
 
 
-def _blob_axis(bp, window) -> Axis | None:
+def _blob_axis(bp: np.ndarray, window: Window) -> Axis | None:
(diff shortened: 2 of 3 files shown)


DavidDobas (approved):
Works in sim and on real robot, small diff, approving

PR #747, opened 2026-09-03
sim: share one Three.js stage across pages, plus Respawn and touch panel fixes

DavidDobas:
Started as one fix and picked up a handful of small sim/UI items along the way. They're described separately below; the first is the substantial one.

## 1. One Three.js stage for the whole app

Switching between Agent and Teleop in sim crashed the tab on a phone, and after a few crashes Safari showed *"A problem repeatedly occurred"* for the origin.

The sim stage was per-page. `robotSession.js` gave real robots the shared `WebRtcSession` but said sim sessions "stay per-page: they render local canvases, so there is no link to keep warm" — a local canvas turned out not to be cheap. Every rail click destroyed the stage and built a new one: a fresh GL context, the apartment + robot URDF/STLs, and `prefetchPropModels()` reparsing every prop glb (the human alone is 23 MB of glb → 67 MB of decoded texture), unconditionally, on both pages.

Measured against a running sim, 5 agent↔teleop round trips:

| | before | after |
|---|---|---|
| model/mesh requests per switch | 22 | **0** |
| renderer+GPU RSS over 5 round trips | **+4.4 GB** (~450 MB a switch, never reclaimed) | +20 MB |
| JS heap after forced GC | +3 MB per round trip, linear | flat, 7–8 MB |

The memory not coming back is what made this fatal rather than merely slow.

- **`simStage.ts`** — `attach(parent)` / `detach()`. Detach stops the render loop (rAF is throttled for a hidden tab but not for a detached canvas), reclaims the debug stack the agent page lifts into its own layout, clears the agent dock's safe-inset and any armed prop, and parks the wrap out of the DOM.
- **`robotSession.js`** — one module-level sim session + stage with acquire/release and a 60s linger, mirroring `sharedVideoSession`. A page's `stage.destroy()` only detaches.
- **`scene.ts`** — `frameRobot()`, the camera half of `spawnAt` extracted. `spawnAt` fires once per *session*, so a shared session lost the "face the robot on entry" framing; `attach()` restores it along with free orbit mode, matching what a rebuilt scene gave.

## 2. Respawn chip

A `Robot › Respawn` chip in the sim's Scene setup panel: robot back to its spawn pose, arm home, every prop parked.

`/virtual_mars/reset` already did exactly this. It was built for a Reset button in the old standalone viewer (the comment in `node.py` still said so) and survived into a state where **nothing published it** — reachable only by hand with `ros2 topic pub`. The chip publishes it over the rosbridge socket the app already holds, rather than adding a world-server stage command, so `_on_reset` fails the in-flight arm trajectory before teleporting the base.

**This changes a topic's message type, which is the part worth a careful look.** `/virtual_mars/reset` was `std_msgs/Empty`, and a browser cannot publish that: rws cannot deserialize a zero-field message and *takes the subscriber node down with it* — the reason `STEREO_CALIB_CAPTURE_TOPIC` is already a Bool. So the subscription is now `std_msgs/Bool`. Since the topic had no publishers, the only thing this breaks is a hand-typed `ros2 topic pub ... std_msgs/msg/Empty`; `sim/README.md` now names the type. If you'd rather not touch a published contract, the alternative is a second Bool topic alongside the Empty one, at the cost of two topics doing one job.

## 3. Touch: tap away to close

Scene setup and Challenges could only be closed with their ✕. On a phone both cover the stage, so a pointerdown outside them now closes them. Gated on `(hover: none)`, matching how Scene setup already decides touch behaviour — mouse behaviour is unchanged, where clicking the scene is how an armed prop gets placed. A running challenge still holds its goal list, the rule `dismiss()` already followed.

## 4. Compact agent sheet keeps the newest message

Dragging the sheet shorter left `scrollTop` where it was, pinning the top of the view and pushing the last message out of sight. The transcript now follows its own height — but only for a reader who was already at the bottom, so scrollback is still never yanked.

## 5. Scene setup fits a phone screen

The panel grows upward from a fixed bottom with no height cap, so on a phone its top ran off the stage — Objects and Clear all were simply unreachable. It is now capped to the room above the debug stack, with the object roster taking the squeeze since that is the part that grows with the prop list. The panel becomes a flex column so the roster can actually shrink (grid auto rows do not), with a floor of one row and a scroll on the panel itself as a last resort, so no control is ever cut off.

The section floors at one whole row rather than shrinking on, so the roster scrolls while a row fits and the panel itself scrolls only once it does not; short screens also tighten their gaps. Measured across viewports (agent and teleop, panel open): fits at 844, 700, 667 and 640px tall, where it previously overflowed the top by up to 120px, with no section overlapping another. Desktop is unchanged.

## 6. Cockpit runs under the wordmark on phones

On phones `body` is a two-row grid and the stage sat in row 2, so every cockpit opened below a black band holding the wordmark. The scene is the page there, so it now spans both rows with the wordmark over it. Other pages keep the band — their headers vary too much for each to hold itself clear. The camera strip and the top-left overlay stack both clear the 64px wordmark.

## 7. Eye, not a camera, for the view-tiles toggle

The button that folds the camera tiles away carried a camera-body glyph, which reads as "take a photo". It is an eye now; same stroke weight, same `Camera views` label.

## 8. Run closes the skills menu on touch

Running a skill left the menu open and focused its search box, so the next pick needs no reach for the mouse. On a phone that raises the on-screen keyboard over the stage you launched the skill to watch, and there is no hardware keyboard to hand control back to — so on `(hover: none)` running closes the menu instead. Desktop is unchanged: the menu stays open with the search box focused.

## 9. Skills search had no bottom margin

`margin: 8px 8px 0` left the search field flush against the legend band below it on desktop. Phones already overrode it to `8px` all round, which is why only desktop looked wrong; the base rule now matches and the phone override is down to the reorder that earns it.

## Verified

Browser-driven against a live sim (headless Chrome over CDP, plus an emulated iPhone viewport for the touch items):

- Same canvas element across both page switches, 0 model refetches, no loading pill, scene-setup panel survives the round trip. A detour to Settings parks the stage and returns it intact; past the linger it rebuilds cleanly. 0 console errors.
- Camera framing pixel-identical to the opening view after dragging away and navigating back; mode chips agree.
- The Respawn chip renders and publishes — the topic reads `Publisher count: 1` with the browser as a `std_msgs/Bool` publisher.
- Tap outside closes both panels; taps inside keep them open (the prop chip still closes Scene setup, which is the pre-existing arming behaviour).
- Shrinking the sheet keeps the transcript at the bottom; a scrolled-up reader is left alone.
- The scene setup panel fits every phone viewport tested, and its roster scrolls under a touch drag without dismissing the panel.
- Running a skill on an emulated phone closes the menu with nothing focused; on a real-mouse desktop the menu stays open and the search box keeps focus.
- `tsc --noEmit` clean on `sim/viewer`; the touched webapp files add no new errors (`skillsMenu.js` carries 5 pre-existing ones, unchanged before and after).

`ruff` was not available on this machine, so `node.py` is unlinted here — it is a type swap plus a comment, and CI's pre-commit will cover it.

## Note for reviewers

Two behaviours now persist across a page switch that previously reset with the scene: **camera mode** is deliberately reset to `free` on entry (matching the old behaviour), but **prop placements and the scene-setup panel state** now carry over. That reads as the improvement, not a regression — worth a second opinion.

### DIFF
--- a/webapp/js/robotSession.js
+++ b/webapp/js/robotSession.js
@@ -20,6 +20,8 @@
 import { WebRtcSession } from "./webrtcSession.js";
 import { acquireVideoSession, releaseVideoSession } from "./sharedVideoSession.js";
 import { getConfig } from "./config.js";
+import { ros } from "./rosClient.js";
+import { SIM_RESPAWN_TOPIC } from "./constants.js";
 
 /** @typedef {{ audioEl: null, attach: (parent: HTMLElement) => void, detach: () => void, destroy: () => void, setSafeInsets: (insets: { right?: number }) => void }} SimStage */
 
@@ -46,6 +48,8 @@ export async function robotSessionFactory() {
       // URL, not a module path tsc can resolve.
       // @ts-ignore
       const mod = await import("/sim-viewer/sim-session.js");
+      // rws resolves a bare publish against the graph; nothing else publishes here.
+      ros.advertise(SIM_RESPAWN_TOPIC, "std_msgs/msg/Bool");
       return {
         createSession: () => acquireSimSession(mod, config),
         releaseSession: () => releaseSimSession(),
@@ -93,7 +97,10 @@ function releaseSimSession() {
  */
 function acquireSimStage(mod, root, session) {
   if (simStage) simStage.attach(root);
-  else simStage = /** @type {SimStage} */ (mod.createSimStage(root, session));
+  else {
+    const respawn = () => ros.publish(SIM_RESPAWN_TOPIC, { data: true });
+    simStage = /** @type {SimStage} */ (mod.createSimStage(root, session, respawn));
+  }
   const stage = simStage;
   return { ...stage, destroy: () => stage.detach() };
 }
--- a/webapp/js/robotSession.js
+++ b/webapp/js/robotSession.js
@@ -10,15 +10,27 @@
 //   ...inside buildCockpit:
 //     const session = createSession();
 //     const stage = createStage ? createStage(root, session) : createVideoStage(root, session);
-//   ...inside destroy: releaseSession(session), never session.destroy().
+//   ...inside destroy: stage.destroy() then releaseSession(session), never session.destroy().
 //
-// Real robots hand every page the app-level shared WebRtcSession (see
-// sharedVideoSession.js), so page switches reuse the live link. Sim sessions
-// stay per-page: they render local canvases, so there is no link to keep warm.
+// Either backend hands every page ONE app-level session: real robots keep the
+// live WebRTC link (sharedVideoSession.js), and the sim keeps its whole stage,
+// whose page-facing destroy() only detaches it. Rebuilding that stage per page
+// never gave its memory back, which killed the tab on a phone (simStage.ts).
 
 import { WebRtcSession } from "./webrtcSession.js";
 import { acquireVideoSession, releaseVideoSession } from "./sharedVideoSession.js";
 import { getConfig } from "./config.js";
+
+/** @typedef {{ audioEl: null, attach: (parent: HTMLElement) => void, detach: () => void, destroy: () => void, setSafeInsets: (insets: { right?: number }) => void }} SimStage */
+
+// Long next to sharedVideoSession's 10s: a detour to Settings must not pay to
+// reparse ~80 MB of models.
+const SIM_LINGER_MS = 60_000;
+
+/** @type {any} */ let simSession = null;
+/** @type {SimStage | null} */ let simStage = null;
+let simHolders = 0;
+/** @type {number | null} */ let simLinger = null;
 
 /**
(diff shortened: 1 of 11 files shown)


webapp/js/robotSession.js
```diff
@@ -35,13 +47,53 @@ export async function robotSessionFactory() {
...
       console.error("[robotSession] sim viewer bundle unavailable, falling back to WebRTC:", err);
     }
   }
   return { createSession: acquireVideoSession, releaseSession: releaseVideoSession, createStage: null };
 }
+
+/**
+ * @param {any} mod
+ * @param {any} config
+ */
```
axelpey, line 68: tf is this docstring lol

axelpey (approved):
maybe remove some stupid docstrings first tho

PR #748, opened 2026-09-03
feat(agent): add Enter shortcut for chat composer

axelpey:
## Summary

- focus the Agent chat composer when Enter is pressed from the neutral page
- show Teleop’s existing compact `↵` keycap treatment inside the empty Agent text box
- leave focused controls, typing fields, modified Enter, repeats, and already-consumed events alone

## Testing

- `node --check webapp/js/agent/agentPanel.js`
- `git diff --check origin/main...HEAD`
- browser-verified the Agent keycap and Enter-to-focus behavior against the running simulator

### DIFF
--- a/webapp/css/app.css
+++ b/webapp/css/app.css
@@ -9193,8 +9193,18 @@ select.skill-input {
   transition-duration: 100ms;
 }
 
+.agent-compose-focus-key {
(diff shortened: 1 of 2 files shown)


DavidDobas (approved):
Good

PR #750, opened 2026-09-03
fix(sim): play Mad Mars sound in browser

axelpey:
## Summary
- launch the existing Mad Mars motor-sound synth in simulation
- stream the exact configured synth to the webapp because the simulator container has no audio device
- schedule PCM with a bounded browser queue and single-tab speaker ownership
- stop and clear queued motor audio when the browser tab is backgrounded
- leave the physical robot ALSA path unchanged

## Validation
- `uv run --no-project --with pytest --with pyyaml --with numpy python -m pytest tests/ -q` (68 passed)
- focused browser PCM decode, scheduling, backlog, and reset test
- all other existing webapp Node tests pass; the preload test has unrelated baseline drift on current main
- `pre-commit run --all-files --show-diff-on-failure -c .config/pre-commit-config.yaml`
- live simulator validation: Mad published non-silent `/motor_sound/audio` at ~10 Hz with rosbridge subscribed; returning to Med stopped publication

This PR intentionally contains no changes to Nav2, navigation speed, speed-mode behavior, settings, or speed-mode UI.


axelpey:
actually a bad idea for the real rbot

DavidDobas (commented):
The Web Audio scheduler is justified — `ttsAudio.js` plays discrete WAV clips through an `HTMLAudioElement`, which has tens of ms of start latency and no sample-accurate scheduling, so chaining 100 ms chunks through it would click and stutter. Gapless playback needs `AudioBufferSourceNode.start(at)` on a shared clock. But the browser half can shrink to about one module:

- **Delete `tests/pcmAudioPlayer.test.js`.** Project rule: tests written while developing come out before the PR unless tests were asked for. The class lives in its own module mainly so that test can import it.
- **Fold `PcmAudioPlayer` into `motorSoundAudio.js`.** Without the test there is no reason for two files. That makes one ~90-line module beside `ttsAudio.js`, and the extra `modulepreload` line in `index.html` goes away.
- **Reuse the base64 decoder.** `ttsAudio.js` already has `base64ToBytes`; export it instead of re-implementing the same loop inside `push`.
- **Optional: drop the source tracking.** The `Set` of sources and the `onended` bookkeeping exist only so `reset` can stop queued audio when the tab is hidden. `context.suspend()` already silences output, and the backlog is capped at 0.5 s, so on resume at most half a second of stale hum plays. If that is acceptable, `reset` becomes `nextAt = 0` plus the suspend, and another ~15 lines disappear.

Unrelated to size: the TTS player sets the mic-gating flag so the robot does not hear itself, while the motor hum has no such gating. With the mic open in sim, the hum is fed to STT continuously. Probably harmless, just noting it.

DavidDobas:
Pushed 4d86f598 applying the review above: the PCM scheduler is folded into `motorSoundAudio.js` as a single `schedule` function, `pcmAudioPlayer.js` and its test are deleted along with the extra `modulepreload`, `ttsAudio.js` exports its existing `base64ToBytes` for reuse, and a backgrounded tab now just resets the clock and suspends the context instead of tracking every scheduled source.

Net: 52 insertions, 127 deletions, i.e. 75 lines removed. The webapp side of the PR goes from 170 added lines across 6 files to 96 across 4. No TypeScript errors in the touched files; the other webapp Node tests pass; the preload test still fails only on the pre-existing `agentSheet.js`/`keyboardInset.js` drift.

DavidDobas (approved):
Slimmed down, approving

PR #763, opened 2026-09-05
fix(sim): seed the demo with every environment's map, not only the apartment

DavidDobas:
Switching the public demo to the **backrooms** environment hangs on
"loading robot and backrooms…"; the console says
`RuntimeError('Nav2 did not load backrooms.yaml')`.

The map is in the assets image. The demo launcher copied only
`sim_apartment.*` into `data/maps/`; the dev launcher copies every
environment pack's `.yaml`/`.pgm`, and its comment explains why. The demo now
does the same.

Surfaced by the first CI-built demo image, whose asset layer is the first to
include the backrooms environment.


DavidDobas:
Superseded by #764.

PR #765, opened 2026-09-05
Remove chess agents and skills

axelpey:
Removes the shipped Chess Agent and Board Calibration Agent, all five chess skills (piece movement, opponent detection, calibration, game reset, and state updates), and the now-unused `python-chess` dependency. These agents and skills will no longer be discovered from the shipped workspace.

Adds a ROS-independent check for dangling imports into shipped agent and skill modules. Shared arm/head skills remain available.

Validation:
- Workspace import-target test: passed.
- Audited remaining code for references to the deleted modules/classes: none found.
- `git diff --check`: passed.
- Existing agent-loading suite could not collect on macOS because `rclpy` is unavailable; live ROS discovery was not verified.


tests/test_workspace_import_targets.py
```diff
@@ -0,0 +1,33 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 Innate Inc
+"""Catch dangling shipped agent/skill imports without requiring ROS hardware."""
+
+import ast
+from pathlib import Path
+
```
DavidDobas, line 9: Is this test needed?

DavidDobas (approved):
Check if test needed

tests/test_workspace_import_targets.py
axelpey: it is; in the future we might remove more skills / agents and we want to make sure that the ones provided by default don't get messed up if a skill has been worngfull deleted

PR #768, opened 2026-09-05
Replace skill category color dots with icons

axelpey:
Replace skill-category dots with small SVG icons: a green network for Learned, a blue replay arrow for Replay, and purple code brackets for Digital. Rows and the legend share the icons; row icons retain accessible category names, while legend icons are decorative to avoid duplicate announcements. Unknown and failed-load skills have fallback symbols.

Validation: visually checked in the browser and exercised search, parameter entry, and Enter-to-run/close using a temporary fixture. No test fixture is included in this PR. `git diff --check` passes. The local Node suite previously passed 7/8 files, with an existing preload-test failure; the project typecheck also has existing errors.


webapp/tests/skillsMenu.html
```diff
@@ -0,0 +1,75 @@
...
+<pre id="result">Ready</pre>
+<div id="launcher"></div>
+<script type="module">
+import { createSkillsMenu } from '../js/teleop/skillsMenu.js';
+import { AVAILABLE_SKILLS_TOPIC } from '../js/constants.js';
+const callbacks = new Map();
+const goals = [];
+const ros = {
+  state: 'connected',
+  subscribe(topic, callback) { callbacks.set(topic, callback); return () => callbacks.delete(topic); },
```
DavidDobas: Remove the test

DavidDobas (approved):
Remove the test before merging

PR #769, opened 2026-09-05
ci(sim): point staging at every demo image main publishes

DavidDobas:
## What

After the publish workflow pushes `innate-os-sim-demo:sha-<short>` from `main`, one more step logs in to the **staging** broker's admin (`sim-staging.svc.innate.bot`), sets that tag as the simulator image, and retires the sessions that are already READY, since those were built on the previous image and the pool only applies the new image to sessions it builds from then on.

Staging therefore always hands out the newest simulator. Production is unchanged: promoting is still pasting the same tag into production's admin page.

## Why

The counterpart of innate-cloud's deploy pipeline (innate-cloud #100/#101): both halves of the demo, the broker and the simulator image, land on staging by themselves; both promotes to production stay deliberate.

## Needs

Two repository secrets: `SIM_STAGING_URL` (staging's origin, e.g. `https://…`) and `SIM_STAGING_ADMIN_TOKEN` (the `ADMIN_TOKEN` key of the `sim-broker-secrets` secret in the `sim-staging` namespace). Neither is in the repo: this is a public repository, and a clone has no staging to point at, so with either missing the step logs a notice and does nothing. The image is published regardless.

## Verified

- Workflow parses and passes `actionlint`.
- The three admin calls (form login, `POST /admin/settings` with `simImage`, `POST /admin/session/<id>/destroy`) are the ones used by hand against staging earlier today.

### DIFF
--- a/.github/workflows/publish-sim-demo.yml
+++ b/.github/workflows/publish-sim-demo.yml
@@ -66,10 +66,12 @@ jobs:
           echo "viewer=$(resolve viewer)" >> "$GITHUB_OUTPUT"
 
       - name: Submit Cloud Build
+        id: build
         shell: bash
         run: |
           set -euo pipefail
           short_sha="$(git rev-parse --short HEAD)"
+          echo "tag=sha-${short_sha}" >> "$GITHUB_OUTPUT"
           # Only main moves `latest`.
           latest_tag=""
           if [ "${GITHUB_REF}" = "refs/heads/main" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-sim-demo.yml
```diff
@@ -85,3 +87,24 @@ jobs:
...
+          IMAGE: ${{ vars.GCP_REGION }}-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/sim/innate-os-sim-demo:${{ steps.build.outputs.tag }}
+        run: |
+          set -euo pipefail
+          jar="$(mktemp)"
+          curl -fsS -o /dev/null -c "$jar" --data-urlencode "token=${ADMIN_TOKEN}" "${STAGING}/admin/login"
+          curl -fsS -o /dev/null -b "$jar" -H 'content-type: application/json' \
+            -d "{\"simImage\":\"${IMAGE}\"}" "${STAGING}/admin/settings"
+          for id in $(curl -fsS -b "$jar" "${STAGING}/admin/state" | jq -r '.sessions[] | select(.state == "ready") | .machineId'); do
+            curl -fsS -o /dev/null -b "$jar" -X POST -H 'content-length: 0' "${STAGING}/admin/session/${id}/destroy"
+          done
```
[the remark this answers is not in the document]
DavidDobas: Applied as suggested in be8afeca: the state response is captured and parsed before the loop, so a failed /admin/state or malformed JSON fails the step instead of silently retiring nothing.

.github/workflows/publish-sim-demo.yml
```diff
@@ -85,3 +87,26 @@ jobs:
...
           echo "Published sha-${short_sha}${latest_tag:+ and }${latest_tag}" >> "$GITHUB_STEP_SUMMARY"
+
+      # Staging always hands out the newest image; production is promoted by
+      # pasting the same tag into its admin page. Ready sessions were built on
+      # the previous image, so they are retired and the pool rebuilds them.
+      - name: Point staging at this image
+        if: github.ref == 'refs/heads/main'
+        shell: bash
+        env:
+          STAGING: https://sim-staging.svc.innate.bot
```
axelpey: would prefer having ythis a secret tbh

axelpey (changes requested):

.github/workflows/publish-sim-demo.yml
DavidDobas: Done in 054bc49: the address is now the SIM_STAGING_URL secret, and the step does nothing when either secret is absent, so a clone of the repo never tries to reach it.

axelpey (approved):

PR #770, opened 2026-09-05
sim: warn on sustained slowdown with opt-in resolution reduction

axelpey:
## Summary

Split the viewer slowdown warning out of #753, as requested in review.

- Keep full rendering resolution by default. Warn after two consecutive 3 s windows under 24 fps or with the simulation clock under 0.8x real time; never reduce resolution automatically.
- Offer explicit Reduce resolution / Use full resolution controls. Reduction always shrinks the buffer, including on small DPR-1 displays, respects a 3-million-pixel ceiling, and applies even while the stage is hidden behind the map. While reduced, the notice stays as the way back (it cannot be dismissed); restoring full resolution hides it and lets the detector speak again.
- Never warn on hidden tabs, parked stages, clock rollbacks, 30 Hz display scheduling, or while the loading scrim is up (client mesh loads, server world rebuilds, failures). A window whose world-state stream stalled is judged on frame rate alone, so a stall neither reads as slow physics nor blinds the detector.
- Dismissal is remembered for the tab. The resolution choice lasts until reload and survives the shared stage's rebuild after its 60 s linger (module state, not stage state).

## Verification

- `sim/viewer/test/slowdown.test.ts` pins sustained slowdown, 30 Hz jitter, single-window recovery, resets, inactive views, clock rollbacks (including one landing above the window's baseline), and stalled streams. Typecheck and bundle build pass.
- Browser (Chrome, DPR 2, 1216x720 stage): Reduce → 1824x1080 buffer, also with the map primary; leaving for Settings 70 s rebuilds the stage and keeps it reduced with the notice shown; Use full → 2432x1440 and the notice hides; dismissal persists across reload.
- Reviewed by Codex and Claude Opus in independent sessions; every should-fix finding is addressed above.


theo-michel (approved):
LGTM

PR #772, opened 2026-09-06
feat(brain): configure per-agent thinking intervals

axelpey:
Agents can override the pause after each completed thought with two class attributes:

```python
idle_turn_interval = 0.5
supervision_turn_interval = 0.3
```

Both default to `None`, retaining the configured global intervals (3s/5s by default). The loader validates finite positive numbers and caches them once, so the turn loop and heartbeat read resolved values without calling workspace getters. Changes take effect on agent reload; the brain trace reports the effective interval.

Scoped to cadence; model/backend support is handled separately by #824. Existing tests cover overrides, defaults, cached values and invalid configuration; no new test functions. All 62 brain tests and 19 loader tests pass, along with Ruff and diff checks. The local onboarding trial's 0.01s override remains outside this PR.

### DIFF
--- a/ros2_ws/src/brain/brain_client/brain_client/agents/types.py
+++ b/ros2_ws/src/brain/brain_client/brain_client/agents/types.py
@@ -7,9 +7,7 @@ Agent Type Definitions
 Base class and types for robot agents.
 """
 
-import math
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
 from typing import TYPE_CHECKING, TypeAlias, Union
 
 from brain_client.common.script_paths import Source
@@ -29,25 +27,6 @@ SkillRef: TypeAlias = Union["type[Skill]", "type[TrainedSkill]", str]
 InputRef: TypeAlias = Union["type[InputDevice]", str]
 
 
-@dataclass(frozen=True)
-class TurnIntervals:
-    """Optional per-agent overrides for the brain's visual observation cadence.
-
-    ``None`` keeps the global ROS parameter.  A positive value is the pause
-    after a completed model turn; model latency is additional.  Keeping this
-    on the agent lets a visually supervised navigation agent look frequently
-    without making every directive on the robot equally chatty and expensive.
-    """
-
-    idle: float | None = None
-    supervision: float | None = None
-
-    def __post_init__(self) -> None:
-        for name, value in (("idle", self.idle), ("supervision", self.supervision)):
-            if value is not None and (not math.isfinite(value) or value <= 0):
-                raise ValueError(f"turn interval {name} must be a finite positive number or None")
-
-
 class Agent(ABC):
     """
     Base class for all agents.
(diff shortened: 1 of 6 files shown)


ros2_ws/src/brain/brain_client/innate/__init__.py
```diff
@@ -99,6 +99,7 @@ def execute(self):
 
 __all__ = [
     "Agent",
+    "TurnIntervals",
     "Arm",
```
DavidDobas, line 102: Seems alphabetical, put at the right place

DavidDobas:
`_interval()` now calls into workspace code, and `_snapshot()` calls `_interval()`. The heartbeat that drives `_snapshot()` is a bare loop with no error boundary:

```python
async def _heartbeat(self) -> None:
    while True:
        self._lidar.tick()
        self._snapshot()
        await asyncio.sleep(1.0)
```

So an agent whose `get_turn_intervals()` raises, or returns something without `.idle` on a later call, takes the heartbeat task down and telemetry stops for the rest of the session. The `isinstance` check in `build_agent_instances` only covers the single call made at construction; nothing constrains later calls.

Suggestion: resolve the override once in `build_agent_instances`, where every other agent surface is already probed and where a failure becomes a broken-roster row with its message. Store the resolved floats on the agent, and have `_interval()` read those. The brain then never calls workspace code on the hot path, and it also stops rebuilding and re-validating a `TurnIntervals` on every heartbeat tick and every turn.

ros2_ws/src/brain/brain_client/brain_client/agents/types.py
```diff
@@ -27,6 +29,25 @@
...
+    without making every directive on the robot equally chatty and expensive.
+    """
+
+    idle: float | None = None
+    supervision: float | None = None
+
+    def __post_init__(self) -> None:
+        for name, value in (("idle", self.idle), ("supervision", self.supervision)):
+            if value is not None and (not math.isfinite(value) or value <= 0):
+                raise ValueError(f"turn interval {name} must be a finite positive number or None")
```
DavidDobas: To me it's a bit strange that agents will have to import a dataclass just to set two attributes of the agent. 
How about idle_turn_interval = 0.5 and supervision_turn_interval = 0.3 as annotated class attributes on Agent, defaulting to None?

Maybe one day we could make agents even simpler, something like
```
class SecurityGuard(Agent):
    id = "security_guard"
    display_name = "Security Guard"
    display_icon = "assets/security_guard.png"
    skills = [SearchMemory, NavigateToPosition]
    inputs = [MicroInput]
    gaze = True
    idle_turn_interval = 0.5
    prompt = """You are a security guard robot..."""
```

ros2_ws/src/brain/brain_client/brain_client/brain/agent.py
```diff
@@ -418,8 +418,14 @@ async def _pause(self, seconds: float, *, seen: int = 0, user_only: bool = False
             self._pause_until = 0.0
 
     def _interval(self) -> float:
+        directive = self._state.current_directive
+        overrides = directive.get_turn_intervals() if directive is not None else None
```
DavidDobas: This makes `_interval()` call into workspace code, and `_snapshot()` calls `_interval()`. The heartbeat that drives `_snapshot()` is a bare loop with no error boundary:

```python
async def _heartbeat(self) -> None:
    while True:
        self._lidar.tick()
        self._snapshot()
        await asyncio.sleep(1.0)
```

So an agent whose `get_turn_intervals()` raises, or returns something without `.idle` on a later call, takes the heartbeat task down and telemetry stops for the rest of the session. The `isinstance` check in `build_agent_instances` only covers the single call made at construction; nothing constrains later calls.

Suggestion: resolve the override once in `build_agent_instances`, where every other agent surface is already probed and where a failure becomes a broken-roster row with its message. Store the resolved floats on the agent, and have `_interval()` read those. The brain then never calls workspace code on the hot path, and it also stops rebuilding and re-validating a `TurnIntervals` on every heartbeat tick and every turn.

DavidDobas (approved):

PR #775, opened 2026-09-06
Keep service credentials out of public ROS parameters

axelpey:
The public simulator launches UniNavid with `INNATE_SERVICE_KEY`, then exposes that value as the ROS `service_key` parameter. A visitor can request `/uninavid_node/get_parameters` through the public `/ws` bridge. The same credential parameter pattern exists in the logger and training nodes.

Credentials now come directly from the local environment and are never declared as ROS parameters. The unused UniNavid `/brain/backend_config` credential subscription is removed as well, since anything sent on ROS is readable by bridge clients. Owner authentication and navigation configuration continue to use the existing environment; key changes require a node restart.

Validation: reproduced baseline disclosure with a synthetic key through the actual aiohttp `/ws` front door, real RWS server and real UniNavid node inside a dedicated networkless ROS container. After the fix, credential reads return no value while normal `forward_speed` reads work; two hardened integration tests pass. No provider requests or robot motion. Live checks only listed parameter names on both simulator deployments; no live credential value was retrieved.

This draft is the smallest independent mitigation for the confirmed ROS disclosure. It does not remove keys from process environments or close separate file/provider-error paths; those are covered by paired follow-up drafts. Production remains unchanged until an approved image is released and old sessions are retired. Do not merge or mark ready without Axel's approval.


vignesh-anand:
The credential-disclosure fix verified successfully: both disclosure tests exposed the synthetic key against the pre-patch UniNavid node, and all five checks passed against this PR using Zenoh in a network-disabled container. The additional checks covered parameter events/legacy overrides and logger/training credential configuration, with cloud clients stubbed for the latter.

A few stale references remain to clean up:

- **`ros2_ws/src/cloud/innate_uninavid/PROTOCOL.md:35` and `:46–53`** — The parameter table still lists `service_key`, and the runtime-update section says `/brain/backend_config` accepts `service_key`/`token` and enables future goals without restarting. Remove the parameter entry and replace the runtime-update instructions with environment/`.env` configuration via `INNATE_SERVICE_KEY`, followed by a node restart. Publishing a credential to the old topic would still expose it to ROS subscribers even though UniNavid no longer consumes it.
- **`ros2_ws/src/cloud/innate_training_node/agent-api.md:16–23`** — The parameter table still lists `service_key`, and the required-configuration text does not distinguish it from ROS parameters. Document `INNATE_SERVICE_KEY` separately as a required environment/`.env` setting and explain that changing it requires restarting the node.
- **`ros2_ws/src/cloud/innate_logger/innate_logger/logger_node.py:83–85`** — The fatal log and exception still say `service_key is required`. Name `INNATE_SERVICE_KEY` and direct the operator to the environment/`.env` so the message does not suggest setting the removed ROS parameter.
- **`ros2_ws/src/cloud/innate_training_node/innate_training_node/node.py:180–182`** — Likewise, `server_url and service_key are required` should distinguish the `server_url` ROS parameter from the `INNATE_SERVICE_KEY` environment setting and explain how to configure the credential.

I found no other stale credential-parameter examples in the repository search. These are documentation and diagnostic-message follow-ups; no functional blocker was found within this PR's credential-parameter scope.

vignesh-anand (approved):

PR #779, opened 2026-09-06
Onboarding for sim.innate.bot

axelpey:
Review fixes: changing missions now cancels local controller work without recording a terminal Skip or marking onboarding seen. The saved attempt remains available if stopping fails. Pickup checks cancellation before preparing another grasp attempt; the committed close/lift still finishes when it holds an object. Also fixes the missing blank line that failed CI formatting.

Validation of these fixes: all eleven webapp test files, six focused pickup state-machine cases, and the complete pre-commit checks pass. A browser fixture using the real broker and onboarding components verified mission → chooser without completion, a fresh iframe origin still showing onboarding, and explicit Skip recording completion while keeping its agent active. The fixture models backend state; no simulator motion or physical hardware was run. Merged main at a90458060, resolving only the competing imports for the environment bridge and slowdown indicator; the viewer build, three viewer tests and typecheck pass.

The simulator offers three conversational missions. The cloud header's **Change environment** action returns to the original three-card mission picker. It drains pending activation, confirms the owned IntroAgent has stopped, closes that attempt, then shows the chooser. Choosing a card starts a fresh mission, including when choosing the same environment. Opening an already-visible chooser does nothing; reloading it keeps it open. Settings still uses the same flow. Skip mission continues to reveal the interface while leaving MARS working.

The header bridge pins the parent source/origin, deduplicates requests, and waits for the picker to open before acknowledging success. Failed stops remain retryable without abandoning ownership. Reopening from Teleop routes back to Agent. The matching cloud action is in innate-cloud#110.

The mobile conversation uses one opaque dark sheet with a compact mission heading, goal and quiet Skip mission action. Completed actions are small expandable rows; running and failed actions retain their status accents. Drag snapping measures the current header, keeping the shorter mission strip aligned with its closed state.

Current validation: all eleven webapp test files and both companion broker tests pass. Focused tests cover trusted senders, all mission choices, same-environment restart, repeated and duplicate requests, chooser reload, late activation, concurrent requests, and failed stop/retry. Live physics/ROS/browser checks exercised Backrooms → chooser → fresh Backrooms at 0:00, chooser → Home with a new goal and model greeting, and Teleop → chooser. Desktop, 390px and 320px layouts were checked. Full webapp typecheck retains the same 5,183 baseline diagnostics with none added. Full mission completion, physical hardware and hosted broker allocation were not retested in this follow-up.

## Manually trigger the initial onboarding (browser console)

To test the initial three-challenge chooser, even if this browser has already completed or skipped onboarding:

1. Open the running simulator's **Agent** page and wait for it to load.
2. Open browser DevTools → **Console**. If the simulator is embedded in the cloud page, select the **simulator iframe** in the console's execution-context dropdown (not the parent cloud page).
3. Paste and run:

```js
(await import("/js/onboarding.js")).startFirstRun(true);
```

This opens the initial challenge chooser immediately, including for someone who has never tried a mission. Pick **Put it away**, **Find a way out**, or **The other side** to start. No simulator process restart or page reload is needed. If this browser already owns an active mission, this explicit replay closes that attempt and stops its IntroAgent before reopening the chooser; selecting a mission then prepares a fresh attempt. **Skip Mission** remains different: it only reveals the interface and leaves MARS running. The question-mark button opens the separate interface tour.

For a **standalone local simulator**, to test automatic onboarding on a fresh page load instead, run this when no mission is in progress:

```js
localStorage.removeItem("innate.firstMission.v1");
localStorage.removeItem("innate.onboardingSeen.v4");
location.reload();
```

Run this on the same simulator origin (hostname and port) you are testing. It clears only the two onboarding flags. For an embedded cloud session, use the first snippet: the parent page can otherwise restore its saved completion after reload.

First arrival offers three prepared missions: put away a LEGO brick, find the Backrooms exit, or cross the street. IntroAgent guides ordinary conversation using map-matched spatial memories. Only the physical mission result or explicit Skip unlocks the interface; reopening resumes the same attempt. Completed/skipped first runs persist on the stable broker origin. Help is a separate, replayable interface tour with cards placed beside desktop controls and above/below phone controls, and completing/skipping a first mission suppresses the redundant Challenges introduction.

Adds SuggestUserPrompts so IntroAgent can offer up to three contextual, clickable requests after its introduction, when the user needs help, after completing the authorized sequence, or after a failure. The agent instructions explain these uses and forbid interrupting an action or splitting an already-requested sequence for suggestions. Suggestion buttons are populated exclusively by this skill; there are no preset mission prompts or rotating composer examples. The composer uses the neutral "Message MARS" placeholder, and starting or reopening a mission clears old suggestions. Empty suggestions clear the buttons. Failed, duplicate and obsolete tool completions do not update them, and the tool's internal status cards stay out of the conversation and history replay.

Missions start without a camera gate. At the first actual pickup/navigation task-start event, the interface speaks a short invitation to switch to Main view and then reveals Main, Arm, and third-person (Top View). Greetings, memory searches and suggestion calls do not reveal cameras. Map and the full interface stay unavailable until physical success or Skip; selecting a camera is optional. Reload preserves the attempt and revealed stage without repeating the invitation. The mission panel sits at the bottom left on desktop. On mobile it becomes a compact strip inside the chat sheet, with the title, timer, goal and Skip visible even when chat is collapsed or fully expanded. It moves back to the desktop position on resize, and Skip restores the normal Challenges panel. Responsive browser checks used the real challenge and sheet components at desktop, 390px and 320px widths; focused tests cover reparenting, timer updates, resize and Skip. The LEGO starts at 90 degrees, putting its 32 mm side across the jaws rather than its 48 mm side. Cleanup suggests full pickup-and-careful-placement; throwing remains an explicit user choice.

Selected first missions use the existing Challenges panel, showing the authored scene brief, goal checklist, elapsed time and Skip. Completion stays visible as a normal challenge result; retry during the first run stays conversational. Every bundled challenge is scoped to an environment, and the panel shows only that environment's roster and saved progress. Results remain persisted by the existing challenge engine across page reloads and simulator sessions. The trusted broker's Change environment action reopens the original mission chooser. It drains late activation and confirms the owned agent stopped before closing the attempt. Choosing a mission there prepares a fresh scene. The action stays usable during guided missions. Requests are pinned to the parent source/origin, serialized, deduplicated and acknowledged once the chooser opens. The existing mission panel retains interruption/retry controls. The companion compact cloud header is in innate-cloud#110.

Adds ThrowObject for a small object already held, with synchronized gripper release and grasp checks before the committed swing. Preserves wrist roll, guides the agent to approach within short-toss range, and tells it to finish explicitly requested action sequences. Pickup retention is checked after carry. Misses remain recoverable through chat.

Rebased directly onto main at 5d8a02cf0. Retains the prerequisite onboarding/pickup commits previously carried through #721/#757; merged environment and traffic history now comes from main. Includes OS environment bridge #776; paired broker controls/completion persistence are in innate-cloud#110. The local combined preview also includes CTA change innate-cloud#107. Durable LEGO pickup is the separate frozen #780 dependency at dbe4de1ad36ecd21c3d9eea549effd6cf3b8fc0a. Shared cancellation fix #781 is stacked above #772. Foreign pickup and shared brain files are NOT copied into this PR. The final local integration at a0c4bb7d2 physically passed pickup→hold→approach→throw with the exact frozen pickup files.

Current follow-up validation: focused onboarding, suggestion-event, Help and preload tests pass; the ROS-backed suggestion input-validation test and all six first-mission Python tests pass; Ruff and diff checks pass. Live browser testing confirms that MARS generated clickable suggestions, clicking one started pickup, and the task-start invitation revealed camera controls while excluding Map. Main and Arm switching, reload without another invitation, 390px layout, model-generated retry suggestions after failure, and Skip restoring Map while the agent continues were verified. No typecheck diagnostics in changed modules; full webapp typecheck retains unrelated existing errors.

Current checkout MuJoCo delivery remains unverified. The original-orientation pickup missed. In a second trial, the brick was rotated to 90 degrees in the current world before the user's complete pickup-and-placement request: pickup reported success, but the brick fell before DropInBox and physics showed it on the floor; the physical goal did not pass. The test attempt was aborted and the brick restored at 90 degrees with the agent stopped. The authored 90-degree setup is loaded on a new simulator process; the live test applied the same pose without restarting the simulator. The older combined #780 evidence below does not establish current-source reliability.

Validation: all ten webapp test files and 15 focused first-mission/challenge-runtime Python tests pass. Actual onboarding and challenge-panel components were exercised together in an isolated browser fixture on desktop and mobile: selection, authored goals, conversational retry, reload/resume, completion, Skip, per-environment saved results and an empty scene. This panel follow-up did not rerun simulator motion; prior physical validation below is unchanged. The live local interface was also checked at desktop and phone widths for tour steps 1–3, complete-and-replay from Help, and the collapsed chat highlight. Focused tour tests cover resize, fallback targets, keyboard close and teardown. Ruff and diff checks pass, with no typecheck diagnostics in the changed JavaScript modules; focused challenge/memory/intersection physics tests pass; three ROS-backed throw tests pass; viewer environment tests, build and typecheck pass. Full webapp typecheck still has existing errors outside these changes. Live desktop/mobile chooser, conversational retry, reopen during pickup, Skip, keyboard handling and passive Help verified. All three missions passed real physical judges. A fresh first-run street pass revealed the UI automatically; completion survived a fresh iframe origin, without a second introduction. Tests caught a mistaken agent success claim: it did not unlock the interface, and normal chat correction continued the mission.

Final frozen-pickup compatibility: pickup completed in 20.62 seconds from tool start using two Astra calls, held the brick for 79.30 seconds, then a new chat request moved the base 0.258 m and threw the brick into the box. The physical judge passed; video, raw physics and usage are preserved in outputs/visual-evidence/final-frozen-summary.json and final-{pickup,throw}.mp4. This replay used a manual mission attempt; the first-run gate matrix was verified separately. The first frozen attempt exposed stale navigation after a long process pause and an unsolicited throw after Stop. That failed evidence is retained; #781 fixes remaining-request cancellation with 91 focused Gemini/native Responses checks and live cancellation/conversation verification. Agent distance/destination judgments can still need user correction. No physical robot testing or deployment. 

Rebase validation: all ten webapp test files, 20 focused Python tests (including clean Crossroads asset/physics build and grasp retry), both viewer environment tests, viewer build/typecheck, and diff checks pass. Range-diff preserves the 41 remaining commits; the only manual conflict resolution retains both wheel-rotation and latched-contact assertions. Earlier full conversational/physical mission evidence predates this rebase.

Third-person follow-up: Top View remains in the revealed onboarding camera list when Main or Arm is primary. Verified Main -> Top View with the real camera switch component in its restricted camera stage; Map remains unavailable.

Skip Mission now only exits the guided interface: it preserves the running agent, challenge, current action, and existing agent suggestions, including an activation already in flight. Picking another challenge still drains and stops the previous owned attempt. IntroAgent now explicitly encourages failed pickup and Backrooms navigation retries in speech and through SuggestUserPrompts, waiting for the user to request the retry.

Validation: focused onboarding, prompt suggestion, and preload tests passed; Python lint passed. A browser interaction through the actual onboarding and challenge-panel components verified that clicking Skip reveals the full interface with zero agent-stop and zero challenge-abort calls. The tests also cover pending activation and choosing another challenge after Skip. The new retry wording has not been exercised in a fresh model-driven failure rollout.


webapp/js/onboarding.js
```diff
@@ -0,0 +1,162 @@
...
+    // Storage denial cannot make a route remount replay this tab's mission.
+    unstoredFirstRun = saved;
+  }
+}
+
+function embeddingOrigin() {
+  if (window.parent === window || !document.referrer) return null;
+  try {
+    const url = new URL(document.referrer);
+    return ["http:", "https:"].includes(url.protocol) ? url.origin : null;
```
vignesh-anand, line 47: **[P2] Preserve the trusted parent origin across iframe self-reloads**

`document.referrer` is not a stable record of the embedding origin after the child reloads itself. I reproduced this in Chromium using the current broker and OS messaging modules served by two real local HTTP servers on different ports: initially the referrer was the broker origin; after `location.reload()` in the simulator iframe it became the simulator's own origin. `embeddingOrigin()` then expects parent messages from the simulator origin and rejects every broker poll. `sim/viewer/src/environmentBridge.ts:12–14` independently makes the same assumption.

The broker remains alive but Change environment becomes disabled once replies go stale. Publishing `completed("done")` after the reload also failed to update the broker's completion storage. This occurs even with no picker request pending, and does not recover through normal polling.

Please preserve/re-establish the actual trusted embedding origin across document reloads, with the same source/origin validation on both channels. Add an embedded self-reload regression that verifies get-controls, get-state, and completion persistence still work. This affects the paired controls in innate-inc/innate-cloud#110.

Validation used the real browser navigation and messaging modules at OS `8d3daa4` / cloud `6542ea9`; simulator backend work was modeled.

webapp/js/agent/agentOnboarding.js
```diff
@@ -0,0 +1,287 @@
...
+    if (active && !mission()) return Promise.resolve(true);
+    restarting = (async () => {
+      // Drain the old attempt before replacing its cancellation signal. Late
+      // activation acknowledgements must not start MARS behind the chooser.
+      // Changing missions is not a terminal Skip. Cancel this controller's
+      // work without marking onboarding seen or publishing broker completion.
+      // Keep the saved attempt until the stop succeeds so failure is retryable.
+      active = false;
+      clearTimeout(reconnectTimer); abort.abort(); render();
+      await Promise.allSettled([operation, activation].filter(Boolean));
```
vignesh-anand, line 219: **[P2] Fence timed-out activation before acknowledging the chooser**

Waiting for `activation` to settle does not guarantee that its backend request has finished. The shared `agentState.setDirective()` absorbs service errors and refresh failures; a ROS service timeout rejects the local promise without cancelling the server request. If the refresh still reports the previous inactive state and empty directive, the conditional stop below is skipped. Restart then forgets the attempt and acknowledges success to cloud #110, although the earlier activation can still take effect behind the chooser.

I reproduced this in Chromium with the real onboarding controller and shared agent-state adapter, mocking the ROS transport: hold the activation request, request restart, reject the activation with a service timeout, and return the old inactive state from refresh. Restart returns true with `phase: "choosing"` and sends no stop request. Applying the delayed backend activation then makes the adapter report an active brain while the chooser is visible. This is transport fault injection, not a live ROS reproduction.

Please retain ownership while activation is unresolved and establish an authoritative stop/order guarantee for that activation before acknowledging restart. Settled frontend promises and a cached inactive snapshot are insufficient.

webapp/js/agent/agentOnboarding.js
```diff
@@ -0,0 +1,287 @@
...
+      // attempt and stop its agent before offering a replacement scene.
+      if (saved?.attemptId && challenge?.active?.attempt_id === saved.attemptId) {
+        // setDirective refreshes state but absorbs service errors. Keep the
+        // attempt owned and retryable unless that refresh confirms the stop.
+        if (agentState.get().currentDirective === INTRO_AGENT_ID) {
+          await agentState.setDirective("");
+          if (agentState.get().brainActive) throw new Error("MARS has not stopped yet");
+        }
+        if (destroyed) return false;
+        session.abortChallenge(saved.attemptId);
```
vignesh-anand, line 234: **[P2] Confirm challenge abort before discarding the owned attempt**

`session.abortChallenge()` is a fire-and-forget call. Its underlying `WorldStateController.send()` silently drops commands unless the world socket is OPEN. That socket can disconnect while restart is awaiting the ROS stop, even if cloud #110 received fresh controls immediately before the click. This call can therefore send nothing, yet restart immediately replaces `saved` with `{phase: "choosing"}` and acknowledges success. The server still owns the old active challenge, and the client has discarded the attempt ID needed to retry its abort. Another chooser request returns early as successful.

I reproduced the dropped-command case using the real onboarding controller and `WorldStateController.send()`, with a closed WebSocket and a matching active challenge in the backend fixture: restart returned true, persisted `choosing`, and emitted zero wire commands; the old challenge remained active. The disconnected transport/backend state was modeled.

Please retain the attempt ID and wait for authoritative challenge state confirming that this attempt has closed before clearing ownership and acknowledging the request. A disconnect or confirmation timeout should leave the operation retryable, rather than count a send attempt as a completed abort.

webapp/js/agent/agentOnboarding.js
```diff
@@ -0,0 +1,287 @@
...
+      // acknowledgement uses the same UUID, which the server makes idempotent.
+      await waitFor(() => !!environment, "The simulator is still connecting. You can wait here or skip.");
+      if (challenge?.active?.attempt_id !== saved.attemptId && agentState.get().brainActive) {
+        await agentState.setDirective("");
+        await waitFor(() => !agentState.get().brainActive, "Waiting for MARS to stop before preparing the mission.");
+      }
+      if (!active || destroyed) return;
+      if (environment.environment?.id !== selected.environment) session.switchEnvironment(selected.environment);
+      await waitFor(() => environment.environment?.id === selected.environment && !environment.switch,
+        "This environment could not load. You can wait here or skip.", 45000);
```
vignesh-anand, line 151: **[P2] Allow recovery to the current world after a failed switch**

The backend retains `environment.switch` with `state: "failed"` after a switch error. Consider starting in Home/apartment, selecting Backrooms, having that switch fail, then reopening the chooser through cloud #110 and selecting Home. The current environment already matches Home, so line 149 sends no switch command, but this wait requires `!environment.switch`. The retained Backrooms failure keeps that condition false indefinitely. Each 45-second timeout retries the same path without ever calling `startChallenge()`.

I reproduced this with the actual onboarding controller and the backend's failure-shaped environment roster: the only switch request was for Backrooms, no challenge start was sent, and the Home mission remained in `starting`. Existing successful-switch behavior does not cover this recovery path. Also, simply sending a same-world switch is insufficient with the current backend: `switch_environment()` returns early when the requested world is already current.

Please distinguish a switch still in progress from a retained failed-switch result, or coordinate explicit clearing of the failure with the backend. Selecting the already-loaded world after another world fails should be able to prepare its mission while still waiting for any genuinely pending switch.

webapp/js/agent/agentOnboarding.js
```diff
@@ -0,0 +1,287 @@
...
+      await Promise.allSettled([operation, activation].filter(Boolean));
+      if (destroyed) return false;
+      // Picking another mission is distinct from Skip: close this exact
+      // attempt and stop its agent before offering a replacement scene.
+      if (saved?.attemptId && challenge?.active?.attempt_id === saved.attemptId) {
+        // setDirective refreshes state but absorbs service errors. Keep the
+        // attempt owned and retryable unless that refresh confirms the stop.
+        if (agentState.get().currentDirective === INTRO_AGENT_ID) {
+          await agentState.setDirective("");
+          if (agentState.get().brainActive) throw new Error("MARS has not stopped yet");
```
vignesh-anand, line 231: **[P2] Wait for the owned skill to finish before allowing scene replacement**

`brainActive === false` confirms brain deactivation, but does not confirm that its running skill or robot motion has finished. The backend lifecycle sets the inactive flag and calls `PrimitiveRunner.interrupt_for_deactivation()`. That runner clears its local running slot and calls `cancel_goal_async()` without awaiting the action's terminal result. Pickup intentionally finishes committed motions and can still execute arm cleanup after cancellation is requested.

Consequently, this check can pass and restart can acknowledge the chooser to cloud #110 while the old skill is still executing. Selecting a replacement mission can then call `startChallenge()` and reset the scene before that work finishes. The same concern applies to `connectMission()`'s inactive-brain check before switching/resetting a world. The simulator driver's explicit reset handler clears trajectories, but challenge initialization resets the world directly; the world-epoch handling does not provide an equivalent skill/motion drain.

Please gate scene replacement on termination of the owned skill and completion of its remaining motion, while preserving pickup's intentionally committed close/lift sequence. An inactive brain flag alone does not establish that boundary.

This finding is source-traced, not reproduced against live ROS/physics. Backend runtime validation was blocked by missing generated `brain_messages`; a cancellation-during-pickup integration test should verify the ordering.

webapp/js/onboarding.js
```diff
@@ -0,0 +1,162 @@
...
+  try {
+    const saved = JSON.parse(localStorage.getItem(FIRST_RUN_KEY) || "null");
+    if (!saved || typeof saved !== "object") return inheritedCompletion;
+    if (saved.phase === "choosing") return {phase:"choosing"};
+    if (["done", "skipped"].includes(saved.phase)) return saved;
+    if (!["starting", "playing"].includes(saved.phase)
+      || !["put_it_away", "way_out", "other_side"].includes(saved.id)
+      || typeof saved.attemptId !== "string"
+      || !/^[0-9a-f]{8}(-[0-9a-f]{4}){3}-[0-9a-f]{12}$/i.test(saved.attemptId)
+      || !Number.isFinite(saved.startedAt) || saved.startedAt <= 0) return null;
```
vignesh-anand, line 100: **[P3] Preserve inherited completion when an invalid local record cannot be replaced**

When the broker in cloud #110 replies with `done` or `skipped`, initialization correctly keeps an in-memory `inheritedCompletion` if the localStorage write fails. However, if storage already contains a JSON object that fails this validation, `readFirstRun()` returns `null` instead of that fallback. The visitor is then shown onboarding again despite having received a valid terminal completion.

I reproduced this in Chromium using the real broker and OS modules: seed `innate.firstMission.v1` with `{"phase":"playing","attemptId":"broken"}`, make subsequent writes throw `QuotaExceededError`, and let the broker return `done`. The reply arrives successfully, but `readFirstRun()` remains null and `shouldAutoStartOnboarding()` returns true. Empty storage with denied writes, and malformed storage with successful writes, both work; this requires the combination of an invalid record and a failed replacement write.

Please return `inheritedCompletion` when the stored record fails validation, or use the existing in-memory save fallback when inheriting completion. Add coverage for this combined case. This is a low-priority recovery edge case; storage failure was injected, not caused by filling a real browser quota.

webapp/js/onboarding.js
```diff
@@ -0,0 +1,162 @@
...
+  const origin = embeddingOrigin();
+  if (!origin) return Promise.resolve();
+  completionReady = new Promise(resolve => {
+    const requestId = crypto.randomUUID();
+    const finish = () => {
+      clearTimeout(timer);
+      window.removeEventListener("message", receive);
+      const saved = readFirstRun();
+      if (saved?.phase) publishFirstRunCompletion(saved.phase);
+      resolve(undefined);
```
vignesh-anand, line 71: **Non-blocking resilience suggestion: reconcile completion replies after the bootstrap deadline**

The 1.5-second bound reasonably lets startup proceed if the parent does not respond. However, `finish()` also removes the message listener, and the resolved `completionReady` promise is retained permanently. If cloud #110's valid completion reply arrives later, it is ignored, and later initialization calls cannot retry. A returning visitor can therefore see onboarding for the rest of the document lifetime even though the broker has their completion and the visitor has not selected a mission.

I reproduced this in Chromium with both real modules and an injected 1.8-second delay in parent message handling: the child received the matching `done` reply after initialization resolved, but `readFirstRun()` remained null and onboarding remained enabled. Calling initialization again sent no additional request. This demonstrates the delayed-message path; it is not evidence that normal browser messaging routinely exceeds the deadline.

Consider separating the startup deadline from later reconciliation, or permitting a subsequent retry. Any late result must preserve an explicit chooser or mission the visitor has started, and should update the mounted UI consistently. Given the documented bounded, best-effort handshake, I am treating this as an optional resilience improvement rather than a merge blocker.

PR #782, opened 2026-09-06
fix(sim): prefer checkout robot model over stale ROS install

vignesh-anand:
A sourced ROS overlay can make the local simulator load an outdated installed `mars.urdf`. When that model lacks the current wheel collision shapes, `tune_contacts` crashes with `AttributeError: 'NoneType' object has no attribute 'condim'`, which startup reports as a rendering-backend failure.

Prefer the robot model in the current checkout so it matches the simulator code. Fall back to the ROS package share for installed-only deployments.

Validation: reproduced the stale installed model failure; verified source selection and installed fallback; initialized the physics world and rendered JPEGs from both main and wrist cameras with the fix in the development checkout. Ruff and `git diff --check` pass on the PR branch.


axelpey (approved):
ok

PR #818, opened 2026-09-10
cleanup: attribution, mars_description rename, stale docs, shipped agents and skills, fork-PR CI

DavidDobas:
Cleanup pass from an open-source readiness audit. The repo is already public, so several of these are compliance corrections rather than tidying. Eight commits, each self-contained and reviewable on its own.

Net: **207 insertions, 861 deletions** across 55 files. Rebased on `main` after #819 and #821.

---

## 1. Third-party attribution

**`ACT.py` is derived from LeRobot** and said so only in a docstring 700 lines down, under an Innate copyright header. Attribution is now in the file header with the paper reference.

**Added a root `NOTICE`**, which Apache-2.0 §4 expects and the repo did not have. It collects the four third-party works we ship: LeRobot (Apache-2.0), the vendored KDL parser (BSD-3), NVIDIA's disparity sample (Apache-2.0, notice already correctly retained in that file), and three.js (MIT).

**The two vendored three.js helpers** now carry the MIT notice the minified bundle beside them already had.

**`mars_sim_driver` declared `<license>Proprietary</license>`** while its nineteen siblings and the root LICENSE say Apache-2.0 — and it is the package the no-hardware quickstart runs on, so the declaration blocked redistributing the exact thing the README invites people to try. Treated as a copy-paste default. **Flag if that is wrong**; it is the one change here that is a decision rather than a correction.

**`mars_arm/urdf.py` carried an Innate Apache-2.0 header on someone else's BSD code.** It is vendored from `kdl_parser_py` (Copyright 2008 Willow Garage, authors Jonathan Bohren and Jackie Kay); the upstream typo `recurslively` came with it.

The first attempt deleted it and moved `ik.py` onto the real package, since `package.xml` already declared the `exec_depend`. **That was wrong and is reverted** — see [§6](#6-review-response) for why. The file is back with a real BSD-3-Clause header naming the upstream project and authors, a line explaining why it is in-tree, and an entry in `NOTICE`. The unsatisfiable `exec_depend` is dropped.

**Asset credits.** `sim/README.md` credited one of six sources, and the paragraph it did carry was a byte-identical copy of `ATTRIBUTION.md`'s, free to drift. It now names every author once and points at `ATTRIBUTION.md` as the canonical per-asset list. The dog was added to that list. Nothing linked that file from prose before — the only reference in the repo was the asset-image hash input list.

**The root README gained a License section**: Apache-2.0 for the source, `NOTICE` for third-party code, and a note that the sim assets are CC BY 4.0 and fetched at build time, which are different terms a downstream user should not have to infer. It absorbs the Apache-2.0 sentence #821 had put under Contributing rather than stating the license twice.

## 2. `mars_sim` renamed to `mars_description`

`mars_sim` contains no simulation. It is the robot description — `mars.urdf`, `arm.srdf`, nine STL meshes — and `mars_arm`, `mars_bringup` and `mars_sim_driver` all `exec_depend` on it, so it is load-bearing on real hardware. Its own `package.xml` already said so.

The name is why the simulator reads as three competing implementations when it is one: a newcomer takes `mars_sim` for the simulator and `mars_sim_driver` for its driver, when it is the reverse.

44 occurrences across 21 files, each verified as the standalone package name rather than the `mars_sim_driver` / `mars_sim_bringup` prefixes that a naive search matches by accident. **Four references have to agree or the robot model silently fails to load, and all four now do:** the `package://` URIs in `mars.urdf`, `scene.ts`'s `loader.packages` key, `viz.js`'s replace target, and the proxy's `/robot/` route.

## 3. Documentation

`docs/` held four documents that between them described a codebase that no longer exists, and the README linked two of them as references:

- `INPUT_DEVICES.md` was built around a `Directive` class that appears nowhere in the tree, named directories that do not exist, invoked the input manager under the wrong executable name, and used `time.sleep` in its examples against the rule stated in three other places.
- `SYSTEM_OVERVIEW.md` was a draft whose table of contents linked five sections nobody wrote, cited a camera topic that does not exist, and shipped an unedited note-to-self about how to embed diagrams.
- `TESTING.md` named four test files, three of them deleted, and its copy-paste command errored.
- `PARAMETERS.md` had no inbound link at all.

All four deleted; docs.innate.bot covers the same subjects accurately and is now what the README and `ci/README.md` link to. `docs/assets/` stays — that is where the README's images live.

Also removed: the "For Developers" section of `scripts/update/README.md`, which documented a nonexistent workflow, a private release mirror and read-only deploy keys in a table calling this repo developers-only; CONTRIBUTING's Releases section, which promised artifacts built by a workflow whose extension is `.disabled`; and three dangling links to two dependency guides that were never written.

README: #821 rewrote both READMEs as intros while this branch was open, and it already landed the demo-link and docs.innate.bot changes this section originally made. What survives the rebase is the one link #821 could not know about -- its docs list still pointed at `docs/SYSTEM_OVERVIEW.md`, which this PR deletes, so it now points at `docs.innate.bot/software/ros2-core`. Every external link in README, `sim/README.md`, CONTRIBUTING and `ci/README.md` was checked and all resolve.

## 4. Shipped agents and skills

**The security guard prompt was a walkthrough of one particular home** — the laundry room with squares on the floor, the bedroom with the black bed, which back door is unsafe. It now asks the user which places to patrol.

It also told the model to use `turn_and_move`, which exists nowhere in the repo. Dropped. "Never use `go_to_point_in_view`" is kept: that one **is** a real built-in, and `build_tools` only offers it when `navigate_to_position` is active, which this agent grants — so it is live guidance, not leftover text. Nothing outside the file references the agent, its id or its display name.

**`SearchMemory` is now granted** so the prompt is actually executable — see [§6](#6-review-response).

**Removed `pick_socks`.** Catalogued but unrunnable: its metadata names a checkpoint with no `downloads` block and no file in the repo, while the roster reports it as still in training with zero episodes. It listed in the catalog and failed on call.

**Filled in `wave`'s guidelines.** Both `guidelines` and `guidelines_when_running` shipped empty, and those are the entire surface the model sees for a physical skill, so it was choosing from a bare id. README.md documents the field filled in, so the one shipped example contradicted the documented one.

## 5. CI and internal leftovers

**`integration-test.yml` failed on every outside contributor's first PR.** It authenticates to GCP with repository variables a fork never receives.

Worth flagging for review: the guard the other privileged workflows use, `github.repository == 'innate-inc/innate-os'`, would **not** have fixed this. On a `pull_request` event `github.repository` is always the base repo, so that condition is always true here. This checks the head repo instead and keeps the repository check for the push and dispatch paths.

**Deleted `.github/workflows/sim-images-build-cost-analysis.md`** — an internal spend memo with per-run dollar figures, a commits-per-month count and a line about the price of keeping the ugly-but-cheap path, sitting in the workflows directory where it is also not a workflow. Nothing referenced it.

## 6. Review response

Greptile raised two P1s. **Both were correct.** One of the two fixes then needed correcting in turn.

**Remembered places could not resolve.** The rewritten guard prompt described patrolling remembered places, but `NavigateToPosition` takes coordinates and resolves no names, and it was the only skill granted. `SearchMemory` is the piece that turns a description into coordinates — its own docstring says to search first and drive to the match — so it is granted now and the prompt names both steps. Note this predates the PR: the old prompt said "You can navigate from memory to the laundry room and the bedroom" with the same single skill granted, so it could not have worked either. The rewrite made the gap load-bearing rather than introducing it.

**The simulator lost its parser.** Deleting `urdf.py` moved `ik.py` onto `kdl_parser_py`, which neither image installs — the sim image applies the apt lists verbatim and runs no `rosdep`. Since `launch_sim_in_tmux.zsh:120` runs `ros2 run mars_arm ik.py`, the node would have died on import.

The first fix added `ros-humble-kdl-parser-py` to the common apt list. **That cannot work, and would have been worse than the bug.** `kdl_parser_py` was never released for ROS 2: rosdistro's humble entry for the `kdl_parser` repo releases a single package (the C++ one) while noetic's lists both, and the archive agrees — jammy/humble ships `ros-humble-kdl-parser` and its dbgsym, with no `-py` variant. Because `sim/Dockerfile` pipes the apt list through `xargs apt-get install -y`, an unlocatable package fails the `RUN` and takes the whole sim image build with it.

So the vendored copy stays. The defect was never the copy — there is nothing to depend on — it was the false header, and that is what is fixed.

---

## Checks

- `ruff check` clean repo-wide; `ruff format` clean on every file touched (the six it flags are pre-existing Markdown code fences)
- 88 root tests pass, 1 skipped
- All 7 webapp tests pass
- Pre-commit hook suites pass
- Two root tests excluded locally because MuJoCo is not installed on this machine; both fail identically on unmodified `main`

## One consequence to expect

`sim/ATTRIBUTION.md` and the `mars_description` path are both entries in `ASSETS_IMAGE_INPUT_FILES`, so this branch renames the asset image. `./innate-sim up` falls back to a local build until CI publishes `inputs-<new hash>`. It resolves on merge. The README changes do not feed any image hash.

## Deliberately not done

The soccer ball still has no attribution entry. It is under Sketchfab's Free Standard License rather than Creative Commons, and that license restricts redistributing a model on a stand-alone basis — arguably what shipping the raw `.obj`/`.glb`/`.png` in a public asset image does. That needs a decision, not an attribution line.


DavidDobas:
Both Greptile findings were valid. Fixed in 683eed9.

**Missing parser dependency (P1) — confirmed, and it would have broken the robot too, not just the sim.**

Verified the chain rather than taking it on trust:

- `kdl_parser_py` appears in no dependency list. `grep -i kdl` across all three `apt-dependencies.*.txt` and `pip-requirements.txt` returns nothing.
- `sim/Dockerfile:41-48` installs `apt-dependencies.common.txt` + `sim.txt` verbatim and never runs `rosdep`, so `mars_arm`'s `<exec_depend>` does not get resolved.
- `scripts/update/post_update.sh` has no `rosdep` call either, so the same applies on hardware.
- `scripts/launch_sim_in_tmux.zsh:120` runs `ros2 run mars_arm ik.py`, so the node really is started in sim.

Added `ros-humble-kdl-parser-py` to the common list, since both modes launch the node.

One thing this surfaced that is worth a separate look: `PyKDL` has no declared source either. `ik.py` has always imported it at module level and nothing in the apt or pip lists provides `python3-pykdl`. It has presumably been arriving transitively. `ros-humble-kdl-parser-py` depends on `python_orocos_kdl`, so this change likely makes that dependency explicit as a side effect, but the gap predates this PR.

**Remembered places cannot resolve (P1) — confirmed.**

`NavigateToPosition.execute` is `(x, y, theta_degrees=0.0, local_frame=False)`. It resolves no names, and it was the only skill granted. `SearchMemory` is the piece that turns a description into coordinates, and its own docstring already says to search first and drive to the match with `navigate_to_position(local_frame=false)`. It is now granted, and the prompt names both steps explicitly instead of implying the robot can navigate by name.

For the record, this one predates the PR. The prompt I replaced said *"You can navigate from memory to the laundry room and the bedroom"* with the same single skill granted, so it could not have worked either. The rewrite made the gap load-bearing rather than introducing it, and the fix applies to both.

theo-michel (approved):
LGTM

PR #820, opened 2026-09-10
docs(workspace): open with a hello world, and make the reference reachable

DavidDobas:
The workspace guide explained **how loading works** — packages, namespacing, hot reload, symlinked packs. That is the second thing you need. Nothing answered *how do I write one*. Meanwhile the complete API contract lives in the `innate` package's module docstring, which nothing links to and no newcomer would think to open.

Neither file was badly written. The reference is accurate and complete, and the mechanics section is good. The gap was a missing entry point.

## What changed

**A working hello world opens the guide** — 12 lines, reading the battery — followed by the three rules it demonstrates: the class *is* the registration, the docstring is what the agent reads, and a bare annotation is how you ask for a feed. Then where to go next. The loading mechanics keep every word and move below it.

**The reference is linked, not copied.** This mattered enough to call out: duplicating the feed catalogue into the guide is exactly how the four documents in `docs/` went stale. So the guide points at `help(innate)` and at docs.innate.bot for skills, agents and inputs, and stays the single source for the mechanics only.

**Three exemplars are named by size**, so nobody's first click is one of the two large vision-servoing files (915 and 485 lines):

| File | Lines | Shows |
|---|---|---|
| `innate_skills/arm/arm_rest_position.py` | 18 | The smallest useful shape |
| `innate_skills/search_memory.py` | 28 | Waiting, failing, structured payload |
| `innate_skills/turn_in_place.py` | 55 | A cancellable loop with a deadline |

**All six shipped agents now import from `innate`.** They were importing `Agent`, `InputRef` and `SkillRef` from `brain_client.agents.types` while the documented public API is `from innate import Agent`. Every example a reader copies contradicted the reference, and the public facade was used by zero shipped agents. The skills were already correct.

## Verified, not assumed

- The hello world parses, and each name it imports is in `innate.__all__`
- `Battery.percentage` and `Battery.charging` both exist on the dataclass
- All three exemplar paths exist and the line counts in the table are exact
- `ruff check` and `ruff format --check` clean across `workspace/`
- `tests/test_workspace_import_targets.py` passes (8 tests)

## One correction to the audit

It named `check_battery.py` (17 lines) as the de-facto hello world. Main deleted that skill, along with `move_straight`, in 87cca75b. So the entry point was slightly worse than the finding described, and the new example fills the same role.

## Note for whoever merges

Touches `security_guard_agent.py` (import line only). #818 touches the same file's prompt. Different lines, so it should merge cleanly either order.

## Not addressed here

Input devices still import `brain_client` internals — `InputDevice` is not part of the `innate` facade. That is a real gap in the public surface, but a bigger decision than this PR.

## Added: the README's third style is gone

The finding named three ways to import the API. Two are now one, and this commit removes the third.

README taught: import `Agent` from `brain_client.agents.types`, then list skills and inputs as **id strings** (`"innate-os/navigate_to_position"`, `"micro"`). Both agents worth copying, **Basic Navigation** (`BasicAgent`) and `DemoAgent`, use classes. The strings are legacy.

The README agent example now matches them: `from innate import Agent, InputRef, SkillRef`, skill and input imported as classes, `get_skills` / `get_inputs` annotated. I added the one caveat that keeps strings honest, since `BasicAgent`'s own docstring makes the same point: physical skills have no class, so those stay ids.

The input-device section had the same string form (`["thermometer"]`) and still said "an agent or directive". Both fixed. `directive` no longer appears anywhere in the README.

Verified: all three python blocks in the README parse, and `NavigateToPosition`, `MicroInput` and the three `innate` exports were each confirmed to exist.

**Merge note:** this touches README sections #818 does not (#818 changes the top requirements, the simulator section and the docs links). Different regions, should merge cleanly either order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,8 +374,9 @@ class ThermometerInput(InputDevice):
 An agent can then request the input by class:
 
 ```python
-from innate import InputRef
 from inputs.thermometer_input import ThermometerInput
+
+from innate import InputRef
 
 
 def get_inputs(self) -> list[InputRef]:
--- a/README.md
+++ b/README.md
@@ -374,6 +374,7 @@ class ThermometerInput(InputDevice):
 An agent can then request the input by class:
 
 ```python
+from innate import InputRef
 from inputs.thermometer_input import ThermometerInput
 
 
--- a/README.md
+++ b/README.md
@@ -233,29 +233,36 @@ Here is an example of a simple agent to navigate:
 A minimal agent file, saved as `workspace/custom_agents/navigate_agent.py`:
 
 ```python
-from brain_client.agents.types import Agent
+from innate_skills.navigate_to_position import NavigateToPosition
+from inputs.micro_input import MicroInput
+
+from innate import Agent, InputRef, SkillRef
 
 
 class NavigateAgent(Agent):
     """An agent that can navigate to requested positions."""
 
     @property
-    def id(self):
+    def id(self) -> str:
         return "navigate_agent"
 
     @property
-    def display_name(self):
+    def display_name(self) -> str:
         return "Navigate"
 
-    def get_skills(self):
-        return ["innate-os/navigate_to_position"]
+    def get_skills(self) -> list[SkillRef]:
+        return [NavigateToPosition]
 
-    def get_inputs(self):
-        return ["micro"]
+    def get_inputs(self) -> list[InputRef]:
+        return [MicroInput]
 
-    def get_prompt(self):
+    def get_prompt(self) -> str:
         return "You are a helpful robot. When asked, navigate to the requested location using the navigate_to_position skill."
 ```
+
+List skills and inputs as the classes themselves, so your editor catches a typo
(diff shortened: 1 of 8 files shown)


DavidDobas:
Greptile's finding was correct. Fixed in the latest commit.

The standalone input snippet annotated `get_inputs()` with `InputRef` while importing only `ThermometerInput`, so anyone copying that block on its own would hit a `NameError` at definition time. Added `from innate import InputRef` to the block.

Took the suggestion as given, with the two imports ordered so the block matches the shipped agents (`innate` after the workspace packages, as in `basic_agent.py`).

Re-checked afterwards: all three Python blocks in the README parse, and every name they import is either exported from `innate` or defined in the file the snippet names.

DavidDobas:
Correction to my previous comment: when I wrote it the imports were *not* ordered like the shipped agents, they were the other way round. Fixed in b653752 — workspace package first, then `innate` after a blank line, matching `basic_agent.py`. All three README blocks still parse.

axelpey (approved):

PR #822, opened 2026-09-11
feat(brain): an OpenAI-compatible provider behind the seam — Nemotron, NIM, vLLM, Ollama

theo-michel:
Stacked on #809 (the provider seam). Merge that first; this is one more package under `brain/llm/`.

## Why

NVIDIA wants the robot's brain on their own models. Behind the `Conversation` seam that is a
Chat Completions provider pointed at whatever answers it at `OPENAI_COMPAT_BASE_URL` — NVIDIA's
hosted NIM API (`https://integrate.api.nvidia.com/v1`), a NIM or vLLM container on the LAN,
Ollama. It is the provider for models Innate does not host, so it never goes through the proxy.

**Chat Completions, not Responses, on purpose.** It is the one request shape every
OpenAI-compatible server answers; Responses (what the `openai` provider uses, for reasoning
items) is not. The trade is thoughts: a server only reports them when it runs a reasoning
parser (`reasoning_content` on vLLM/NIM, `reasoning` elsewhere), so reasoning models should run
with their parser or with thinking off.

## What is in it

```
llm/openai_compat/__init__.py   build(): endpoint from env, thinking level -> request fields
llm/openai_compat/transport.py  Endpoint (base URL + optional bearer key), one httpx client, SSE
llm/openai_compat/wire.py       neutral types -> chat.completions messages / tools
llm/openai_compat/context.py    OpenAICompatConversation: stream assembly, history, pruning
```

- Text deltas go to the voice as they land; tool calls arrive fragmented by `index` and are
  assembled at stream end. The server's call ids are stored on the assistant message and quoted
  back on `role: tool` messages (a server that omits ids gets one minted).
- A stream that stops before `finish_reason`, or ends on anything but `stop`/`tool_calls`,
  fails the turn instead of committing a truncated call — the rule #809 applied to
  `response.incomplete`.
- Same append-only history and chunked compaction as the other two providers, so a server's
  prefix cache stays warm; wrist frames are latest-only exactly as before.
- `brain_thinking_level` in this provider's vocabulary: `off` → `chat_template_kwargs.enable_thinking=false`
  (the vLLM/NIM switch for models that reason by default: Nemotron 3, Qwen3), `low|medium|high` →
  `reasoning_effort`, `""` → server default.
- `parse_arguments` moves to `llm/types.py`; the `openai` provider now shares it.
- Sim launcher: `resolve_brain_backend` mirrors the precedence (the URL decides alone), the
  status pane labels it, the wizard leaves a hand-configured endpoint alone. Settings page gets
  the option; `.env.template` documents the three lines.

## Configuration

```
BRAIN_BACKEND=openai_compat
BRAIN_MODEL=nvidia/nemotron-3-nano-omni-30b-a3b-reasoning
OPENAI_COMPAT_BASE_URL=https://integrate.api.nvidia.com/v1
OPENAI_COMPAT_API_KEY=nvapi-…          # omit for an open LAN server
```
plus `brain_thinking_level: "off"` in settings for Nemotron 3 (it reasons by default and the
traces are long).

## Verified

- ruff clean; basedpyright 0 errors on `brain/llm`, `core/config.py`, the launcher files.
- 297 pass in the real-ROS brain_client bucket (sim image); launcher `tests/` pass except one
  intersection-pack test that fails identically on the base commit.
- A two-turn exchange through the real httpx transport against a fake Chat Completions server:
  bearer auth, data-URI frames, a tool call split across three deltas, `tool_call_id` round
  trip on the next request, cached-token usage, wrist-frame masking, `enable_thinking=false`
  on the body.

## Not yet verified — needs someone with an NVIDIA key

- No live run against NIM. The wire is the documented one; server quirks are where this breaks.
- Behavioural, not plumbing: the 0-1000 image-coordinate convention behind
  `go_to_point_in_view` and the same-turn answer+call discipline were measured on Gemini. Run
  the reaction-latency probes on Nemotron before trusting it on a robot.

Development tests (12, pure fake-transport) were written and deleted per the repo rule; say
the word and they go into `test_local_brain.py`.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -17,8 +17,15 @@
 # GEMINI_API_KEY=
 # OPENAI_API_KEY=
 
-# BRAIN_BACKEND picks the provider: gemini (default) or openai.
+# BRAIN_BACKEND picks the provider: gemini (default), openai, or openai_compat.
 # BRAIN_BACKEND=gemini
+#
+# openai_compat is any server speaking /v1/chat/completions — NVIDIA's hosted
+# NIM API, a NIM or vLLM container on your network, Ollama. It never goes
+# through the proxy. Set BRAIN_MODEL to that server's model id, and a thinking
+# level of "off" for models that reason by default (Nemotron 3).
+# OPENAI_COMPAT_BASE_URL=https://integrate.api.nvidia.com/v1
+# OPENAI_COMPAT_API_KEY=
 
 # BRAIN_MODEL overrides the brain's model (default gemini-3.6-flash). It must
 # belong to BRAIN_BACKEND. GEMINI_MODEL is the name this had before the brain
--- a/ros2_ws/src/brain/brain_client/brain_client/README.md
+++ b/ros2_ws/src/brain/brain_client/brain_client/README.md
@@ -6,7 +6,7 @@ code, start from what it *does*:
 | Folder | What lives here |
 |---|---|
 | `nodes/` | The runnable ROS entry points (the only files with `main()`). Thin composition roots — they build collaborators, wire them, and spin. No behaviour. |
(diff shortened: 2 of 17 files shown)


theo-michel:
Superseded by #823: the same capability as a standalone exit under the existing transport seam on main, with the endpoint and model as robot settings. Direction chosen by Theo so it does not wait on #809. If #809 lands first, #823's openai_compat.py becomes the body of an llm/openai_compat provider unchanged.

PR #823, opened 2026-09-11
feat(brain): think with any OpenAI-compatible server — a setting, not a fork

theo-michel:
## Why

NVIDIA wants the robot's brain on Nemotron; people want it on a vLLM or Ollama running on a
computer in the same room. Until now the only way out was Gemini. This adds a third exit under
the existing transport seam and leaves the agent loop, `GeminiContext`, the pruning policy, the
memory search and the brain monitor untouched.

## How to use it

Settings page → **AI models**, or `config/settings.yaml`:

| Setting | Value |
|---|---|
| OpenAI-compatible endpoint (`llm_base_url`) | `http://192.168.1.20:8000/v1` (vLLM), `http://192.168.1.20:11434/v1` (Ollama), `https://integrate.api.nvidia.com/v1` (NVIDIA hosted) |
| Brain model (`gemini_model`) | that server's model id, e.g. `qwen3-vl:8b`, `nvidia/nemotron-3-nano-omni-30b-a3b-reasoning` |
| Endpoint request extras (`llm_extra_body`) | optional JSON, e.g. `{"chat_template_kwargs": {"enable_thinking": false}}` for Nemotron 3 / Qwen3 |

Only the key is a secret: `LLM_API_KEY` in `.env` (a LAN server usually needs none). `LLM_BASE_URL`
in `.env` is the launch default; settings.yaml wins. Empty endpoint = Gemini, exactly as before.
Restart to apply, like the other brain settings.

## What is in it

- `brain/openai_compat.py` — pure translation. A native Gemini request goes out as a Chat
  Completions request (system message, data-URI frames, JSON Schema tools, `functionCall` /
  `functionResponse` as `tool_calls` / `role: tool` with the server's ids); the deltas come back
  as Gemini chunks (text per delta so the voice starts on the first sentence, reasoning as thought
  parts, tool calls assembled from fragments at the end, usage + finish reason on the last chunk).
  A stream that stops before `finish_reason`, or ends on `length` / `content_filter`, fails the
  turn instead of committing a truncated call.
- `brain/transport.py` — `Endpoint` (URL + key + extras), `Backend.OPENAI_COMPAT`, and
  `pick_transport` / `pick_rest` preferring the endpoint outright (it is the operator's own server
  and never goes through the proxy). The REST half serves memory search's blocking generate
  (`responseSchema` → `response_format: json_schema`); the context-cache and Files paths answer
  501, which those callers already latch off, so frames ride inline.
- Two new `brain_client_node` parameters, wired through `BrainConfig`, both launch files, the
  Settings catalog, `settings.yaml.template` and `.env.template`.
- Sim launcher: recognises the endpoint (from `.env` or by reading settings.yaml's
  `llm_base_url` line — it has no YAML parser), labels it in the status pane, and the setup
  wizard leaves a configured endpoint alone.

Thoughts: a server only reports them when it runs a reasoning parser (`reasoning_content` on
vLLM/NIM). Gemini's thinking level is not translated — it has no universal counterpart — so a
reasoning model's switch rides `llm_extra_body`.

## Verified

- ruff clean; basedpyright 0 new errors on `brain/`, `core/config.py`, the launcher files (one
  pre-existing `runtime.py` tuple error, untouched).
- 293 pass in the real-ROS brain_client bucket (sim image); the patched node imports there.
- Through the real httpx transport against a fake Chat Completions server: two turns with a tool
  call split across three deltas, `tool_call_id` round trip on the next request, cached-token
  usage on the monitor's counters, wrist-frame masking, extra body on the wire, bearer auth; a
  memory-search-style blocking call with `json_schema`; cache and upload → 501.

## Not yet verified — needs a real server

No live run: no NVIDIA key or LAN GPU on hand. The wire is the documented one; server quirks are
where this breaks. Behavioural, not plumbing: the 0-1000 image-coordinate convention behind
`go_to_point_in_view` and the same-turn answer+call discipline were measured on Gemini only.

Relation to #809 / #822: #809 is a provider seam with an OpenAI Responses backend; #822 was this
same capability stacked on it. This PR is the standalone version on main, chosen so it does not
wait on #809; if #809 lands first, `openai_compat.py` becomes the body of an `llm/openai_compat`
provider with no logic change.

Development tests (10, pure) were written and deleted per the repo rule; say the word and they
go into `test_local_brain.py`.


theo-michel:
Superseded by #824, which does the simpler thing: instead of translating native Gemini bodies to Chat Completions under the transport, the brain now speaks Chat Completions only and Gemini becomes the default endpoint through Google's OpenAI-compatible layer. That deletes this translator along with the context-cache and Files tiers, and the measured gates (discipline 5/5, latency at parity, prefix cache 8101/13103 vs native 8100/13086) all hold. Leaving this open until #824 is reviewed, in case NVIDIA needs something sooner.

theo-michel:
Superseded by #836, which merged `innate_llm` into main — that library is the shipped way to talk to any LLM, so this transport has no consumer left.

PR #828, opened 2026-09-11
Fix tutorial memories leaking into the next map

axelpey:
Skipping or finishing onboarding could save a tutorial image into the next map: a buffered frame survived the switch, and AMCL could keep publishing confident poses from its previous filter.

Discard the buffered capture on map changes, and clean up AMCL before announcing the new map. Reactivation requires a new initial pose under the existing `always_reset_initial_pose` configuration. If AMCL cannot reset, the old map is retained and the activation loop restores navigation before returning the failure; restoration failures are reported too.

Validation: 144 spatial-memory/navigation tests pass. Coverage stays small: one memory regression and a parameterized reset-order/failure test, wired into CI. An isolated real-AMCL check confirmed that poses stop across cleanup/reconfiguration and resume only after reseeding. Ruff and diff checks pass. Full onboarding UI not retested.

### DIFF
--- a/ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
+++ b/ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
@@ -938,13 +938,13 @@ class ModeManager(Node):
         self.get_logger().error(f"Failed to load map after {max_retries} attempts")
         return False
 
-    def _efficient_map_switch(self) -> tuple[bool, str]:
+    def _efficient_map_switch(self, requested_map: str) -> tuple[bool, str]:
         """
-        Efficiently switch maps by transitioning bt_navigator down, loading map, then bringing all nodes up.
+        Switch maps with navigation stopped and the previous AMCL filter discarded.
 
         Algorithm:
         1. Transition bt_navigator to inactive
-        2. Load new map on map_server
+        2. Reset AMCL, then expose and load the new map
         3. Transition all nodes to active
         """
         nodes = modes_nodes.get("navigation", [])
@@ -959,6 +959,16 @@ class ModeManager(Node):
         if not success:
             failures.append("bt_navigator")
             self.get_logger().warning("Failed to transition bt_navigator down")
+
+        # Stop and clear AMCL before publishing the new map identity. Otherwise
+        # its old filter can keep supplying confident poses during the switch.
+        # Cleanup + configure honors always_reset_initial_pose; deactivation
+        # alone retains the filter. The activation loop below brings it back.
+        if not transition_node(
+            self._service_clients, self.get_logger(), "navigation_amcl", State.PRIMARY_STATE_UNCONFIGURED
+        ):
+            return False, "Could not reset AMCL before switching maps"
+        self.current_map = requested_map
 
         # Step 2: Load new map
         self.get_logger().info("Step 2: Loading new map")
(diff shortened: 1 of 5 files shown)


ros2_ws/src/mars_bot/mars_nav/test/test_map_switch.py
```diff
@@ -0,0 +1,39 @@
...
+from lifecycle_msgs.msg import State
+
+from mars_nav import mode_manager as module
+
+
+@pytest.mark.parametrize("reset_ok, restore_ok", [(True, True), (False, True), (False, False)])
+def test_reset_amcl_before_announcing_new_map(monkeypatch, reset_ok, restore_ok):
+    manager = SimpleNamespace(
+        current_map="A.yaml", _service_clients={}, get_logger=Mock(), _load_map_on_server=Mock(return_value=True)
+    )
```
DavidDobas, line 16: Nit. Standing `self` up as a `SimpleNamespace` and calling the unbound method couples the test to the internals of `_efficient_map_switch` rather than to its contract. Any new `self.` access added inside the method fails here with an `AttributeError` that has nothing to do with the invariant being pinned. It fails noisily rather than silently, so it is liveable.

ros2_ws/src/mars_bot/mars_nav/test/test_map_switch.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def transition(clients, logger, node, state, **kwargs):
+        calls.append((node, state, manager.current_map))
+        if state == State.PRIMARY_STATE_UNCONFIGURED:
+            return reset_ok
+        return restore_ok if node == "navigation_amcl" and state == State.PRIMARY_STATE_ACTIVE else True
+
+    monkeypatch.setattr(module, "transition_node", transition)
+    success, message = module.ModeManager._efficient_map_switch(manager, "B.yaml")
+    assert success == reset_ok
```
DavidDobas: This reads as a contract but only holds for the three combinations sampled above. With `reset_ok=True, restore_ok=False` the activation loop appends `navigation_amcl` to `failures`, so `success` is `False` while `reset_ok` is `True`.

Either make it `success == (reset_ok and restore_ok)` and add the fourth case, or drop this line and keep the ordering assertions, which are the actual lesson.

ros2_ws/src/mars_bot/mars_nav/mars_nav/mode_manager.py
```diff
@@ -960,11 +960,24 @@ def _efficient_map_switch(self) -> tuple[bool, str]:
...
+        # Cleanup + configure honors always_reset_initial_pose; deactivation
+        # alone retains the filter. The activation loop below brings it back.
+        if not transition_node(
+            self._service_clients, self.get_logger(), "navigation_amcl", State.PRIMARY_STATE_UNCONFIGURED
+        ):
+            # Reset may have partially transitioned AMCL. Keep the old map,
+            # but always run the activation loop to restore navigation.
+            failures.append("navigation_amcl_reset")
+        else:
+            self.current_map = requested_map
```
DavidDobas, line 974: Observation for a follow-up, not for this PR. If the reset succeeds here and a later step fails, `_load_map_on_server` has already put the new map on the server while `change_map_callback` rolls `current_map` back to the previous one. `/current_map` then advertises a map the server is not holding, which is the same family as the bug being fixed here.

Pre-existing, and this PR narrows the window rather than widening it.

DavidDobas (commented):
The fix reads right, and the ordering invariant is worth pinning: `current_map` must not advance until AMCL is unconfigured, and that assignment used to live in `change_map_callback`, so a refactor that hoists it back reintroduces the bug silently.

Three non-blocking notes below.

DavidDobas (approved):

PR #848, opened 2026-09-19
MARS on LeRobot: bridge, plugin, dataset converter, publish to Hugging Face

DavidDobas:
## What this does

Makes MARS usable from [LeRobot](https://github.com/huggingface/lerobot) and gets MARS data into LeRobotDataset v3, without replacing anything in innate-os. The app's teleop and recorder stay the primary workflow; this adds a second door into the same data and a way for any lerobot policy to drive the robot.

About 40% of the line count is the generated `lerobot/uv.lock`; see [Where the lines are](#where-the-lines-are) at the end.

## The four parts

**1. Bridge in `manipulation_server`** (`manipulation/lerobot_bridge.py`, ROS-free, plus wiring in `manipulation_server.py`)
- ZMQ: PULL `:5555` for actions and heartbeats, PUB `:5556` for state and JPEG observations. One frame per message so clients conflate to the newest.
- No new node. On by default (`lerobot_bridge.*` params), idle until a client heartbeats, ignores commands while `manipulation_server` itself executes a behavior, stops the base 0.5 s after a commanding client goes silent. Logs `LeRobot client connected` / `idle`.
- Forwards a head-tilt command and reports the measured tilt.
- Needs `pyzmq`, added to `pip-requirements.txt`.
- Camera subscriptions open with the first LeRobot client, not at boot. `lerobot_bridge.bind_address` can keep the bridge to the robot itself. No login, like rosbridge; speed and joint limits stay in the drivers, and the bridge itself holds gripper targets to the safe range. Follow-up on unsubscribing when idle: #849.

**2. `lerobot/` plugin, `lerobot_robot_mars`** (own uv project, Python 3.12, lerobot 0.6)
- `--robot.type=mars`: the bridge client. `--teleop.type=mars_passthrough`: reads back what the app or leader arm commanded, so `lerobot-record` works while the phone app drives.
- Robot address from `MARS_HOST`. Head angle set at connect and held, written to the dataset's `meta/mars.json`.
- `mars2lerobot`: converts a skill recorded by the app to LeRobotDataset v3, incrementally, optionally pushing to the Hub. What was exported is recorded in `data/lerobot_export.json`, a file only the converter writes, and only after the dataset is finalized; a converted copy that does not hold exactly those episodes is rebuilt.
- `mars-push`: uploads a dataset recorded with `lerobot-record`; lerobot has no command for that.
- Head angle for replay, resume and rollout comes from the dataset behind the run, so nobody sets it by hand.
- `schema.py` is the single definition of feature names shared by the live client and the converter: state 6 joints, action 6 joints + base `x.vel`/`theta.vel`, cameras `head` and `wrist` in RGB.
- Base install is lean (dataset conversion only); the `workstation` group, on by default with uv, adds record, viewer, and training. Opt-in extras for diffusion, SmolVLA, π.

**3. Publish to Hugging Face in the webapp**
- Settings: **Keys** is now its own top-level page after Agent, with the model providers' keys (moved from Agent) and a Hugging Face token.
- Datasets: **Publish to Hugging Face** on training datasets. The dialog handles a missing token, a one-time install of the converter's environment on the robot (1.3 GB, CPU-only torch), account or organization, repo name, visibility, and shows progress. The job survives closing the dialog.
- Runs inside the webapp server (`proxy/hub_publish.py`, `hub_routes.py`), which already owns the keys and serves the skill files: no ROS node, no new message types. Disabled on read-only webapps.

**4. Sim**
- Container publishes 5555/5556 (port-block offsets 7 and 8), mounts `lerobot/` read-only with its environment in a named volume; image gains `pyzmq`.
- `INNATE_SIM_HARDWARE_CAMERA_RATES=1` renders the raw camera topics at the robot's 15/30 fps while something subscribes to them. Off by default; the compressed streams keep their rates either way. Measured on a Mac: raw 14.4/28.9 Hz with the flag, 9.9/6.0 without.

## Verified

- Unit tests: 12 plugin, 6 bridge, 2 new in the webapp proxy suite (41 there in total). The plugin tests load the real bridge module, so both halves of the wire protocol are pinned together. Ruff clean.
- Sim, end to end: observations at 30 Hz with both cameras, arm and base follow commands, passthrough echo of app commands, watchdog, `lerobot-record` producing a valid dataset, `lerobot-replay` moving the arm, head hold and sidecar, a real conversion through the publish job, every dialog state in headless Chrome.
- On a robot (David): teleop passthrough, recording 50 episodes, replay, ACT trained with `lerobot-train`, rollout reaching toward the object. Dataset: `innate-inc/mars-pick-tv`.
- On a robot (David), the publish path: `pyzmq` arriving through `innate update`, the converter's environment installing on the Jetson from the dialog, converting an app-recorded skill whose images the background encoder had already stripped (the MP4-decode path), and a successful upload from the webapp with a real token.

## Known limits

- Only `manipulation_server`'s own behaviors block a LeRobot client. The base is arbitrated by the velocity mux, where app teleop wins; the arm has no mux, so a client driving the arm while a code skill or the agent moves it will fight them. Same trust model as rosbridge.
- The head camera is 15 fps under a 30 Hz stream, so about every other head frame repeats. In the sim, cameras run slower still unless `INNATE_SIM_HARDWARE_CAMERA_RATES=1` is set, which needs a GPU on the host.
- Publishing only adds episodes. Deleting or relabeling after a publish does not propagate to the Hub.
- lerobot-side inference on the Jetson would be CPU-only (no CUDA torch wheels for Python 3.12 on JetPack 6). Production inference stays on the TensorRT path; an exporter from lerobot ACT checkpoints to that path is the natural follow-up.

## Try it

`lerobot/README.md` is the step-by-step guide: requirements, setup, record, replay, train, run, publish.

## Where the lines are

About 6,000 added lines. Roughly 2,400 of them are code to review.

| Part | Lines | What it is |
|---|---|---|
| `lerobot/uv.lock` | 2,476 | Generated. 133 packages pulled in by lerobot, torch and the Hugging Face stack, with a checksum per platform. Limited to the four platforms torch ships wheels for, so nothing that could install before was dropped; down from 3,222. Marked `linguist-generated`, so GitHub collapses it. Kept so every install, including the robot's, gets the tested set. |
| LeRobot plugin | 1,279 | Robot client, passthrough teleop, wire protocol, schema, converter (419, the largest: two image sources, incremental export, locking against the recorder), dataset metadata, `mars-push`. |
| Webapp publish | 782 | Publish dialog (320), background job (227), routes (116), Keys page and styles. |
| Tests | 664 | 20 tests, see below. |
| Docs | 474 | `lerobot/README.md` (444), a section in the main README, `.gitattributes`. |
| Bridge on the robot | 363 | `lerobot_bridge.py` (236, ROS-free) and its wiring in `manipulation_server.py`. |
| Sim | 48 | Ports, mount, camera-rate flag. |

### Why these tests and no others

The branch had 40 tests while it was being built. Under the project rule, 20 stay: each guards a contract between two codebases, physical safety, security, or data loss. Feature behaviour and one-off regression tests were deleted.

| Tests | Count | What breaks silently without them |
|---|---|---|
| Wire protocol, plugin against the real bridge module | 5 | The two sides share no code and run on different Pythons (3.12 and the robot's 3.10). Constants match, an observation survives the round trip, an action reaches the robot, the client sends nothing while an operator teleoperates, the recorded action is the operator's command. |
| Bridge safety, in CI | 6 | Gripper targets are held to the range that cannot overcurrent the servo, the watchdog stops the base, malformed and non-finite actions are dropped, commands are ignored during a skill without tripping the watchdog, a skill starting mid-drive gets one stop, nothing is published with no client connected. |
| Converter | 5 | Converted data matches the live schema (two producers, one format). A rerun appends only new episodes even after the recorder rewrites its metadata file. A lost local copy is rebuilt in full instead of overwriting the Hub dataset with only the new episodes. A copy left by a run killed before finalize is rebuilt, not resumed. A folder the converter did not write is never cleared. |
| Plugin discovery | 2 | lerobot finds the plugin by naming conventions alone; a rename breaks it with no error. |
| Webapp security | 2 | Publish refuses paths outside the skills root and unmet preconditions; cross-site posts are refused. |


ros2_ws/src/brain/manipulation/manipulation/lerobot_bridge.py
```diff
@@ -0,0 +1,234 @@
...
+    def _apply(self, payload: dict, now: float) -> None:
+        joints = _floats(payload, JOINT_KEYS, required=True)
+        base = _floats(payload, BASE_KEYS, required=False)
+        if joints is None or base is None:
+            self._log(f"LeRobot bridge ignored a malformed action: {payload}")
+            return
+        self._last_action = now
+        self._base_stopped = False
+        if self.commands_blocked:
+            return
```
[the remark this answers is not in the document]
DavidDobas: Fixed in 97230e63. A blocked command now returns before it touches `_last_action` or `_base_stopped`, so a connected client can no longer arm the watchdog while a behavior runs. The one remaining stop is deliberate: if a behavior starts while the client was driving the base, the bridge sends a single zero on the next tick, before the behavior has moved, and none after. Without it the base would keep the client's last velocity until the velocity mux declares the source stale. Both cases are pinned in `test_lerobot_bridge.py` (`test_blocked_bridge_keeps_hands_off`, `test_a_behavior_starting_mid_drive_gets_one_stop_up_front`).

ros2_ws/src/brain/manipulation/manipulation/lerobot_bridge.py
```diff
@@ -0,0 +1,234 @@
...
+    def bind(self) -> None:
+        context = zmq.Context()
+        pull = context.socket(zmq.PULL)
+        pull.setsockopt(zmq.LINGER, 0)
+        pull.setsockopt(zmq.RCVHWM, 64)
+        pull.bind(f"tcp://{self._bind_address}:{self._port_actions}")
+        pub = context.socket(zmq.PUB)
+        pub.setsockopt(zmq.LINGER, 0)
+        pub.setsockopt(zmq.SNDHWM, 8)
+        pub.bind(f"tcp://{self._bind_address}:{self._port_observations}")
```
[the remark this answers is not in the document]
DavidDobas: This matches the robot's existing trust model and is intended. MARS trusts its local network: the rosbridge websocket the web app and phone app use is unauthenticated and already lets any network peer publish to every topic, including `/mars/arm/commands`, `/cmd_vel_skills` and `/mars/head/set_position`, and subscribe to both cameras. A password on these two ZMQ ports would not reduce what a network peer can do. It is also the transport LeRobot users expect: LeKiwi's host uses the same unauthenticated PUSH/PULL and PUB/SUB pair.

On plausibility limits: they live downstream, once, for every command source. `bringup.py` clamps base commands to `safety.max_speed` and `safety.max_angular_speed`, and `arm_control.cpp` clamps joint targets to the joint limits. The bridge rejects non-finite and malformed actions, ignores commands while a behavior executes, and stops the base after 0.5 s of client silence.

For people who want it narrower, ff9ede0c adds `lerobot_bridge.bind_address`: `127.0.0.1` keeps the bridge to lerobot processes on the Jetson, and `lerobot_bridge.enabled: false` closes it. Authenticating the robot's network surfaces as a whole is a separate piece of work from this PR.

webapp/proxy/hub_routes.py
```diff
@@ -0,0 +1,104 @@
...
+        return _refuse(404, "no such dataset on this robot")
+    if not hub_publish.REPO_ID_RE.match(repo_id):
+        return _refuse(400, "the repository must look like owner/name")
+    token = await asyncio.to_thread(_token)
+    if not token:
+        return _refuse(400, "save a Hugging Face token in Settings first")
+    if not hub_publish.env_ready():
+        return _refuse(409, "the LeRobot environment is not installed yet")
+    try:
+        hub_publish.start_publish(skill_dir, repo_id, private=private, include_failures=include_failures, token=token)
```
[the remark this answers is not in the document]
DavidDobas: Partly fixed in 97230e63, the rest is intended. The realistic attack here is cross-site: `request.json()` accepts any content type, so a page the operator merely visits could have posted to the robot and published with the stored token. Both POST routes now refuse a request whose `Origin` is not the robot's own web app (`_cross_site`, covered by `test_a_page_from_another_site_cannot_publish`).

What remains is that a peer on the robot's network can call these routes without a login. That is the web app's model for every write route: `POST /keys.json` and `POST /settings.json` next to these are the same, and a read-only web app registers none of them. The token is never returned by any route, the dataset path is confined to the recordings root, new repositories default to private, and one job slot means setup cannot be stacked. Adding login to the web app is out of scope for this PR.

DavidDobas:
@greptileai the three findings have replies on their threads: the watchdog bug is fixed in 97230e63, the publish routes refuse cross-site posts in 97230e63, and the bridge gained a bind address parameter in ff9ede0c. The rest is the robot's existing trusted-network model, explained on each thread. Please re-review.

ros2_ws/src/brain/manipulation/manipulation/lerobot_bridge.py
```diff
@@ -0,0 +1,236 @@
...
+        joints = _floats(payload, JOINT_KEYS, required=True)
+        base = _floats(payload, BASE_KEYS, required=False)
+        if joints is None or base is None:
+            self._log(f"LeRobot bridge ignored a malformed action: {payload}")
+            return
+        if self.commands_blocked:
+            return
+        self._last_action = now
+        self._base_stopped = False
+        self._on_arm(joints)
```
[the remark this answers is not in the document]
DavidDobas: Agreed, fixed in b58a8e69. The bridge now clamps `joint6.pos` to `GRIPPER_RANGE = (-0.6, 0.85)` before it forwards the arm command, mirroring `Manipulation.GRIPPER_OPEN` and `GRIPPER_MAX_STRENGTH`. It clamps rather than rejects: the leader arm records gripper commands slightly past open (0.90 in `innate-inc/mars-pick-tv`), and dropping those whole actions would stall a replay. Pinned by `test_gripper_target_is_held_to_the_safe_range` in the CI bridge tests, and checked end to end in the sim: a client sending -2.0 results in a commanded -0.60.

webapp/proxy/test/test_hub_routes.py
```diff
@@ -0,0 +1,50 @@
...
+
+@sync
+async def test_a_page_from_another_site_cannot_publish(tmp_path, robot):
+    body = {"dir": str(robot), "repo_id": "me/mars-s"}
+    async with serve(ROOT=make_app_root(tmp_path)) as (s, base):
+        foreign = {"Origin": "https://evil.example"}
+        assert (await s.post(base + "/hub/publish", json=body, headers=foreign)).status == 403
+        assert (await s.post(base + "/hub/setup", headers=foreign)).status == 403
+        own = await s.post(base + "/hub/publish", json=body, headers={"Origin": base})
+        assert own.status == 400 and "token" in (await own.json())["message"]
```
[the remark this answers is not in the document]
DavidDobas: This one is deliberate and follows the team's testing policy (`CLAUDE.md`, "Tests are a development tool, not a deliverable"): tests written while building a change are deleted before the PR, and only tests that guard architecture stay. The bar used here was a contract between two codebases, physical safety, security, or data loss; the PR description lists the 18 that remain and why. The removed publish tests covered feature behaviour (progress events, the happy path, account lookup, the exact install command line). Those flows were verified end to end instead: every dialog state in headless Chrome against the sim, and a real install, conversion and upload on a robot. The security-relevant publish behaviour keeps its tests: the skills-root fence, the preconditions, and the cross-site guard.

DavidDobas:
@greptileai the gripper finding is fixed in b58a8e69 and both threads have replies. Please re-review.

DavidDobas:
## Follow-up to the review of b71d113f

Two commits: 5deda35f (data fixes) and 3246ddb0 (cleanups). The review's findings held up when checked; testing its open question also surfaced a second data bug.

### Data bugs, fixed

**1. The recorder could wipe the export record (finding 1).** Confirmed in `task_manager.cpp`: `save_metadata` rewrites `dataset_metadata.json` from the snapshot taken at task activation and merges back only `dataset_type` and `video_files`. Publishing while the skill is active, then recording again or leaving the screen, dropped `lerobot_export`, and the next publish would have appended every episode a second time.
- The record now lives in `data/lerobot_export.json`, which only the converter writes. `_locked` and the `fcntl` import are gone, and nothing in the C++ changes.
- The old key in `dataset_metadata.json` is still read once as a fallback, so skills published before this change keep their history. The webapp's "published before" line reads the new file with the same fallback.

**2. A conversion killed midway could never be resumed (the question at the end of finding 4).** I built a six-episode skill, sent SIGKILL after three episodes, and reran. lerobot v3 only writes the parquet footer in `finalize()`, so the killed run left an unreadable data file while the record already listed three episodes as exported. The rerun failed for good, and with a misleading message: lerobot fell back to the Hub, got a 404, and we reported "Hugging Face refused the upload".
- Episodes are now recorded as exported only after `finalize()`.
- A converted copy is appended to only when it holds exactly the recorded episodes (`total_episodes` in `info.json` against the record). Anything else is rebuilt from every episode: a killed run, a cleared cache, a lost record.
- Rebuilding deletes the copy, so it is limited to folders the converter made: the record names that `root`, or the dataset's `meta/mars.json` says `mars2lerobot`. Pointing `--root` at, say, a dataset recorded with `lerobot-record` now fails with a clear message instead of being appended to or cleared.
- Re-tested with real kills after one and after three episodes: the rerun exits 0 and the dataset has all six episodes, every index readable.

Tests, under the data-loss rule: `test_rerun_appends_only_new_episodes` now restores the recorder's file from its snapshot between the two runs; new `test_a_copy_left_by_a_killed_run_is_rebuilt` and `test_a_folder_the_converter_did_not_write_is_never_cleared`.

### Design gaps

**Finding 2, "commands are ignored while a behavior executes".** Narrowed everywhere to what the code does: only `manipulation_server`'s own behaviors block a client. The bridge docstring, the README's safety list and its troubleshooting row say that the base is arbitrated by the mux and the arm is not, and the PR description lists it under Known limits.

**Finding 3, the sidecar write.** `recording_root` is only used for writing, so I tightened it rather than splitting read and write: the directory must be the exact name or the name plus lerobot's `_YYYYMMDD_HHMMSS` stamp (the format in `lerobot/configs/dataset.py`). `pick` no longer matches `pick_tv`. The network access in `connect()` is unchanged: it happens only when a policy's training dataset is not on disk, and failures fall back to the default angle.

### Cleanups

- **Protocol documented once**, in `lerobot_bridge.py`, now including the header fields. `wire.py`, the yaml comment and the `__init__` comment point to it.
- **Lazy bridge import** inside the function that uses it, like `act_trt`. `LeRobotBridge = None`, `LEROBOT_BRIDGE_IMPORT_ERROR` and the top-level `try` are gone; the annotation's import sits under `TYPE_CHECKING`.
- **`_start_lerobot_bridge` split** into opening the bridge, following the operator's command topics, and starting the timer. The declare-then-get pairs became `declare_parameter(...).value`.
- **One JPEG per frame.** The bridge reuses the encoded bytes while it is handed the same image object, so the 15 fps head frame is no longer encoded twice per 30 Hz tick.
- **`has_dataset(root)`** in `hub.py` replaces the six copies. `_convert()` is inlined. `Link.latest` reads one message, which is all CONFLATE keeps.
- **`sim/README.md`** port table gained the +7 and +8 rows.

### Not done, and why

- **Finding 5, CPU torch through `[tool.uv.sources]`.** Worth trying, but as its own change. A platform marker for Linux aarch64 would also give CPU-only torch to an ARM PC with an NVIDIA GPU, and it replaces the install path that was verified on a real Jetson. It should be tested there before it lands.
- **Shared config mixin.** Both configs register with lerobot's draccus choice registry through dataclass inheritance; I did not want to risk that for five repeated fields.
- **Finding 4, persistent job state.** With fix 2 a rerun recovers whatever a lost job left behind, so the in-process job stays. A `_cross_site` middleware for all POST routes is out of scope here, as the review says.

### Verification

- Plugin tests 12, bridge tests 6, webapp proxy 41: all pass; ruff clean. The `LeRobot Plugin Tests` workflow passed on 3246ddb0.
- The `manipulation_server` refactor has no unit test, so I built `manipulation` and its message packages in a throwaway container with no network and booted the node: parameter overrides applied, `LeRobot bridge listening`, then on a client heartbeat `LeRobot client connected`, `Sensor subscriptions started`, and `LeRobot client idle` after it left.
- Not re-tested on a robot: the publish flow with the new record file. The conversion logic is covered by the tests and the kill experiment above; the webapp only changed where it reads the record from.

lerobot/lerobot_robot_mars/convert.py
```diff
@@ -0,0 +1,407 @@
...
+    exported = {int(i) for i in record.get("episode_ids", [])}
+    if exported and has_dataset(out) and _read_json(out / "meta" / "info.json").get("total_episodes") == len(exported):
+        return exported
+    if not out.exists():
+        return set()
+    ours = record.get("root") == str(out) or (read_sidecar(out) or {}).get("source") == CONVERTER
+    if not ours:
+        raise FileExistsError(f"{out} already holds something this converter did not write; choose another --root")
+    log(f"{recording.name}: the converted copy at {out} is incomplete; rebuilding it from every episode")
+    shutil.rmtree(out)
```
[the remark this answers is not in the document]
DavidDobas: Agreed, fixed in ad782f2f. The export record only names a path, so it can no longer prove ownership. The converter now writes its marker into the output folder itself at creation (`meta/mars.json` with `source: mars2lerobot`, the head angle filled in after the episodes are read), and `_episodes_in_copy` clears a folder for a rebuild only when that folder carries the marker. A dataset recorded with `lerobot-record` through our client carries `source: lerobot`, and a foreign dataset carries none, so both are refused with a message instead. `test_a_folder_the_converter_did_not_write_is_never_cleared` now includes your case: a stale record of ours naming the very path a foreign dataset sits at. Re-tested with real SIGKILLs after two and after five episodes: the rerun still rebuilds to all six episodes.

DavidDobas:
@greptileai the stale-record finding is fixed in ad782f2f, with a reply on the thread. Please re-review.

lerobot/lerobot_robot_mars/convert.py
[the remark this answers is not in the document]
DavidDobas: The re-review called the fix incomplete without naming the gap, so I looked for it and found one: ad782f2f checked the folder's marker before clearing a folder, but the append path returned earlier, without the check. A foreign dataset at that path whose episode count happened to equal the stale record's would have had our episodes appended to it. In dfd45916 the ownership check comes first for anything that already exists at `out`: no marker, no append and no clear, only the refusal. The test now sets up exactly that case, a stale record naming the path and a matching episode count, and asserts the folder is byte-for-byte untouched; it fails on ad782f2f. If you had a different path in mind, please name it.

DavidDobas:
@greptileai dfd45916 closes the remaining gap in the stale-record finding, details on the thread. Please re-review.

DavidDobas:
@greptileai

DavidDobas:
## Follow-up to the re-check of dfd45916

One commit, ac9cf425.

**The new concern, fixed.** The sidecar marked a folder as the converter's, not as this skill's, so publishing skill Y under skill X's repository name would have rebuilt the folder from Y's episodes and pushed that over X's dataset. The marker now carries the skill (`skill`: the skill's folder name, which stays put where the display name can be edited), written when the folder is created and again with the head angle. `_episodes_in_copy` raises the same `FileExistsError` when the source or the skill differs, before it appends or clears anything, and the message tells a webapp user what to do: publish under another repository name. Pinned by `test_a_second_skill_cannot_take_over_the_first_ones_dataset`, which asserts X's copy still has its episodes afterwards.

A copy converted before this commit has no `skill` in its sidecar. It is taken as this skill's, which gives the one safe rebuild you describe on robots that already published, rather than an error nobody can clear from the dialog.

**Dead weight, removed.** `LEGACY_EXPORT_KEY` is gone from `convert.py`, and `EXPORT_KEY` and `DATASET_METADATA` from `hub_publish.py`. `convert.py` keeps its own `DATASET_METADATA` because it still reads the episode list from that file.

**The misleading log line.** Alternating one skill between two repositories still rebuilds each time; the log now says the copy does not match this skill's export record instead of calling it incomplete.

**Verification.** Plugin tests 13, webapp proxy 41, ruff clean. The real SIGKILL experiment still reruns to all six episodes, and the sidecar ends up with the converter, the skill and the head angle.

Left as they were, as optional in your list: publish job state in memory, the `uv.lock` regex for CPU torch (worth its own change, tested on a Jetson), and the five repeated connection fields.

lerobot/lerobot_robot_mars/convert.py
```diff
@@ -0,0 +1,422 @@
...
+
+    Our own copy is appended to only when it holds exactly the recorded episodes. One left by a run
+    killed before finalize() is unreadable, and one that lost episodes must not be topped up and uploaded
+    over the full dataset; both are rebuilt from every episode.
+    """
+    if not out.exists():
+        return set()
+    sidecar = read_sidecar(out) or {}
+    # No skill named: a copy from before the marker carried one, taken as this skill's.
+    if sidecar.get("source") != CONVERTER or sidecar.get("skill", recording.skill_id) != recording.skill_id:
```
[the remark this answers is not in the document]
DavidDobas: Agreed, fixed in 652ce6ce. The default was deliberate, to give robots that already published one rebuild instead of a refusal, but you are right that it reopens exactly the cross-skill replacement the marker exists to prevent. The check is now `sidecar.get("skill") != recording.skill_id`: a marker that names no skill is refused like any other folder that is not provably this skill's, before anything is appended or cleared. `test_a_second_skill_cannot_take_over_the_first_ones_dataset` now covers both cases, a marker naming another skill and a marker naming none, and asserts the first skill's copy keeps its episodes. The cost is that a copy converted before the marker carried a skill has to be moved away by hand once; the feature has not shipped, so that is one robot.

DavidDobas:
@greptileai the legacy-marker finding is fixed in 652ce6ce, reply on the thread. Please re-review.

theo-michel (approved):
LGTM

ros2_ws/src/brain/manipulation/manipulation/lerobot_bridge.py
```diff
@@ -0,0 +1,255 @@
...
+        if joints is None or base is None:
+            self._log(f"LeRobot bridge ignored a malformed action: {payload}")
+            return
+        if self.commands_blocked:
+            return
+        self._last_action = now
+        self._base_stopped = False
+        joints[5] = min(max(joints[5], GRIPPER_RANGE[0]), GRIPPER_RANGE[1])
+        self._on_arm(joints)
+        self._on_base(base[0], base[1])
```
theo-michel, line 221: **Any connected LeRobot client overrides Nav2 for the whole session.**

`_apply` calls `_on_base` unconditionally, and on the client side `send_action` always fills the base keys (`mars.py:124`, `action.get(name, 0.0)`), so even a pure-arm policy or leader-arm session sends `x.vel = theta.vel = 0.0` at 30 Hz. That becomes `_bridge_base` → `_publish_base` on `/cmd_vel_skills`.

`mars_nav/cmd_vel_mux.py` ranks the sources:

```python
SOURCES = [("teleop", "/cmd_vel_teleop", 0.5),
           ("skills", "/cmd_vel_skills", 0.5),
           ("nav",    "/cmd_vel_nav",    0.5)]
```

`/cmd_vel_skills` outranks `/cmd_vel_nav`, so a 30 Hz stream of zero twists keeps the skills channel permanently inside its 0.5 s window and `_fresh_source_above` drops every Nav2 command, with only a throttled log line. The watchdog never rescues it: `_last_action` is refreshed on every tick, so `_base_stopped` stays `False`.

To be fair, the existing learned and replay paths do the same thing (`_publish_base` on every step, lines 609 and 1061) — a *running skill* owning the base is the established model. The difference is that a skill is bounded and visible (`execution_running`), while a lerobot session is neither: an hour of leader-arm recording, or a client left connected, silently blocks navigation the whole time, and nothing in Known limits says so.

`_floats(payload, BASE_KEYS, required=False)` already distinguishes "absent" from "zero", so: only call `_on_base` when the payload actually carries a base key, and have `send_action` omit the keys when the action has none.

_(Edited: added the comparison with the existing skill paths.)_

ros2_ws/src/brain/manipulation/manipulation/lerobot_bridge.py
```diff
@@ -0,0 +1,255 @@
...
+        """Drain the action socket, apply the newest command, and run the base watchdog."""
+        if self._pull is None:
+            return
+        newest: dict | None = None
+        while True:
+            try:
+                raw = self._pull.recv(zmq.NOBLOCK)
+            except zmq.Again:
+                break
+            self._last_seen = now
```
theo-michel: **Any garbage byte on :5555 counts as a live client.**

`self._last_seen = now` is stamped before `_parse` has validated anything, so a port scanner or an unrelated tool that assumes :5555 is its own (a widely reused default) marks the bridge active with a single byte. `_lerobot_bridge_tick` then opens the camera subscriptions and `publish()` encodes and sends two JPEGs at 30 Hz for as long as the probe keeps poking, on a robot with no LeRobot user at all.

`test_malformed_actions_are_ignored` asserts this on purpose (`assert h.bridge.active(1.0)  # still counts as a live client`), so it is a deliberate choice — but "a live client" should mean "spoke the protocol", not "opened a socket". Moving the stamp below the `_parse` check costs nothing and keeps heartbeats working.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -857,11 +868,116 @@ def image2_callback(self, msg: Image):
...
+            self._bridge_base,
+            self._stop_robot,
+            port_actions=port_actions,
+            port_observations=port_observations,
+            jpeg_quality=int(self._bridge_param("jpeg_quality", 90)),
+            bind_address=str(self._bridge_param("bind_address", "*")),
+            on_head=self._bridge_head,
+            log=lambda message: self.get_logger().warn(message, throttle_duration_sec=2.0),
+        )
+        bridge.bind()
```
theo-michel: **A `bind()` failure here kills the whole manipulation server.**

`bridge.bind()` runs inside `ManipulationServer.__init__` with no guard. If :5555 or :5556 is already held (5555 is a very common default), or `bind_address` names an interface that does not exist, `zmq.Socket.bind` raises `ZMQError`, which propagates out of `__init__` → `main()` → the node never comes up and respawns forever. Poses, replay and learned behaviours all go down with it.

The `ImportError` handler eleven lines up exists precisely to avoid that, and the `act_trt` note at the top of this file states the rule: *"Importing it at module top would crash the whole behavior server (respawn loop)"*. The same degradation should cover `bind()` — log a warning, return `None`, carry on without the bridge.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
```diff
@@ -857,11 +868,116 @@ def image2_callback(self, msg: Image):
...
         self.latest_joint_state = msg
         self.latest_joint_timestamp = rclpy.time.Time.from_msg(msg.header.stamp)
 
+    def _start_lerobot_bridge(self) -> "LeRobotBridge | None":
+        bridge = self._open_lerobot_bridge()
+        if bridge is None:
+            return None
+        self._follow_operator_commands()
+        rate_hz = float(self._bridge_param("rate_hz", 30.0))
+        self.create_timer(1.0 / max(rate_hz, 1.0), self._lerobot_bridge_tick)
```
theo-michel: **This timer has no callback group, so the JPEG encode serialises against the sensor callbacks it depends on.**

There is no `callback_group=` anywhere in the `manipulation` package, so every subscription and timer on this node lives in the node's default `MutuallyExclusiveCallbackGroup`. Under `MultiThreadedExecutor(num_threads=4)` that group still runs exactly one callback at a time.

So with a client connected, `_lerobot_bridge_tick` inserts two 640×480 quality-90 `cv2.imencode` calls (~5–10 ms each on a Jetson) into that single slot 30 times a second, blocking `image1_callback`, `image2_callback` and `joint_state_callback` for that window — the bridge delays the very frames it is publishing, and with `depth=1` BEST_EFFORT queues a delayed callback is a dropped frame.

(The action server's execute callback is *not* in that group — rclpy runs it as a free executor task via `notify_execute` → `executor.create_task` — so this does not slow inference itself, only the sensor feed it reads.)

Giving the timer its own `MutuallyExclusiveCallbackGroup` lets the executor overlap it with the sensor callbacks; moving the encode off the executor thread entirely would be better still.

_(Edited: removed the claim that this competes with the inference path; checked against rclpy humble's `action/server.py`.)_

lerobot/lerobot_robot_mars/mars.py
```diff
@@ -0,0 +1,208 @@
...
+        missing = [name for name in STATE_NAMES if name not in action]
+        if missing:
+            raise ValueError(f"MARS action needs all six joint targets; missing {missing}")
+        payload = {name: float(action[name]) for name in STATE_NAMES}
+        payload.update({name: float(action.get(name, 0.0)) for name in BASE_NAMES})
+        if self.config.external_commands:
+            self._link.heartbeat_if_due()
+        else:
+            self._link.send(payload)
+        return {name: payload[name] for name in ACTION_NAMES}
```
theo-michel: **The recorded gripper action is not the one the robot executed — and the return value of `send_action` can't fix it.**

The bridge clamps joint6 to `GRIPPER_RANGE = (-0.6, 0.85)` before calling `_on_arm` (`lerobot_bridge.py:214`). On the recording side, lerobot 0.6.1's `record_loop` does:

```python
_sent_action = robot.send_action(robot_action_to_send)
...
action_frame = build_dataset_frame(dataset.features, action_values, prefix=ACTION)
```

i.e. it discards what `send_action` returns and records the teleop/policy output as-is (lerobot's own TODO on that line says the sent action *should* be what is recorded). So a leader arm or a policy that commands `joint6.pos = -1.2` produces a dataset teaching the next policy to emit `-1.2`, which the bridge clamps again — the training target and the executed command disagree at the range limits, and nothing in the data shows it.

Since the return value is ignored, the fix has to happen before the send: a `robot_action_processor` step in the plugin that applies the same clamp (the pattern lerobot points at), so the recorded action is the one the robot ran. At minimum, `GRIPPER_RANGE` deserves a mention in the plugin's README so nobody is surprised when a policy's gripper output "doesn't go all the way".

Severity: only bites when a source commands outside `[-0.6, 0.85]`, so lower than I first ranked it.

_(Edited after checking `lerobot_record.py` at v0.6.1: my first version claimed the return value was recorded; it isn't.)_

theo-michel (commented):
Reviewed the full diff at xhigh effort (10 finder angles + a gap sweep), excluding `uv.lock`. Very solid piece of work — the wire-protocol test that loads the real bridge module, the converter's "only a folder that says it is mine may be cleared" rule, and the honest Known-limits section are all the right instincts.

14 findings inline. The two I'd fix before merging:

1. **`lerobot_bridge.py:216` — any connected LeRobot client overrides Nav2 for its whole session.** `send_action` always fills `x.vel`/`theta.vel`, so even an arm-only policy or leader-arm session publishes zero twists on `/cmd_vel_skills` at 30 Hz, and `cmd_vel_mux` ranks that channel above `/cmd_vel_nav` with a 0.5 s window. Skills do the same while they run, but a skill is bounded and visible; a lerobot session is neither. Not in Known limits, and the watchdog can't catch it.
2. **`manipulation_server.py:909` — `bridge.bind()` in `__init__` can take the whole node down.** `EADDRINUSE` on :5555 crashes the manipulation server into the `respawn=True` loop, poses and replay included. The `ImportError` handler right above exists to prevent exactly this.

Then, in rough order: the two silent-data-loss paths in the client (no staleness check; blank frames on a camera that never comes up), the converter's narrow kill window that wedges a repo name with no way out of the UI, and the gripper clamp making the recorded action differ from the executed one.

The rest are smaller: the sim render loop's 2 ms spin leaking outside its flag, the default `bind_address: "*"`, two webapp job-failure modes, a Hub round-trip inside `connect()`, and a couple of coupling nits.

_Edited after a second pass against the lerobot 0.6.1 and rclpy humble sources: five inline comments were corrected in place (each says what changed) — most notably, lerobot-record discards `send_action`'s return value, so the gripper finding is lower-severity and needs a different fix than I first wrote; and the 64 KiB-line trigger in `hub_publish` is unlikely with the current tools, so that one is hardening, not a reproduced bug._

lerobot/lerobot_robot_mars/mars.py
```diff
@@ -0,0 +1,208 @@
...
+    @check_if_already_connected
+    def connect(self, calibrate: bool = True) -> None:
+        deadline = time.monotonic() + self.config.connect_timeout_s
+        self._absorb(self._link.open(self.config.connect_timeout_s))
+        while self._missing_cameras() and time.monotonic() < deadline:
+            message = self._link.latest(timeout_ms=200)
+            if message is not None:
+                self._absorb(message)
+        if self._missing_cameras():
+            logger.warning(
```
theo-michel: **Small deadline bug, plus a data-loss path that only warns once.**

1. `deadline` is computed on line 84, *before* `_link.open(self.config.connect_timeout_s)` on line 85, which can itself burn the whole 5 s. The camera-wait loop on line 86 is then left with whatever remains, possibly zero. Resetting the deadline after `open()` returns fixes that.

2. When cameras are still missing at the deadline, `connect()` logs and proceeds, and `get_observation` substitutes `_blank_frame()` (line 115). Frames *are* picked up as soon as the bridge starts sending them (`_absorb` fills `self._frames`), so this self-heals in the normal case — the warning message already says so. The residual risk is a camera that never comes up (driver down, wrong topic): the entire `lerobot-record` session then writes black frames for that camera with one warning at connect, and the loss is only visible on opening the dataset. Raising `DeviceNotConnectedError` when a camera is still missing at the deadline — or at least re-warning per episode — would turn that into a loud failure.

_(Edited: my first version said the blanks were never re-checked; they are, once frames arrive.)_

lerobot/lerobot_robot_mars/mars.py
```diff
@@ -0,0 +1,208 @@
...
+        return [camera for camera in CAMERA_ORDER if camera not in self._frames]
+
+    @check_if_not_connected
+    def disconnect(self) -> None:
+        self._link.close()
+
+    @check_if_not_connected
+    def get_observation(self) -> RobotObservation:
+        self._link.heartbeat_if_due()
+        message = self._link.latest(self.config.poll_timeout_ms)
```
theo-michel, line 109: **No staleness check — a dead bridge reads as a frozen but healthy robot.**

If the robot reboots, the Wi-Fi drops, or `manipulation_server` restarts mid-recording, `Link.latest()` returns `None` on every call, `_absorb` never runs again, and `get_observation` keeps handing back the same `self._state` and `self._frames` forever. `is_connected` stays `True` because it only checks that the socket object exists.

`lerobot-record` then writes hundreds of identical frames with fresh timestamps and no error. `MarsPassthroughTeleop.get_action` has the same shape (`dict(self._command)`).

The header already carries `seq` and `t` — rejecting an observation older than a few periods and raising `DeviceNotConnectedError` would turn a silent data-loss into a loud failure.

lerobot/lerobot_robot_mars/convert.py
```diff
@@ -0,0 +1,421 @@
...
+    skill published under the same repository name, which a rebuild would wipe locally and on the Hub.
+
+    Our own copy is appended to only when it holds exactly the recorded episodes. One left by a run
+    killed before finalize() is unreadable, and one that lost episodes must not be topped up and uploaded
+    over the full dataset; both are rebuilt from every episode.
+    """
+    if not out.exists():
+        return set()
+    sidecar = read_sidecar(out) or {}
+    if sidecar.get("source") != CONVERTER or sidecar.get("skill") != recording.skill_id:
```
theo-michel, line 375: **A run killed in one narrow window wedges this repo name forever, with no way out of the UI.**

`_episodes_in_copy` refuses any existing `out` whose sidecar does not name this converter and skill. But `convert_skill` writes that sidecar at line 307 — *after* `open_dataset` at line 303 has already created `out/meta/info.json`.

Kill the job in between (Ctrl-C, OOM, the webapp restarting) and the folder holds a lerobot dataset with no `mars.json`. Every later run, including the Datasets page's Publish button, now raises `FileExistsError: … holds a dataset that is not this skill's; publish under another repository name or move it away`. The user has to ssh in and `rm -rf`.

`test_a_copy_left_by_a_killed_run_is_rebuilt` covers the neighbouring case (killed *after* the sidecar) and misses this one. Writing the sidecar as part of creating the folder closes it.

webapp/proxy/hub_publish.py
```diff
@@ -0,0 +1,230 @@
...
+    finally:
+        job.running = False
+
+
+async def _run_step(job: Job, command: list[str], env: dict[str, str]) -> int:
+    process = await asyncio.create_subprocess_exec(
+        *command, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.STDOUT, env=env
+    )
+    assert process.stdout is not None
+    async for raw in process.stdout:
```
theo-michel, line 189: **Hardening: a non-`OSError` failure in `_run` leaves the job neither running, nor done, nor failed.**

`_run` catches only `OSError`. Anything else raised inside `_run_step` escapes into a task nobody awaits: the `finally` sets `job.running = False`, but `stage` stays `"installing"` / `"converting"` and `error` stays `""`. `publishModal.pollSoon` then sees a job that is not running, not `stage === "error"`, and (for setup) not `kind === "publish"`, so it calls `refresh()` → `showNeedsSetup` with an empty error: the dialog loops silently, the subprocess keeps running, and asyncio only logs *Task exception was never retrieved*.

The one concrete raiser I know of is `StreamReader`'s 64 KiB line limit (`ValueError` from `async for raw in process.stdout` on a line with no newline). I could **not** find a producer of such a line in this pipeline — uv and pip hide progress bars off-TTY, `HF_HUB_DISABLE_PROGRESS_BARS=1` is set, and lerobot 0.6.1 encodes video in-process with PyAV with no tqdm on the create/save/finalize path — so treat this as belt-and-braces rather than a reproduced bug: `except Exception as e: job.stage, job.error = "error", str(e)` is a one-line safety net.

_(Edited: downgraded — my first version presented the long-line trigger as likely; it isn't with the current tools.)_

webapp/proxy/hub_publish.py
```diff
@@ -0,0 +1,230 @@
...
+    _start(job, [_niced(command)], {"HF_TOKEN": token, "HF_HUB_DISABLE_PROGRESS_BARS": "1"})
+
+
+def start_setup() -> None:
+    uv = _uv_command()
+    steps = []
+    if uv is None:
+        steps.append([sys.executable, "-m", "pip", "install", "--user", "--quiet", "uv"])
+        uv = [sys.executable, "-m", "uv"]
+    locked = _locked_versions(LEROBOT_DIR / "uv.lock")
```
theo-michel, line 128: **If the lock regex matches nothing, the setup installs the 3.5 GB CUDA torch these lines exist to avoid.**

`_LOCK_ENTRY` requires exactly `[[package]]\nname = …\nversion = …` with `\n` line endings. If it matches nothing — a uv lock-layout change, or a CRLF checkout — `_locked_versions` returns `{}` and:

- `skipped` is empty → the `uv sync` on line 119 gets no `--no-install-package`, so it installs torch from PyPI: the ~3.5 GB of CUDA libraries the comment on lines 123–124 explicitly says a Jetson cannot use.
- `torch` is empty → line 127 expands to `uv pip install --python … --index-url …` with no packages, which exits non-zero.

So the user waits through a 3.5 GB download on the robot's uplink and *then* sees a failure. A guard like `if "torch" not in locked: raise …` before `_start` fails fast and cheaply.

_(Edited: my first version blamed the `OSError` path; that one isn't reachable in practice because `uv sync --frozen` fails on the same unreadable file. The regex path is the real one.)_

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/node.py
```diff
@@ -506,16 +516,24 @@ def _render_loop(self) -> None:
...
+                # not the raw rate is in force (10 of the head's 15 fps is 2 frames in 3).
+                jpeg_due = now >= next_jpeg[camera] - 0.005
+                if jpeg_due:
+                    step = jpeg_period[camera] * stretch
+                    next_jpeg[camera] = max(next_jpeg[camera], now - step) + step
+                self._publish_camera_frames(pub if jpeg_due else None, raw_pub, camera, jpeg)
             if not rendered:
-                time.sleep(0.02)
+                # A 20 ms nap would eat most of a 30 fps period; poll finely while anyone is watching.
+                time.sleep(0.002 if any(wanted.values()) else 0.02)
```
theo-michel: **The 2 ms spin is not gated on the flag it was added for.**

The condition is `any(wanted.values())`, not `HARDWARE_CAMERA_RATES`. So every sim user who opens the web app — with the flag unset, the default — now pays a 500 Hz poll: six `get_subscription_count()` calls plus the demand/stretch arithmetic, ~500 times a second, on a render thread the surrounding code already treats as saturation-prone, for a main camera whose period is 100 ms.

The comment justifies it with *"a 20 ms nap would eat most of a 30 fps period"*, and that period only exists under the opt-in flag. Either gate on `HARDWARE_CAMERA_RATES`, or derive the nap from `min(period[s] for s in wanted if wanted[s])` so it adapts instead of assuming the worst case.

ros2_ws/src/brain/manipulation/config/manipulation_server.yaml
```diff
@@ -18,3 +18,15 @@ manipulation_server:
...
+    # in lerobot/README.md.
+    lerobot_bridge:
+      enabled: true
+      port_actions: 5555        # PULL: clients push JSON actions / heartbeats
+      port_observations: 5556   # PUB: 'state' (JSON) and 'obs' (JSON + one JPEG per camera)
+      rate_hz: 30.0
+      jpeg_quality: 90
+      # No login, like the rosbridge websocket: anyone on the robot's network can connect.
+      # "127.0.0.1" keeps the bridge to lerobot processes on this Jetson; enabled: false closes it.
+      bind_address: "*"
```
theo-michel, line 32: **`enabled: true` + `bind_address: "*"` ships unauthenticated arm control on every interface.**

Anyone on the robot's network — a café Wi-Fi, a shared lab LAN, a guest VLAN — can PUSH six joint targets to :5555 and move the arm, drive the base, or tilt the head, with no credential and no pairing; and read the head and wrist camera streams off :5556.

The sim's own compose file binds these two ports to `127.0.0.1` and calls it *"Loopback only, like rosbridge"*. `bind_address` exists for exactly this, so `"127.0.0.1"` with LAN exposure as an opt-in seems like the right default — otherwise the open bind reaches every robot at the next `innate update`, including ones whose owners never read `lerobot/README.md`.

lerobot/lerobot_robot_mars/dataset_meta.py
```diff
@@ -0,0 +1,160 @@
...
+
+def head_angle_for_run(argv: Sequence[str]) -> tuple[float, str] | None:
+    """The head tilt of the dataset behind this run, with where it was found; None if unknown.
+
+    A run relates to a dataset in one of two ways: it names one (resuming a recording, replaying an
+    episode), or it runs a policy, whose checkpoint records the dataset it was trained on. Either
+    way that dataset's ``meta/mars.json`` says how the head was tilted when the frames were taken,
+    which is how it must be tilted now.
+    """
+    for repo_id, root in (_named_dataset(argv), _policy_dataset(argv)):
```
theo-michel: **A blocking Hub round-trip on the documented first-record path.**

The tuple is built eagerly, so `_policy_dataset(argv)` runs even when `_named_dataset` already answered. More costly: for the README's own example (`--dataset.repo_id=YOUR_HF_NAME/mars-tidy-up`, a dataset being *created*), `_sidecar` finds nothing locally and falls through to `hf_hub_download(repo_id, "meta/mars.json", repo_type="dataset")` on line 133 — a Hub request that must 404 before recording can start, on every run, and that hangs for the full `huggingface_hub` timeout on a captive or offline network.

Make the tuple lazy, and skip the Hub lookup when the named dataset has no local root yet.

webapp/proxy/hub_routes.py
```diff
@@ -0,0 +1,116 @@
...
+
+from __future__ import annotations
+
+import asyncio
+from urllib.parse import urlsplit
+
+import aiohttp
+import hub_publish
+from aiohttp import ContentTypeError, web
+from media_routes import _resolve_under_root
```
theo-michel: **Two small coupling nits.**

- `_resolve_under_root` is imported across modules by its private name. Worth promoting it to a public helper in `media_routes` since it is now the shared skill-root fence for two route modules.
- The export-record contract is written twice: `convert.py:41` defines `EXPORT_FILE` and the `{repo_id, root, episode_ids, pushed}` shape, `hub_publish.py:28` re-declares the filename and reads the fields. `convert.py` also falls back to `skill_dir` when `data/` is absent, while `hub_publish.published()` only ever looks in `data/`. A rename on the converter side leaves the Datasets page quietly reporting "never published" with nothing failing.

That is the same two-codebases hazard the wire protocol has, but the wire constants are pinned by `test_protocol_constants_match` and these are not. Either share the constant or add the equivalent test.

DavidDobas:
## Two gaps David found in the publish flow, fixed in 745bd928

**1. A deleted episode stayed on the Hub for good.** Publishing only appended. Measured before the fix: publish a 4-episode skill, delete one episode on the Datasets page, publish again, and the dataset still had 4 ("Nothing new to publish"); record one more and it had 5 while the skill had 4. The same held for an episode marked failed after a publish. People delete an episode precisely to keep a bad demonstration out of training.
- `_episodes_in_copy` now also takes the skill's current episode ids. A copy holding an episode that is no longer among them is rebuilt in full from the current episodes, through the existing rebuild path. The recorder never reuses a deleted id (`number_of_episodes` is the id counter), so an id can never point at different data.
- The rebuilt copy can have fewer files than the last upload, and `push_to_hub` uploads without deleting. That would leave stale parquet on the Hub, and lerobot loads every `data/*/*.parquet` and `meta/episodes/*/*.parquet` it finds (`io_utils.load_nested_dataset`). New `hub.push_mirror` deletes Hub files under `data/`, `meta/` and `videos/` that the local copy lacks, then pushes. It deletes first on purpose: lerobot moves its `v3.0` tag only after the upload, so readers stay on the old, consistent revision until the new one is complete. `README.md`, `.gitattributes` and anything outside those three folders are never touched.
- `mars-push` keeps plain `push_to_hub`. Its local copy may be a partial download, and mirroring that would delete episodes on the Hub.
- The dialog says so: "1 of its episodes has been deleted since, so the dataset is rebuilt from this skill's current episodes and replaces the one on Hugging Face." `published()` reports the count.

**2. The Private box did nothing on a republish.** `create_repo(exist_ok=True)` ignores `private` for an existing repository (huggingface_hub documents this), so a public dataset stayed public while the box, ticked by default, suggested otherwise.
- New `GET /hub/repo?repo_id=…` asks the Hub whether the dataset exists and whether it is private, with the saved token. `whoami` and it now share one `_ask_hub` helper.
- For an existing dataset the dialog replaces the box with "Public on Hugging Face. Publishing keeps that; change it in the dataset's settings", linking to its settings page. It re-checks as the name or owner changes, and ignores a stale answer.
- The robot never changes visibility itself: flipping a public dataset to private from a box that defaults to ticked is the surprise to avoid.
- `mars-push --private` on an existing public dataset now says it stays public and where to change it.

**Tests and checks**
- New `test_a_deleted_episode_leaves_the_published_dataset` (data-loss rule): delete a published episode the way the Datasets page does, publish again, and the dataset and the export record hold only the remaining episode. Plugin 14, webapp proxy 41, bridge 6; ruff clean; `tsc` reports no errors in `publishModal.js`, against the webapp's existing baseline.
- `push_mirror` against a stand-in `HfApi`: only the two stale files under `data/` and `videos/` are deleted, `README.md` and a file outside the dataset folders are left, and the delete comes before the push. Not run against the real Hub, to avoid writing to anyone's account.
- `/hub/repo` against the real Hub, read-only: `innate-inc/mars-pick-tv` gives exists and public, a missing name gives not existing, a malformed one a 400.
- The dialog in headless Chrome against this branch's webapp server: the existing public dataset shows the visibility line and settings link, a new name brings the Private box back (ticked), the removed-episode summary appears, and no page errors. The link was unreadable in the browser's default blue on the dark panel, so it is styled in the text colour.

DavidDobas:
@greptileai please review 745bd928.

lerobot/lerobot_robot_mars/hub.py
```diff
@@ -0,0 +1,62 @@
...
+        remote = api.list_repo_files(dataset.repo_id, repo_type="dataset")
+        stale = [name for name in remote if name.split("/", 1)[0] in DATASET_DIRS and name not in local]
+        if stale:
+            api.delete_files(
+                dataset.repo_id,
+                delete_patterns=stale,
+                repo_type="dataset",
+                commit_message="Remove files the rebuilt dataset no longer has",
+            )
+    dataset.push_to_hub(private=private)
```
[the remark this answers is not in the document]
DavidDobas: Agreed, fixed in 908b5213. The delete and the upload were two commits, so a failed upload left main without the deleted files (the `v3.0` tag still pointed at the old revision, but main and the dataset viewer did not). `push_mirror` now makes one `upload_folder` commit that uploads this copy and deletes every other file under `data/`, `meta/` and `videos/` (`delete_patterns`; huggingface_hub builds the deletions and additions into the same commit and drops deletions of files it is re-uploading). If that commit fails, the Hub keeps the previous dataset whole. `push_to_hub` then finds nothing to upload, which huggingface_hub skips instead of making an empty commit, and only refreshes the card and moves the tag; if that step fails, main already holds the complete new dataset and only the tag lags. A new repository skips the replacing commit and goes straight to `push_to_hub`.

DavidDobas:
@greptileai the upload-ordering finding is fixed in 908b5213, reply on the thread. Please re-review.

DavidDobas:
## @theo-michel, your review is addressed in 3c576918 and ddd33337

Thank you, this was a careful one. Twelve of your thirteen comments are fixed as you suggested, or close to it. One is left for David to decide.

**The two you would fix before merging**
- **Nav2 override** (`lerobot_bridge.py`): the bridge drives the base only when an action carries `x.vel` or `theta.vel`, and `Mars.send_action` no longer fills them in with zeros. An arm-only session (leader arm, arm-only policy) now leaves the base to navigation. The watchdog is armed only by base commands. A client that does send base velocities still outranks navigation while it does, as a skill does; that is now under Known limits and in the README. Test: `test_an_arm_only_action_leaves_the_base_to_navigation`.
- **`bind()` crashing the node** (`manipulation_server.py`): a `ZMQError` from bind now closes whatever was opened, logs `LeRobot bridge disabled: cannot bind …`, and the server carries on. `bind()` now holds each socket as it opens it, so `close()` releases a half-bound pair. Boot-checked in a throwaway container: normal boot, then again with :5655 already held; the node stays up with the warning.

**Silent data loss in the client**
- **No staleness check:** `Link.ensure_alive()` raises `DeviceNotConnectedError` once the bridge has not answered for over 1 s of asking. `Mars.get_observation` and `MarsPassthroughTeleop.get_action` both call it. The clock starts at the first unanswered poll, not at the last message, so lerobot pausing to save an episode is not mistaken for a dead robot. Test: `test_a_bridge_that_goes_silent_is_a_disconnect_not_a_frozen_robot`.
- **Blank frames and the deadline:** the camera deadline now starts after `open()` returns. A camera still missing at the deadline raises instead of recording blank frames, so `_blank_frame` is gone.

**Converter kill window** (fixed in ddd33337; my first attempt in 3c576918 cleared any episode-less folder, and Greptile rightly showed that would delete a lerobot-record session that had just started): lerobot's `create()` refuses an existing folder, so the marker cannot come first. A new copy is therefore built in a hidden staging folder beside `out` (`.<name>.partial`), marked, and moved into place with `os.replace` only after `finalize()`. `out` never exists half-made, the ownership check stays strict, and a killed build leaves only the staging folder, which the next run removes.

**Gripper recorded vs executed:** agreed on your corrected reading. lerobot-record builds its processors with `make_default_processors()` and gives a plugin no hook to add a step, so the clamp cannot happen before recording from our side. The range and the recording caveat are now in the README's safety list and in Known limits. Our own recording path, the passthrough teleop, records `/mars/arm/commands`, which is what the arm actually received.

**Smaller ones**
- **Junk bytes waking the bridge:** `_last_seen` is stamped only after a message parses; the test now checks a raw junk byte on its own.
- **Callback group:** the bridge timer has its own `MutuallyExclusiveCallbackGroup`.
- **2 ms spin:** gated on `HARDWARE_CAMERA_RATES`; the default goes back to the 20 ms nap.
- **Hub round trip in `connect()`:** the named dataset is looked up on disk only, since lerobot-replay and lerobot-record load it before they connect. The policy's training dataset is looked up only if that finds nothing, and it is the only path that may still ask the Hub.
- **`_run` catching only `OSError`:** now `except Exception` with the reason on the line, so any failure ends in an error the dialog shows.
- **Lock regex matching nothing:** `start_setup` raises `LockUnreadable` when the torch versions are missing, before anything downloads, and the setup route returns that message.
- **Coupling:** `_resolve_under_root` is now the public `resolve_under_root`. `hub_publish.published()` reads the record from the same place the converter writes it (`data/`, else the skill folder). The contract is pinned by `test_the_webapp_reads_the_export_record_the_converter_writes`, which runs the real `hub_publish.py` against a real conversion.

**Left for David: `bind_address: "*"`.** Loopback by default would close the exposure you describe. It would also mean the documented setup, LeRobot on your own computer talking to the robot, fails on every robot until the owner edits `manipulation_server.yaml`. The robot already exposes the same arm, base and camera control through rosbridge on all interfaces with no login. This was discussed with Greptile earlier and led to `bind_address` becoming a parameter. Whether to flip the default is a product call. I have not changed it.

**Also in this commit, from the second reviewer:** publishing now refuses a Hub name this skill's export record does not name. `push_mirror` replaces the Hub dataset wholesale, so without the check another robot's dataset or a lerobot-record upload under the same name would have been silently replaced. The check runs before converting, and the dialog disables Publish with a message for such a name.

Verified: plugin tests 17, bridge 7, webapp proxy 41, ruff clean, no new `tsc` errors in the dialog, the node boot-checked as above, and the dialog's three visibility and name states in headless Chrome.

DavidDobas:
@greptileai please review 3c576918.

lerobot/lerobot_robot_mars/convert.py
```diff
@@ -0,0 +1,462 @@
...
+    that lost episodes must not be topped up and uploaded over the full dataset; one holding an episode
+    since deleted or relabelled a failure would keep publishing it. All three are rebuilt in full.
+    """
+    if not out.exists():
+        return set()
+    if _holds_no_episodes(out):
+        # Nothing to lose. It is also what a run killed between lerobot creating the dataset and this
+        # converter marking it leaves behind: lerobot refuses to create into an existing folder, so the
+        # mark cannot come first, and refusing here would wedge the repository name for good.
+        shutil.rmtree(out)
```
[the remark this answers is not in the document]
DavidDobas: Agreed, fixed in ddd33337 by removing the window instead of guessing about it. A new copy is now built in a hidden staging folder beside `out` (`.<name>.partial`), marked with the converter and skill, and moved into place with `os.replace` only after `finalize()`. So `out` never exists half-made, and the episode-less rule is gone: the ownership check is strict again, and a folder without this skill's marker is refused however many episodes it holds. A killed build leaves only the staging folder, which the next run removes. `test_a_folder_the_converter_did_not_write_is_never_cleared` now uses your case, a foreign dataset with zero episodes, as a lerobot-record session that just started would be. It fails on 3c576918 and passes now. A real SIGKILL mid-build still reruns to all six episodes.

lerobot/lerobot_robot_mars/wire.py
```diff
@@ -0,0 +1,168 @@
...
+                socket.close()
+        if self._context is not None:
+            self._context.term()
+        self._context = self._push = self._sub = None
+
+    def latest(self, timeout_ms: int) -> Message | None:
+        """The newest message, waiting up to *timeout_ms* for one; None if nothing arrived."""
+        sub = self._require(self._sub)
+        if not sub.poll(timeout_ms, zmq.POLLIN):
+            self._asking_since = self._asking_since or time.monotonic()
```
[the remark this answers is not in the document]
DavidDobas: Applied in ddd33337: a gap longer than the heartbeat period between asks restarts the clock, so only continuous asking can run it out. For the record, I could not make the false disconnect happen with this bridge. While the client pauses, the bridge keeps streaming until it goes idle, and the conflated SUB socket keeps the newest of those messages, so the first poll after any pause finds a message and resets the clock. A test written for your scenario passed on the old code too, so under this repo's test policy I did not keep it. The change is still the right rule and costs nothing.

DavidDobas:
@greptileai both findings are addressed in ddd33337, with replies on the threads. Please re-review.

ros2_ws/src/brain/manipulation/manipulation/lerobot_bridge.py
DavidDobas: Fixed in 3c576918. `_apply` drives the base only when the action carries `x.vel` or `theta.vel`, and `Mars.send_action` no longer fills them in with zeros, so an arm-only session (leader arm, arm-only policy) leaves the base to Nav2. The watchdog is armed only by base commands. A client that does send base velocities still outranks navigation while it does, as a skill does; that is in Known limits and the README's safety list. Test: `test_an_arm_only_action_leaves_the_base_to_navigation`.

ros2_ws/src/brain/manipulation/manipulation/lerobot_bridge.py
DavidDobas: Fixed in 3c576918: `_last_seen` is stamped only after a message parses as a JSON object. `test_malformed_actions_are_ignored` now checks a raw junk byte on its own and asserts the bridge stays idle.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
DavidDobas: Fixed in 3c576918. A `ZMQError` from `bind()` now closes whatever was opened, logs `LeRobot bridge disabled: cannot bind …`, and the server carries on without the bridge. `bind()` holds each socket as it opens it, so `close()` releases a half-bound pair. Boot-checked in a throwaway container with :5655 already held: the warning appears and the node stays up.

ros2_ws/src/brain/manipulation/manipulation/manipulation_server.py
DavidDobas: Fixed in 3c576918: the bridge timer now has its own `MutuallyExclusiveCallbackGroup`, so the encode overlaps with the sensor callbacks under the 4-thread executor. The tick only reads the image references those callbacks replace, and the JPEG cache compares by identity, so no locking is needed.

lerobot/lerobot_robot_mars/mars.py
DavidDobas: Agreed with your corrected reading. lerobot-record builds its processors with `make_default_processors()` and gives a plugin no hook to add a step, so we cannot clamp before recording from the plugin. Done in 3c576918: the range and the recording caveat are in the README's safety list and in Known limits. Our own recording path, the passthrough teleop, records `/mars/arm/commands`, which is what the arm actually received, so it is not affected.

lerobot/lerobot_robot_mars/mars.py
DavidDobas: Both fixed in 3c576918. The camera deadline now starts after `open()` returns. A camera still missing at the deadline raises `DeviceNotConnectedError` naming it, instead of recording blank frames, so `_blank_frame` is gone.

lerobot/lerobot_robot_mars/mars.py
DavidDobas: Fixed in 3c576918, refined in ddd33337. `Link.ensure_alive()` raises `DeviceNotConnectedError` once the bridge has not answered for over 1 s of continuous asking. Both `Mars.get_observation` and `MarsPassthroughTeleop.get_action` call it. A pause on the client side, such as lerobot saving an episode, restarts the clock, so it is not mistaken for a dead robot. Test: `test_a_bridge_that_goes_silent_is_a_disconnect_not_a_frozen_robot`.

lerobot/lerobot_robot_mars/convert.py
DavidDobas: Fixed in ddd33337, after a first attempt in 3c576918 that Greptile rightly rejected. lerobot's `create()` refuses an existing folder, so the marker cannot be written first. A new copy is therefore built in a hidden staging folder beside `out` (`.<name>.partial`), marked, and moved into place with `os.replace` only after `finalize()`. `out` never exists half-made, the ownership check stays strict, and a killed build leaves only the staging folder, which the next run removes. The killed-run test now covers a leftover staging folder.

webapp/proxy/hub_publish.py
DavidDobas: Fixed in 3c576918: `except Exception as e:  # noqa: BLE001 — a background job must end in a state the dialog can show`, so any failure lands in `stage = "error"` with its message.

webapp/proxy/hub_publish.py
DavidDobas: Fixed in 3c576918. `start_setup` raises `LockUnreadable` when either torch version is missing from the lock, before anything downloads, and `POST /hub/setup` returns that message to the dialog.

ros2_ws/src/mars_bot/mars_sim_driver/mars_sim_driver/node.py
DavidDobas: Fixed in 3c576918: the 2 ms nap now applies only when `HARDWARE_CAMERA_RATES` is on and something subscribes; the default goes back to the 20 ms nap.

lerobot/lerobot_robot_mars/dataset_meta.py
DavidDobas: Fixed in 3c576918. The two lookups are now lazy, and the named dataset is looked up on disk only: lerobot-replay and lerobot-record load it before they connect, so it is always local by then. Only the policy's training dataset may still be fetched from the Hub, and only when it is not on disk.

webapp/proxy/hub_routes.py
DavidDobas: Both fixed in 3c576918. `_resolve_under_root` is now the public `resolve_under_root`. `hub_publish.published()` reads the record from the same place the converter writes it (`data/`, else the skill folder). The contract is pinned by `test_the_webapp_reads_the_export_record_the_converter_writes`, which runs the real `hub_publish.py` against a real conversion, like the wire-protocol tests do for the bridge.

ros2_ws/src/brain/manipulation/config/manipulation_server.yaml
DavidDobas: Not changed; I have left this one for David to decide. Loopback by default would close the exposure you describe, but the documented setup, LeRobot on your own computer talking to the robot, would then fail on every robot until the owner edits `manipulation_server.yaml`. The robot already exposes the same arm, base and camera control through rosbridge on all interfaces with no login, and that trust model is why this was left open when Greptile raised it; `bind_address` became a parameter then. Leaving this thread open until David makes the call.

lerobot/lerobot_robot_mars/convert.py
```diff
@@ -0,0 +1,453 @@
...
+            progress({"event": "done", "url": "", "message": "Nothing new to publish"})
+        return out
+
+    # A new copy is built beside `out` and moved into place only once it is finalized and marked as this
+    # skill's, so `out` never exists half-made: lerobot refuses to create into an existing folder, so a
+    # marker could not come first, and an unmarked folder is one the ownership check must refuse.
+    fresh = not out.exists()
+    build = out.with_name(f".{out.name}.partial") if fresh else out
+    if fresh:
+        shutil.rmtree(build, ignore_errors=True)  # what a killed earlier build left; its name makes it ours
```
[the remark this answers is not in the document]
DavidDobas: Agreed, fixed in 2dbc6ade. `convert_skill` now holds an exclusive, non-blocking `flock` on `.<name>.lock` beside the output folder for the whole conversion. A second conversion into the same folder is refused with "another conversion into … is running; wait for it to finish" before it touches the staging folder or `out`. The lock goes with the process, so a killed run never leaves it held: the real SIGKILL experiment still reruns straight away. The web app already runs one job at a time, so this covers a command-line run overlapping it or another command-line run. Windows has no `flock`; there the lock is skipped, and that is stated where it is taken. Test: `test_a_second_conversion_into_the_same_folder_waits_its_turn`.

DavidDobas:
@greptileai the concurrent-conversion finding is fixed in 2dbc6ade, reply on the thread. Please re-review.

ros2_ws/src/brain/manipulation/config/manipulation_server.yaml
DavidDobas: Decision from David: `bind_address` stays `"*"`. The robot trusts its local network across the board. The web app listens on all interfaces with no login and gives full teleop, both cameras and, through `/ws`, rosbridge publishing to any topic. The Zenoh router listens on the network by default, and SSH takes a password. Loopback on the bridge alone would leave all of that open and only make the documented LeRobot-on-your-computer setup fail until the owner edits the robot's config. Protecting the robot on shared networks needs a login or pairing for all of its network services together, which is its own piece of work. `bind_address: "127.0.0.1"` and `enabled: false` remain for anyone who wants the bridge closed, and the README says so.

